PR #18, opened 2018-03-22
roll back css changes

refi93:
roll back changes made on footer and preserving the contacts in bottom part of the web


frontend/components.js
```diff
@@ -213,7 +213,7 @@ const Footer = (state) => `
     <p>
       Powered with 🚀 tech in
       <a href="https://vacuumlabs.com" target="_blank">
-        <img src="/assets/vacuumlabs-logo-light.png" class="logo" alt="Vacuumlabs Logo" />
+        <img src="/assets/vacuumlabs-logo-dark.svg" class="logo" alt="Vacuumlabs Logo" />
       </a>
     </p>
     &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
darjanin: Please remove all `&nbsp;`. If you want there space, then use the CSS. The `&nbsp;` can be used for example if you article _a_ and you want to connect it with following noun. For example `a&nbsp;man` so _man_ doesn't wrap without article

darjanin: Same for `&nbsp;` between links.

PR #34, opened 2018-03-27
Add links to blockchain explorer

### LINKED ISSUE #32: Add links to blockchain explorer
addresses and transaction IDs (or some icons nearby) should be clickable

git-jano:
closes #32


public/css/styles.css
```diff
@@ -120,6 +120,16 @@ td {
...
   margin-top: 1rem;
 }
+.inline-img {
+  height: 1.1rem;
+  margin-left: 5px;
+}
+.address-wrap {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
```
darjanin: Don't set the default `flex-wrap`. If you are using the default value then don't write it. While other person is reading it, he can think that it's overriding other CSS rule, what's not case now.

public/css/styles.css
```diff
@@ -120,6 +120,16 @@ td {
 .address + .address {
   margin-top: 1rem;
 }
+.inline-img {
```
darjanin: This name is misleading. I was accepting that it sets something like `display: inline-block`, but it changes only padding.

I would prefer if the link will look like this:
ezgif-3-95d991cd59

public/css/styles.css
```diff
@@ -120,6 +120,16 @@ td {
 .address + .address {
```
darjanin: You need to add `.address-wrap + .address-wrap` selector, to add spaces between wrapped inputs with addresses.

public/css/styles.css
```diff
@@ -120,6 +120,16 @@ td {
...
   margin-top: 1rem;
 }
+.inline-img {
+  height: 1.1rem;
+  margin-left: 5px;
```
darjanin: Use pixel value only for `@media` queries and border's width equal to `1px`.

frontend/components.js
```diff
@@ -62,11 +62,18 @@ const WalletHeader = (state) => `
   </div>
 `
 
+const address = (address) => `
+          <div class="address-wrap"z>
+          <input readonly type="text" class="address" value="${address}"/>
+          <a href="https://cardanoexplorer.com/address/${address}" target="_blank"><img src="/assets/link.png" class="inline-img" title="examine via CardanoExplorer.com" alt="examine via CardanoExplorer.com"></a>
```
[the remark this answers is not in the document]
darjanin: Add `title` attribute to `<a` with text that is in `alt`.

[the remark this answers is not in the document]
git-jano: works without  `::after`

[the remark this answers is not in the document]
darjanin: I understand that it works without `::after`. The idea to use `::after` was to have option to add text that will be before the icon. If it wasn't obvious where the link is navigating, you will have button like <kbd>__EXPLORER__ 🔗</kbd>.

frontend/components.js
```diff
@@ -62,11 +62,18 @@ const WalletHeader = (state) => `
   </div>
 `
 
+const address = (address, isTransaction = false) => `
+          <div class="address-wrap">
+          <input readonly type="text" class="address" value="${address}"/>
+          <a href="https://cardanoexplorer.com/${isTransaction ? 'tx' : 'address'}/${address}" target="_blank" title="examine via CardanoExplorer.com"></a>
```
refi93: It would be nice to have there also a button to copy the address to clipboard. But at the same time I'm not really sure if it is a good idea for (smaller) touch screens. Maybe @darjanin could give his opinion/suggestions

darjanin:
@git-jano please fix margin between buttons and last `.address-wrap`.
screen shot 2018-03-29 at 10 31 58

frontend/components.js
darjanin: What about option when the input's focus event is fired, the content of the input is copied to clipboard? We will show some tooltip about it later saying what happened. 
If it has some issues, like you like to select your address without copying, then I would put there two buttons - one is existing link, next will be <kbd>Copy to 📋</kbd> that will be on the end of the address input, and on the mobile this two button will be below address bar. 
 
```
# desktop version
|.address----------------| Copy to 📋 | 🔗 |

# mobile version
|.address----------------|
|--copy to 📋--|---🔗---|
```

refi93: I prefer the second option, i.e. two buttons. I generally don't like when the clipboard gets overwritten without prior notice.

frontend/components.js
```diff
@@ -62,11 +62,18 @@ const WalletHeader = (state) => `
   </div>
 `
 
+const address = (address, isTransaction = false) => `
```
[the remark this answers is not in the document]
git-jano: or It was just as a hint, that it expects "boolean", but `is` prefix means the same

frontend/components.js
git-jano: but for aesthetics: imho choose text or icon, not both, I prefer icons, i they are really intuitive (e.g. 📋 is for me ok, but  🔗 is not)

darjanin: I just used emojis for communicating the idea. At the moment icon is fine. For copy to clibpoard can be used same icon as Github uses.

PR #35, opened 2018-03-28
VirtualDOM reconciliation

mpinter:


frontend/simpleRedux.js
```diff
@@ -5,6 +5,89 @@ const {CARDANOLITE_CONFIG} = require('./frontendConfigLoader')
...
+    // encountered in the virtual DOM, unless the key matches the one on the 'nextChild' virtual
+    // node - this minimizes the amount of reflows, as it allows us to keep the original node,
+    // and ensures we do not lose focus on keyed inputs which did not move in the DOM.
+    if (currentChild.dataset.key && virtualChildren[0].dataset.key !== currentChild.dataset.key) {
+      !virtualKeyedChildren.has(currentChild.dataset.key) && existingNode.removeChild(currentChild)
+      existingChildren.shift()
+      continue
+    }
+
+    const nextChild = virtualChildren.shift()
```
tomaskulich: What is the cost for shift() - is it O(1) or O(n) - in the latter case, the whole reconciliation takes O(n^2) #sad.

tomaskulich: I got curious, so I found out:

```
const a = []

for (let i = 0; i < 1000000; i++) {
  a.push(i)
}

while (a.length > 0) {
  // if I do this, it takes 70ms
  //a.pop()
  // if I do this, it takes ages
  //a.shift()
}
```

frontend/simpleRedux.js
```diff
@@ -16,14 +99,15 @@ const dispatch = (updater, message, payload) => {
       nextState = midFn(nextState, previousState)
     })
     state = nextState
-    root.target.innerHTML = root.component(nextState, previousState)
+    root.virtualDOM.innerHTML = root.component(nextState, previousState)
```
tomaskulich: I'm not sure how does this work / cost. Let's discuss.

frontend/simpleRedux.js
```diff
@@ -36,33 +120,31 @@ const dispatch = (updater, message, payload) => {
...
+    router: {
+      pathname: '',
+      hash: '',
+    },
+  })
   middlewares = middlewareArray
+  Object.assign(root, {
+    component: rootComponent,
+    target: rootTarget,
+    virtualDOM: rootTarget.cloneNode(false),
```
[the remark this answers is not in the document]
mpinter: I don't think so, as all of innerHTML is replaced every time .. this is just to have same attributes on the root node

frontend/simpleRedux.js
```diff
@@ -5,6 +5,102 @@ const {CARDANOLITE_CONFIG} = require('./frontendConfigLoader')
...
+  const [existingChildren, virtualChildren] = [existingNode.children, virtualNode.children].map(
+    (children) => createReverseArray(children).filter((n) => n instanceof HTMLElement)
+  )
+
+  // attributes is a NamedNodeMap, which does not support same api as an array
+  for (let i = 0; i < virtualNode.attributes.length; i++) {
+    if (
+      existingNode.getAttribute(virtualNode.attributes[i].name) !== virtualNode.attributes[i].value
+    ) {
+      existingNode.setAttribute(virtualNode.attributes[i].name, virtualNode.attributes[i].value)
```
mpinter: Setting attributes like this works - I don't know about performance, but I don't see other way around it, apart for skipping those that did not change (now done in the last commit).

frontend/components.js
```diff
@@ -185,7 +185,7 @@ const Navbar = (state) => `
   <div class="navbar">
     <div class="navbar-wrap">
       <a class="title" href="/"><img src="/assets/logo.png" />
-      CardanoLite Wallet<sup>⍺</sup>
+      <span>CardanoLite Wallet</span><sup>⍺</sup>
```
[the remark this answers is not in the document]
mpinter: With how reconciliation works now, we'll probably need all of the text elements surrounded by div / span, as the 'text' nodes (strings without surrounding tags) are completely skipped (apart for the tags containing only text info, which is evaluated differently). Without that, we'd need to deal with whitespaces inside tags, which are evaluated the same way - as text nodes.

PR #37, opened 2018-03-29
Pershing redux

mpinter:
I've also brought in the immutability helpers (nice to have, no dependencies, except for node's assert which looks optional anyway).

Did a change in fdb2105, so that when no reducer is specified, payload is simply merged into current state  (similarly to how setState works) on a given path - seemed handy and made sense to me, but if it's too radical I can remove it again.


frontend/helpers.js
```diff
@@ -0,0 +1,120 @@
+import assert from 'assert'
```
mpinter: this brings in too many dependencies, will replace with a single function

refi93:
not needed anymore since we decided to migrate to preact

PR #38, opened 2018-04-05
Copy to clipboard next to address

git-jano:
display a "button" to copy address to clipboard


refi93:
I think you forgot to upload the copy.png file referenced in the css

refi93:
I realized that from the UX point of view you probably want to notify the user that the operation of copying was successful. Maybe turning green the button or showing some fading message? I was trying to come up with a simple solution (purely css) but it would probably be better to wait for Pinto's refactor, since you probably want to handle the state of the component somehow to do this.

git-jano:
change title to 'copied' (github style), or show tooltip for 3 sec ?

refi93:
I like it the github way, i.e. changing the title - no need to play with styles

public/css/styles.css
```diff
@@ -184,6 +184,52 @@ td {
 .address-wrap + button {
   margin-top: 1rem;
 }
+.copy{
```
mpinter: I'd let @darjanin step in to look over css, I'm not too sure about good manners here

frontend/components.js
```diff
@@ -94,24 +94,73 @@ const Balance = connect('balance')(({balance}) =>
...
+  render({address, isTransaction}, {copied}) {
+    return h(
+      'div',
+      {class: 'address-wrap'},
+      h('input', {readonly: true, type: 'text', class: 'address', value: address}),
+      h('span', {class: 'copy',
+        onclick: () => this.copyTextToClipboard(address),
+        onMouseLeave: () => this.setState({copied: false}),
+      }, '',
+      h('span', {class: 'tooltip'}, copied ? 'Copied' : 'Copy to clipboard'),
```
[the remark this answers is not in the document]
mpinter: Not sure honestly - I'm in for making it the same way everywhere, titles are certainly easier.., but I might even be in for enforcing tooltips like these everywhere ? Due to mobile compatibility where you don't have much place to show titles ?

frontend/components.js
```diff
@@ -94,24 +94,73 @@ const Balance = connect('balance')(({balance}) =>
...
+      ),
+      h('a', {
+        href: `https://cardanoexplorer.com/${isTransaction ? 'tx' : 'address'}/${address}`,
+        target: '_blank',
+        title: 'examine via CardanoExplorer.com',
```
mpinter: maybe capital leter ? 'Examine via...'

PR #41, opened 2018-04-11
Preact

mpinter:
Preact + unistore + new styling


public/css/styles.css
```diff
@@ -362,6 +452,140 @@ td {
...
+  font-size: 2.3rem;
+  font-weight: 600;
+  margin: 0 60px 30px;
+  text-align: center;
+}
+.intro-input-row {
+  display: flex;
+  justify-content: center;
+}
+.intro-button {
```
darjanin: This button positioned absolutely breaks on mobile widths.

darjanin: And what was the reason to use `position: absolute` for it?

PR #44, opened 2018-04-12
Validations and confirmation dialog

git-jano:


refi93:
when I go to the "send" tab, without performing any other action, the button under the form is grey with the spinner and message "Calculating Fee"

frontend/components.js
```diff
@@ -262,37 +305,46 @@ class SendAdaClass extends Component {
...
+            onInput: updateAmount,
+            autocomplete: 'nope',
           }),
-          h('span', {style: `color: ${feeRecalculating ? 'red' : 'green'}`}, `= ${totalAmount} ADA`)
+          (transactionFee > 0) ?
+            h('span', {style: `color: ${feeRecalculating ? 'red' : 'green'}`}, `= ${totalAmount} ADA`)
+            : '',
         ),
-        feeRecalculating
+        (feeRecalculating || sendAmount !== feeCalculatedFrom.amount || sendAddress !== feeCalculatedFrom.address)
```
refi93: Why so many conditions? Isn't the feeRecalculating variable enough? I tried removing the remaining conditions and the validation seems to work OK. Are there any corner cases I'm not aware of?

frontend/components.js
```diff
@@ -262,37 +305,46 @@ class SendAdaClass extends Component {
...
           ? h(
             'button',
             {disabled: true, class: 'loading-button'},
             h('div', {class: 'loading-inside-button'}),
             'Calculating Fee'
           )
-          : h('button', {onClick: submitTransaction, class: 'loading-button'}, 'Submit')
+          : h('button', {disabled: !(!!sendAmount && !!sendAddress && !sendAmountValidationMsg && !sendAddressValidationMsg), onClick: confirmTransaction, class: 'loading-button'}, 'Submit'),
```
tomaskulich: why !! ? does this actually do anything?

frontend/actions.js
```diff
@@ -98,58 +89,156 @@ export default ({setState, getState}) => {
...
-    const sendValue = state.sendAmountFieldValue
-    const amount = parseFloat(state.sendAmountFieldValue || 0)
-    const transactionFee = await wallet.getTxFee(address, amount)
-    // if we reverted value in the meanwhile, do nothing, otherwise update
-    if (state.sendAmountFieldValue !== state.feeCalculatedFrom) {
-      setState({transactionFee, feeCalculatedFrom: sendValue})
+    setState({calculatingFee: true})
+    clearSendAddressValidation(getState())
+    clearSendAmountValidation(getState())
+    const isValidAddress = validateSendAddress(getState())
```
tomaskulich: call getState() only once, store the result

frontend/actions.js
```diff
@@ -98,58 +89,156 @@ export default ({setState, getState}) => {
...
+        feeCalculatedFrom: {
+          amount: getState().sendAmount.fieldValue,
+          address: getState().sendAddress.fieldValue,
+        },
+      })
     }
+    setState({calculatingFee: false})
   }
 
+  const confirmTransaction = async () => ({showConfirmTransactionDialog: true})
```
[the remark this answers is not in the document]
git-jano: refactor (before it used to validate fields and recalculate fee)

frontend/components.js
git-jano: validation on GUI site, 
should not submit empty or invalid fields

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
+    sendAmount: Object.assign(state.sendAmount, {validationMsg: ''}),
+  })
+
+  const clearSendAddressValidation = (state) => ({
+    sendAddress: Object.assign(state.sendAddress, {validationMsg: ''}),
   })
 
-  // is being called debounced, thus we need to getState
-  // TODO maybe it'd be good practice to only use getState and avoid this problem?
+  const validateSendAddress = (state) => {
```
mpinter: Please don't export this as an action.

Actions should always return a state that is merged with a new one (or undefined, in which case ) - if you need helper functions for actions, like these, put them outside the exported object (maybe even into separate file ?

Same for the rest of validate fns.

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
+          validationMsg: `Insufficient funds. Your balance is ${state.balance / 1000000} ADA.`,
+        }),
+      })
+      return false
+    }
+    return true
+  }
+
+  const validateFee = (state) => {
+    if (parseFloat(state.transactionFee) < 0) {
```
mpinter: wat ? this can happen ?

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
+      setState({
+        transactionFee: -state.transactionFee,
+        sendAmount: Object.assign(state.sendAmount, {
```
mpinter: you can return instead of using setState, it's nicer, keep setState for async functions

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
-    const sendValue = state.sendAmountFieldValue
-    const amount = parseFloat(state.sendAmountFieldValue || 0)
-    const transactionFee = await wallet.getTxFee(address, amount)
-    // if we reverted value in the meanwhile, do nothing, otherwise update
-    if (state.sendAmountFieldValue !== state.feeCalculatedFrom) {
-      setState({transactionFee, feeCalculatedFrom: sendValue})
+    if (state.sendAddress.fieldValue !== '' && state.sendAmount.fieldValue !== '') {
+      setState({calculatingFee: true})
+      // if we reverted value in the meanwhile, do nothing, otherwise update
+      if (
```
mpinter: this check does not make sense where it is - you've only made sync calls so far

you need to check for this after calling `await wallet.getTxFee(address, amount)` and before setting state

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
+        state.sendAmount.fieldValue !== state.feeCalculatedFrom.amount ||
+        state.sendAddress.fieldValue !== state.feeCalculatedFrom.address
+      ) {
+        let transactionFee = 0
+        if (validateSendForm(state)) {
+          const address = state.sendAddress.fieldValue
+          const amount = parseFloat(state.sendAmount.fieldValue || 0) * 1000000
```
mpinter: no need to check after validating before

frontend/actions.js
```diff
@@ -98,58 +89,149 @@ export default ({setState, getState}) => {
...
+        }),
+      })
+      return false
+    }
+    return true
+  }
+
+  const validateSendForm = (state) => {
+    if (state.sendAddress.fieldValue !== '' && state.sendAmount.fieldValue !== '') {
+      clearSendAddressValidation(state)
```
mpinter: this is kind of ugly, first clearing and then maybe setting an error message

maybe do it in a way where the error message is a function of validateSend* function ? that function setting either empty string or an error to the error message field ?

frontend/components.js
```diff
@@ -55,12 +69,21 @@ class UnlockClass extends Component {
                 size: '47',
                 value: currentWalletMnemonicOrSecret,
                 onInput: this.updateMnemonic,
+                autocomplete: 'nope',
```
mpinter: lol .. but I see that it works :)

frontend/actions.js
git-jano: this is not exported

frontend/actions.js
git-jano: here it prevents unnecessary fee calculations, but understood

frontend/actions.js
refi93: The Cardano library function for fee computation returns a negative value (-1 * computed fee) if the fee cannot be applied (i.e. you don't have enough funds). Subsequently, this value can be used to estimate (but quite exactly for most practical purposes) the maximum sendable amount (you substract the fee from your balance).

Yes, it would probably be cleaner to have two variants of the function, one returning the computed fee no matter what and the other one throwing an exception if the fee cannot be applied, or simply to always return the computed fee and let the caller of the function do the math themself.

frontend/components.js
git-jano: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

frontend/actions.js
```diff
@@ -98,58 +89,133 @@ export default ({setState, getState}) => {
...
 
-  // is being called debounced, thus we need to getState
-  // TODO maybe it'd be good practice to only use getState and avoid this problem?
+  const validateSendAmount = (state) => {
+    let validationMsg = ''
+    const amount = parseFloat(state.sendAmount.fieldValue)
+    if (isNaN(amount)) {
+      validationMsg = 'Invalid format: Amount has to be a number'
+    }
+    if (amount.toFixed(6) <= 0) {
```
mpinter: this won't work (at least not as expected), .toFixed returns a string

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

please, do test all of the validadtions if you did not yet

frontend/actions.js
```diff
@@ -98,58 +89,133 @@ export default ({setState, getState}) => {
...
+      validationMsg = 'Invalid format: Amount has to be a positive number'
+    }
+    if (amount.toFixed(6) > state.balance / 1000000) {
+      validationMsg = `Insufficient funds. Your balance is ${state.balance / 1000000} ADA.`
+    }
+    setState({sendAmount: Object.assign(state.sendAmount, {validationMsg})})
```
mpinter: ..I think you've misunderstood Object.assign (and sorry for missing this in previous review)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

this would mutate state.sendAmount - what you've meant to do is Object.assign({}, state.sendAmount, {validationMsg})

frontend/components.js
```diff
@@ -205,41 +228,73 @@ const TransactionHistory = connect('transactionHistory')(({transactionHistory})
...
-      this.setState({
-        totalAmount: (parseFloat(next.sendAmountFieldValue || 0) + next.transactionFee).toFixed(6),
-      })
+    const displayTransactionFee = (sendAmount !== '') && !feeRecalculating &&
+      transactionFee > 0 &&
+      (!sendAmountValidationError || sendAmountValidationError === 'INSUFFICIENT_FUNDS_FOR_FEE')
+
+    const validationMsg = (errorCode) => {
+      const errorCodeMap = {
+        INVALID_ADDRESS: 'Invalid address',
```
refi93: this translations map belongs to a separate file imho. It would be nicer to have a "translations.js" file which would export the map with them. If you need to insert a value, you can put a closure in the map (eg INSUFFICIENT_FUNDS: (balance) => `Insufficient funds. Your balance is ${balance} ADA.` For consistency it would be nice to have all the translations as functions that by default take no parameters.

frontend/translations.js
```diff
@@ -0,0 +1,12 @@
+const strings = {
+  SendAddressInvalidAddress: () => 'Invalid address',
+  SendAmountIsNan: () => 'Invalid format: Amount has to be a number',
+  SenAmountIsNotPositive: () => 'Invalid format: Amount has to be a positive number',
+  SendAmountInsufficientFunds: ({balance}) => `Insufficient funds. Your balance is ${balance} ADA.`,
+  SendAmountInsufficientFundsForFee: ({balance, transactionFee}) => `Insufficient funds to cover the transaction fee. 
+    Cannot send more than ${balance - transactionFee} ADA.`,
+}
+
+const get = (key, props) => strings[key] ? strings[key](props) : ''
```
mpinter: this is not Java enterprise

just export the object

frontend/components.js
```diff
@@ -2,6 +2,7 @@ import {h, Component} from 'preact'
 import {connect} from 'unistore/preact'
 import Cardano from '../wallet/cardano-wallet'
 import actions from './actions'
+import {get as __} from './translations'
```
mpinter: when you export just the object with functions you can just import translations from './tranlsations'

sidenote: you should have a good reason to import something as '_' and you should never have reason to import it as '__'

public_wallet_app/css/styles.css
```diff
@@ -95,6 +95,15 @@ button.danger {
 button.danger:hover {
   background-color: #9c2929;
 }
+button.positive {
+  background-color: #38a14f;
+}
+button.positive:hover {
```
mpinter: we are no longer changing colors on most button hovers (expect the logout refresh buttons), just doing the slight size transform instead

public_wallet_app/css/styles.css
```diff
@@ -95,6 +95,15 @@ button.danger {
 button.danger:hover {
   background-color: #9c2929;
 }
+button.positive {
+  background-color: #38a14f;
```
mpinter: please don't define colors in place - ideally use one of those already in defined on top (look around styles.css on how it's done), and if you really need a new one add one there

PR #47, opened 2018-04-20
Css fixes

mpinter:


mpinter:
Still need to fix the mobile version menu in the intro screen (will be a cosmetic change, the rest is ready for review).

frontend/components.js
```diff
@@ -336,6 +332,57 @@ const TopLevelRouter = connect((state) => ({
...
+      if (navigator.clipboard) {
+        await navigator.clipboard.writeText(this.props.value)
+      } else {
+        this.fallbackCopyTextToClipboard()
+      }
+    } catch (err) {
+      console.error('Could not copy text: ', err)
+    } finally {
+      // TODO maybe show in better way, even for mobile
+      this.setState({tempTitle: 'Copied!'})
```
darjanin: It will not show this text in Safari. `onClick` the title will hide. In Chrome it appears again, but in Safari you have no idea what happened. Some sort of alert/notification with this message needs to show.

public/css/styles.css
```diff
@@ -298,51 +293,57 @@ td {
...
 }
 .status button:hover {
-  background-color: var(--brand-light);
+  color: var(--brand);
 }
-.status-text-wrapper {
-  display: flex;
-  flex-direction: column;
-  font-size: 0.6rem;
+.staus button:hover svg {
```
darjanin: `.status`, missing `t` after `a`.

frontend/components.js
```diff
@@ -350,31 +397,33 @@ const LoginStatus = connect(
...
+        'div',
+        {class: 'status-text'},
+        'Balance: ',
+        h('span', {class: 'status-balance'}, `${(balance / 1000000).toFixed(6)} ADA`)
+      ),
+      h(
+        'div',
+        {class: 'status-text', title: activeWalletId},
+        'WalletID: ',
+        h(CopyOnClick, {value: activeWalletId, title: activeWalletId})
```
refi93: I think wallet ids should not be copyable to discourage people from accidentally using them as an address since it is not recommended: https://cardanodocs.com/technical/hd-wallets/ "**Crucial point in design: root public keys are not used to actually store money**." The wallet id is nothing else than the address associated with the root public key. A more detailed explanation is provided in the paragraph above in the linked article.

I think it is enough to display the address as a title on hover

frontend/components.js
refi93: I see the "copied!" title only when I hover again over the link. I'm using Firefox 59.0.2 and the same happens on Chrome as well. If it is the expected behavior, I think it wouldn't serve its purpose since virtually nobody would hover again over the text. Nevertheless, as I mentioned in the comment below, the wallet id should not be copyable for security reasons. Janci in his PR 38 (https://github.com/vacuumlabs/cardano/pull/38) provides a "copy tooltip" that does not dissapear, but you probably already know that.

frontend/components.js
```diff
@@ -484,24 +550,54 @@ const AboutOverlay = connect('displayAboutOverlay', actions)(
...
+            'span',
+            {class: 'centered-row'},
+            h('input', {
+              type: 'checkbox',
+              checked: dontShowAgainCheckbox,
+              onClick: this.checkboxClick,
+              class: 'understand-checkbox',
+            }),
+            h(
+              'span',
```
darjanin: How about using `label` instead? It will change value of `input` if it will be wrapped in it. You can something like `onChange` on input then.

public/css/styles.css
```diff
@@ -434,36 +455,23 @@ td {
...
   justify-content: center;
+  align-items: center;
 }
-.amount-label-row {
-  display: flex;
-  justify-content: space-between;
+.understand-checkbox {
+  margin: 0.6rem;
+}
+.checkbox-text {
```
darjanin: If you will use `label`, you don't need to set the `cursor: pointer`.

frontend/components.js
refi93: I probably like the most the second option, i.e. taking the shortened SHA2 hash as the displayed "wallet id". Shortening the hash should be fine: https://security.stackexchange.com/questions/97377/secure-way-to-shorten-a-hash?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

mpinter: Well, the other thing is that dropping it completely (or moving to some less promiment place) woudl leave more place to display balance on top - which is a more valuable information, and its currently competing for space.

darjanin:
screen shot 2018-04-23 at 13 19 48

Title is wrapped. It's this on purpose? I would make the text smaller and fix the alpha symbol wrong position.

darjanin:
screen shot 2018-04-23 at 13 21 24

It makes more sense to have the label the longer text: "Receiving address" and the placeholder would be only "Address".

darjanin:
screen shot 2018-04-23 at 13 21 46

This text and note above it are distracting with different margin from right.

darjanin:
screen shot 2018-04-23 at 13 20 09

I would make the border little bit longer to the left. You can achieve this look by adding left padding.

frontend/components.js
refi93: Or let's just drop it completely for now. You can use as a reference some of the used addresses in the "send" tab.

mpinter:
I fixed what I could with reasonable time investment, will create issues for the unresolved ones.

PR #48, opened 2018-04-23
Static landing page

mpinter:
Already running in production


refi93:
I suggest renaming the "static" folder to "landing_page" and "public" folder to "wallet_app_public" to make clear the purpose of those folders

PR #49, opened 2018-04-25
Basic stats via redis

mpinter:
You'll need to install redis locally after this is merged for the app to work


mpinter:
@ppershing the idea is that we don't want to store IP adresses in any form as it is part of the promise (or, so to say, 'business model') of the app - keeping them only in application memory is probably good enough to fulfill this (you can't really avoid that either way)

that said, this is quite a naive approach, with not too much thought behind it - should do the job, but I'm ok with scrapping it and doing it differently if anyone comes with a better idea

ppershing:
@mpinter I see, fair enough. While hyperloglog doesn't actually store the original values only (some) hashes, IPs are too low-entropy to fulfill the whole promise as we would be able to probabilistically probe whether an IP was used or not. As with storing in memory -- maybe we should be clear that we can keep IPs in memory (not disk) for statistics. The other things that come to my mind
- do not count unique IPs but page visits (probably the cleanest solution but I am not sure we are interested in this metric)
- we could generate long random session cookies and use this instead of IPs for hyperloglog. Downside is that some people might be paranoid that we set cookies.

refi93:
After running yarn install and yarn dev I get this error:

```
/home/rafael/git_veci/cardano/server/middleware/stats.js:1
(function (exports, require, module, __filename, __dirname) { import redis from 'redis'
                                                              ^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:152:10)
    at Module._compile (module.js:624:28)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Module.require (module.js:604:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/rafael/git_veci/cardano/server/index.js:4:25)

Webpack is watching the files…
...
```

I have node version 9.3. I'm I doing something wrong? Seems like my node does not like the ES6 syntax of imports

mpinter:
Ok, should be good now

PR #55, opened 2018-04-30
Simplify transaction validation behavior

refi93:
II simplified the validation behavior so that the transaction fee does not disapper when computing it but it turns red and stays like that until the transaction fee computation passes.

I tried also to avoid the disappearing of the validation message when the address changes but I could not find a solution that I would be satisfied with. Feel free to suggest any.

### DIFF
--- a/frontend/actions.js
+++ b/frontend/actions.js
@@ -111,12 +111,10 @@ export default ({setState, getState}) => {
 
     if (isNaN(amount) || strAmount.toLowerCase().includes('e')) {
       validationError = {code: 'SendAmountIsNan'}
-    }
-    if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
-      validationError = {code: 'SendAmountPrecisionLimit'}
-    }
-    if (amount <= 0) {
+    } else if (amount <= 0) {
       validationError = {code: 'SendAmountIsNotPositive'}
+    } else if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
+      validationError = {code: 'SendAmountPrecisionLimit'}
     }
 
     setState({sendAmount: Object.assign({}, state.sendAmount, {validationError})})
--- a/frontend/actions.js
+++ b/frontend/actions.js
@@ -95,23 +95,30 @@ export default ({setState, getState}) => {
     }
   }
 
-  const validateSendAddress = (state) => setState({
-    sendAddress: Object.assign({}, state.sendAddress,
-      {validationError: !Cardano.isValidAddress(state.sendAddress.fieldValue) ? {code: 'SendAddressInvalidAddress'} : ''}),
-  })
+  const validateSendAddress = (state) =>
+    setState({
+      sendAddress: Object.assign({}, state.sendAddress, {
+        validationError: !Cardano.isValidAddress(state.sendAddress.fieldValue)
+          ? {code: 'SendAddressInvalidAddress'}
+          : '',
+      }),
+    })
 
   const validateSendAmount = (state) => {
     let validationError = ''
+    const strAmount = state.sendAmount.fieldValue
     const amount = parseFloat(state.sendAmount.fieldValue)
+
     if (isNaN(amount)) {
       validationError = {code: 'SendAmountIsNan'}
     }
+    if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
+      validationError = {code: 'SendAmountPrecisionLimit'}
+    }
     if (parseFloat(amount.toFixed(6)) <= 0) {
-      validationError = {code: 'SenAmountIsNotPositive'}
+      validationError = {code: 'SendAmountIsNotPositive'}
     }
-    if (parseFloat(amount.toFixed(6)) > state.balance / 1000000) {
-      validationError = {code: 'SendAmountInsufficientFunds', params: {balance: state.balance / 1000000}}
-    }
+
     setState({sendAmount: Object.assign({}, state.sendAmount, {validationError})})
   }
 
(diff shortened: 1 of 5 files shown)


wallet/cardano-wallet.js
```diff
@@ -193,10 +193,9 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
     let result
 
     if (usedAddressesAndSecrets.length < usedAddressesLimit) {
-      const highestUsedChildIndex =
-        usedAddressesAndSecrets.reduce((acc, item) => {
-          return Math.max(item.childIndex, acc)
-        }, 0x80000000)
+      const highestUsedChildIndex = usedAddressesAndSecrets.reduce((acc, item) => {
```
ppershing: Is this equivalent to ```Math.max(0x8000000, ...usedAddressesAndSecrets.map(item => item.childIndex))```

frontend/actions.js
```diff
@@ -95,23 +95,30 @@ export default ({setState, getState}) => {
...
     let validationError = ''
+    const strAmount = state.sendAmount.fieldValue
     const amount = parseFloat(state.sendAmount.fieldValue)
+
     if (isNaN(amount)) {
       validationError = {code: 'SendAmountIsNan'}
     }
-    if (parseFloat(amount.toFixed(6)) <= 0) {
-      validationError = {code: 'SenAmountIsNotPositive'}
+    if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
```
ppershing: I am not sure if there is another validation somewhere else but a user can trick this by using
`1.0e-8` as as `strAmount`

frontend/actions.js
```diff
@@ -95,23 +95,30 @@ export default ({setState, getState}) => {
...
+      validationError = {code: 'SendAmountPrecisionLimit'}
     }
-    if (parseFloat(amount.toFixed(6)) > state.balance / 1000000) {
-      validationError = {code: 'SendAmountInsufficientFunds', params: {balance: state.balance / 1000000}}
+    if (parseFloat(amount.toFixed(6)) <= 0) {
```
ppershing: Is this different from `amount < 1e-6`?
Also, supplying `0.0000000001` as `strAmount` will overwrite previous `validationError` to something less expected by the user (e.g., `SendAmountIsNotPositive` error seems weird for positive values < 1e-6)

frontend/actions.js
```diff
@@ -95,23 +95,30 @@ export default ({setState, getState}) => {
...
+    setState({
+      sendAddress: Object.assign({}, state.sendAddress, {
+        validationError: !Cardano.isValidAddress(state.sendAddress.fieldValue)
+          ? {code: 'SendAddressInvalidAddress'}
+          : '',
+      }),
+    })
 
   const validateSendAmount = (state) => {
     let validationError = ''
```
ppershing: I know this is an older design decision but I am a not big fan of polymorphism of `string` vs `dict`. Is there a good reason not to use `null` for non-error scenario?

frontend/actions.js
```diff
@@ -122,20 +129,29 @@ export default ({setState, getState}) => {
...
+    if (sendAmountForTransactionFee + transactionFee > balance) {
+      setState({
+        sendAmount: Object.assign({}, state.sendAmount, {
+          validationError: {code: 'SendAmountInsufficientFunds', params: {balance: state.balance}},
+        }),
+      })
     }
   }
 
+  const isSendFormFilledAndValid = (state) => {
```
ppershing: Is there any reason we use `return` instead of the shorthand notation
`(state) => (... && ... && ... && ...)` ?

frontend/actions.js
```diff
@@ -155,6 +171,7 @@ export default ({setState, getState}) => {
     ) {
       setState({
         transactionFee,
+        sendAmountForTransactionFee: amount,
```
ppershing: I see this value is not guaranteed to be an `int` -- doesn't something downstream break if `sendAmount.fieldValue` is a bit nasty (though I am not sure such nasty value passing validation readily exists)?

frontend/actions.js
refi93: Yes, you are right, i forgot about this case. I added a check for the letter "e" above, so any scientific notation would be rejected as an invalid input.

frontend/actions.js
refi93: fixed - amount is now converted to int explicitly to Int above

refi93:
Currently all the amounts used in actions.js should be ints, but some arithmetics are done in components.js too. This should be fixed in the future.

refi93:
Yes, I don't like as well that we are working with state.sendAmount.fieldValue everywhere. I believe it should be parsed just at one point in the code. It would be nice to refactor it.

wallet/cardano-wallet.js
refi93: fixed

frontend/actions.js
refi93: fixed

frontend/actions.js
refi93: there isn't. I prefer nulls, too, so I changed it. Fixed

frontend/actions.js
refi93: fixed

frontend/translations.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
   SendAddressInvalidAddress: () => 'Invalid address',
   SendAmountIsNan: () => 'Invalid format: Amount has to be a number',
-  SenAmountIsNotPositive: () => 'Invalid format: Amount has to be a positive number',
-  SendAmountInsufficientFunds: ({balance}) => `Insufficient funds. Your balance is ${balance} ADA.`,
-  SendAmountInsufficientFundsForFee: ({balance, transactionFee}) => `Insufficient funds to cover the transaction fee. 
-    Cannot send more than ${balance - transactionFee} ADA.`,
+  SendAmountIsNotPositive: () => 'Invalid format: Amount has to be a positive number',
+  SendAmountInsufficientFunds: ({balance}) =>
+    `Insufficient funds for the transaction. Your balance is ${balance / 1000000} ADA.`,
```
ppershing: Could the division lead to some weird values? A true solution should be (assuming positive balance)
`${Math.floor(balance/1000000)} . ${figure_out_how_to_pad_and_strip_zeroes(balance % 100000)}`
but I guess for now a TODO comment would do

frontend/actions.js
```diff
@@ -95,23 +95,28 @@ export default ({setState, getState}) => {
...
-    if (isNaN(amount)) {
+    let validationError = null
+    const strAmount = state.sendAmount.fieldValue
+    const amount = parseFloat(strAmount)
+
+    if (isNaN(amount) || strAmount.toLowerCase().includes('e')) {
       validationError = {code: 'SendAmountIsNan'}
+    } else if (amount <= 0) {
+      validationError = {code: 'SendAmountIsNotPositive'}
+    } else if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
```
ppershing: I guess we should first check precision and only then if the truncated number if positive

PR #60, opened 2018-05-02
refactor address.js and input validation

refi93:
I made the suggested fixes mentioned in https://github.com/vacuumlabs/cardano/pull/55 and a small refactor of some functions in address.js which we talked about a while ago.

### DIFF
--- a/frontend/actions.js
+++ b/frontend/actions.js
@@ -1,5 +1,6 @@
 import Cardano from '../wallet/cardano-wallet'
 import {CARDANOLITE_CONFIG} from './config'
+import {sendAddressValidator, sendAmountValidator, feeValidator} from './validators'
 
 let wallet = null
 
@@ -79,10 +80,7 @@ export default ({setState, getState}) => {
   const generateNewUnusedAddress = async (state) => {
     setState({address: 'loading...'})
     const offset = state.unusedAddresses.length
-    const newUnusedAddress = await wallet.getChangeAddress(
-      Number.MAX_SAFE_INTEGER,
-      offset
-    )
+    const newUnusedAddress = await wallet.getChangeAddress(Number.MAX_SAFE_INTEGER, offset)
     setState({
       unusedAddresses: state.unusedAddresses.concat([newUnusedAddress]),
     })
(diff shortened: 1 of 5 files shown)


wallet/address.js
```diff
@@ -87,23 +87,22 @@ async function deriveAddressAndSecret(rootSecretString, childIndex) {
     addressPayload = new Buffer(0)
     addressAttributes = new Map()
     derivedSecretString = rootSecretString
-    addressRoot = new Buffer(getAddressRoot(derivedSecretString, addressPayload), 'hex')
   } else {
     // the remaining addresses
     const hdPassphrase = await deriveHDPassphrase(rootSecretString)
     derivedSecretString = deriveSK(rootSecretString, childIndex)
```
ppershing: please move this line to the end of the block so that it is consistent with the previous if branch

wallet/address.js
```diff
@@ -22,7 +22,7 @@ function addressHash(input) {
   const context = blake2.blake2bInit(28) // blake2b-224
   blake2.blake2bUpdate(context, firstHash)
 
-  return new Buffer(blake2.blake2bFinal(context)).toString('hex')
```
ppershing: I hope this is covered by the tests

frontend/actions.js
```diff
@@ -111,32 +108,28 @@ export default ({setState, getState}) => {
...
-    if (isNaN(amount) || strAmount.toLowerCase().includes('e')) {
+    if (!floatRegex.test(strAmount) || isNaN(amount)) {
       validationError = {code: 'SendAmountIsNan'}
-    } else if (amount <= 0) {
-      validationError = {code: 'SendAmountIsNotPositive'}
     } else if (strAmount.split('.').length === 2 && strAmount.split('.')[1].length > 6) {
       validationError = {code: 'SendAmountPrecisionLimit'}
+    } else if (amount <= 0) {
+      validationError = {code: 'SendAmountIsNotPositive'}
     }
```
ppershing: is there any limit on how big sendAmount can be in cardano?

wallet/address.js
refi93: yes. The address encoding test would fail if this returned wrong output

git-jano: are thore bugs or features: ? 
- `1.` is treated as invalid num (any num + `.` 
- if user tires to mess up / del address, the amount validations will  disappear, and red total will be actualized no more until a valid address is typed

refi93: @git-jano "1." passes the validation - that's intended behavior. The fee staying red and submit button being disabled  while the address is missing/invalid is also intended.

refi93: I was wondering why the "1." actually passes the validation despite the regex and I found out that the HTML5 input element sanitizes the input (at least in chrome and firefox) and strips the comma if it is not followed by any number, therefore the value "1." is transformed into "1" before being fed to Javascript. Finally I decided to make the "send-amount" input a standard input and handle the validation just in javascript to avoid such confusion and to have absolute control over all the cases.

PR #64, opened 2018-05-11
Rename SecretString to hdNode

### LINKED ISSUE #61: Refactor `secretString` to `hdnode`
Currently, variable names such as `walletSecretString` are quite misleading because they hold a complex type instead of `string`. Trezor uses `HDnode` for the equivalent variables. One other option would be to simply use `node`

git-jano:
cleanup names
trezor like naming convention
closes #61


wallet/address.js
```diff
@@ -79,24 +79,24 @@ function isValidAddress(address) {
...
   } else {
     // the remaining addresses
-    const hdPassphrase = await deriveHDPassphrase(rootSecretString)
+    const hdPassphrase = await deriveHdPassphrase(parentHdNode)
     const derivationPath = [0x80000000, childIndex]
 
     addressPayload = encryptDerivationPath(derivationPath, hdPassphrase)
     addressAttributes = new Map([[1, cbor.encode(addressPayload)]])
-    derivedSecretString = deriveSK(rootSecretString, childIndex)
+    derivedHdNode = deriveSk(parentHdNode, childIndex)
```
ppershing: I guess we can rename `deriveSk` -> `deriveHdNode`

wallet/cardano-wallet.js
```diff
@@ -188,37 +185,38 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
...
-    if (usedAddressesAndSecrets.length < usedAddressesLimit) {
+    if (usedAddressesWithHdNodes.length < usedAddressesLimit) {
       const highestUsedChildIndex = Math.max(
         0x8000000,
-        ...usedAddressesAndSecrets.map((item) => item.childIndex)
+        ...usedAddressesWithHdNodes.map((item) => item.childIndex)
       )
 
-      result = (await getAddressAndSecret(highestUsedChildIndex + 1 + offset)).address
+      result = (await getAddressWithHdNode(highestUsedChildIndex + 1 + offset)).address
```
ppershing: I think we should assert `offset >= 0` to avoid any weird situations when somebody sends us wrong argument

wallet/cardano-wallet.js
```diff
@@ -188,37 +185,38 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
...
     } else {
       result =
-        usedAddressesAndSecrets[Math.floor(Math.random() * usedAddressesAndSecrets.length)].address
+        usedAddressesWithHdNodes[Math.floor(Math.random() * usedAddressesWithHdNodes.length)]
```
ppershing: Just a question -- pseudorandom generator is safe here?

wallet/cardano-wallet.js
```diff
@@ -188,37 +185,38 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
...
+    const addresses = await getUsedAddressesWithHdNodes()
 
     for (let i = 0; i < addresses.length; i++) {
       const addressUnspentOutputs = await blockchainExplorer.getUnspentTxOutputs(
         addresses[i].address
       )
 
       addressUnspentOutputs.map((element) => {
-        element.secret = addresses[i].secret
+        element.hdNode = addresses[i].hdNode
```
ppershing: uhh, this is a bit of a black magic here. First of all, I guess we should use `forEach` instead of `map` as we modify existing elements and don't really care about returned array either.
Moreover, the outer for can be similarly rewritten as `forEach`.
What do you think?

wallet/cardano-wallet.js
```diff
@@ -255,24 +253,24 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
...
-      result.push(await getAddressAndSecret(0x80000001 + i))
+      result.push(await getAddressWithHdNode(0x80000001 + i))
     }
 
     return result
   }
 
   async function getUsedAddresses() {
-    return (await getUsedAddressesAndSecrets()).map((item) => {
+    return (await getUsedAddressesWithHdNodes()).map((item) => {
```
ppershing: `(item) => item.address`

wallet/cardano-wallet.js
```diff
@@ -255,24 +253,24 @@ const CardanoWallet = (secretOrMnemonic, CARDANOLITE_CONFIG) => {
...
-  async function deriveAddressesAndSecrets(
+  async function deriveAddressesWithHdNodes(
     begin = 0,
     // eslint-disable-next-line no-undef
     end = CARDANOLITE_CONFIG.CARDANOLITE_ADDRESS_RECOVERY_GAP_LENGTH
   ) {
     const result = []
     for (let i = begin; i < end; i++) {
-      result.push(await getAddressAndSecret(0x80000001 + i))
+      result.push(await getAddressWithHdNode(0x80000001 + i))
```
ppershing: Interesting alternative might be
`await Promise.all(_.range(begin, end).map(async (i) => await getAddress()))`

or
```
range = _.range(begin + 0x8....1, end + 0x8...1)
await Promise.all(range.map(getAddressWith...()))
```

But I leave it open as to what is the most readable option

wallet/transaction.js
```diff
@@ -160,21 +160,21 @@ class WalletAddress {
...
-  constructor(secretString) {
-    this.secretString = secretString
+class HdNode {
+  constructor(hdNodeString) {
+    this.hdNodeString = hdNodeString
   }
 
   getSecretKey() {
-    return this.secretString.substr(0, 128)
+    return this.hdNodeString.substr(0, 128)
```
ppershing: Is there any specific reason not to do these in the constructor?

refi93: not really, we can rewrite it. The original idea was to keep the "hdNodeString" as the only source of truth and to derive all the remaining information, but in hindsight I agree that it would probably be cleaner to split the "hdNodeString" in the constructor and we can save those getters

wallet/address.js
git-jano: I've noticed, makes sense, but then also `deriveSkIteration -> deriveHdNodeIteration`

wallet/cardano-wallet.js
git-jano: It's just a simple heuristics, if You run off the new address pool, pick one used randomly, true rnd won't help it much.

wallet/cardano-wallet.js
git-jano: Ok, might be more readable that way

wallet/cardano-wallet.js
git-jano: Of course

wallet/cardano-wallet.js
git-jano: for ... of  suits better

wallet/cardano-wallet.js
git-jano: fixed a few lines up

PR #66, opened 2018-05-15
Issue 62 cleanup refactoring

### LINKED ISSUE #62: Cleanup/refactoring
Small cleanups:
- we can streamline `Buffer` usage from `new Buffer('02', 'hex')` --> `new Buffer([0x02])` for one-byte buffers 
- more 'hex' reducements, e.g., `new Buffer(hmac1.digest('hex'), 'hex')` should not need double conversion through hex
- refactor `walletSecretString.getPublicKey()` to return buffer instead of hex (most of the time when we use it we get `Buffer` out of it back). Again, this will save some hex-double-conversion madness

git-jano:
closes #62 

- [x] we can streamline Buffer usage from new Buffer('02', 'hex') --> new Buffer([0x02]) for one-byte buffers
- [x] more 'hex' reducements, e.g., new Buffer(hmac1.digest('hex'), 'hex') should not need double conversion through hex
  - but it does
- [x] refactor walletSecretString.getPublicKey() to return buffer instead of hex (most of the time when we use it we get Buffer out of it back). Again, this will save some hex-double-conversion madness
  - keeping both, elliptic curve verification needs str
- [x] deriveSK is never used with childIndex = 0x80000000, change if -> assert
- [x]  Suggestion: replace function getCheckSum(input) in address.js with const crc32Unsigned = (inp) => crc32.buf(inp) >>> 0;
- [x]  we can streamline blake2 usage (according to https://github.com/dcposch/blakejs oneliner should work)

### DIFF
--- a/wallet/address.js
+++ b/wallet/address.js
@@ -149,10 +149,7 @@ function deriveHdNode(hdNode, childIndex) {
 }
 
 function getAddressRoot(hdNode, addressPayload) {
-  const extendedPublicKey = Buffer.concat([
-    hdNode.getPublicKeyBuffer(),
-    hdNode.getChainCodeBuffer(),
-  ])
+  const extendedPublicKey = hdNode.getExtendedPublicKey()
 
   return addressHash([
     0,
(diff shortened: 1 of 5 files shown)


refi93:
I would drop the "Buffer" suffix from the function names and work with buffers by default.

wallet/address.js
```diff
@@ -18,11 +18,7 @@ function addressHash(input) {
   const serializedInput = cbor.encode(input)
 
   const firstHash = new Buffer(sha3.sha3_256(serializedInput), 'hex')
-
-  const context = blake2.blake2bInit(28) // blake2b-224
-  blake2.blake2bUpdate(context, firstHash)
-
-  return new Buffer(blake2.blake2bFinal(context))
+  return new Buffer(blake2.blake2b(firstHash, null, 28))
```
ppershing: doesn't `blake2b` already return buffer?

wallet/address.js
```diff
@@ -183,33 +182,34 @@ function decryptDerivationPath(addressPayload, hdPassphrase) {
...
 }
 
-function getCheckSum(input) {
-  return crc32.buf(input) >>> 0
-}
+const crc32Unsigned = (input) => crc32.buf(input) >>> 0
 
 async function deriveHdPassphrase(hdNode) {
-  const extendedPublicKey = new Buffer(hdNode.getPublicKey() + hdNode.getChainCode(), 'hex')
+  const extendedPublicKey = Buffer.concat([
```
ppershing: why not make `getExtendedPublicKey` a method of `hdnode` ?

wallet/transaction.js
```diff
@@ -163,22 +157,32 @@ class WalletAddress {
...
 
+  getPublicKeyBuffer() {
+    return this.publicKeyBuffer
+  }
+
   getChainCode() {
     return this.chainCode
   }
+
+  getChainCodeBuffer() {
```
ppershing: Is there a particular reason to keep both buffer and hex versions?

wallet/helpers/hashBlake2b256.js
```diff
@@ -2,8 +2,5 @@ const blake2 = require('blakejs')
 const cbor = require('cbor')
 
 module.exports = function hashBlake2b256(input) {
-  const context = blake2.blake2bInit(32)
-  blake2.blake2bUpdate(context, cbor.encode(input))
-
-  return new Buffer(blake2.blake2bFinal(context))
+  return new Buffer(blake2.blake2b(cbor.encode(input), null, 32))
```
ppershing: see my comment above

wallet/address.js
git-jano: ok

wallet/address.js
git-jano: Returns a n-byte Uint8Array

wallet/address.js
git-jano: This is the only occurrence. I saw no profit,  but it runs many times, so ok

wallet/helpers/hashBlake2b256.js
git-jano: see my comment above

PR #76, opened 2018-05-24
Send ADA confirmation styling

mpinter:
Styling of confirmation dialog, more in tune with the overall page design


refi93:
screenshot from 2018-05-24 21-04-54

mpinter:
Sorry, tested chrome only, thanks for the catch - should be good now.

PR #87, opened 2018-06-20
trezor gui integration

git-jano:
login via trezor, 
verify address on trezor device screen


frontend/actions.js
```diff
@@ -121,7 +121,7 @@ export default ({setState, getState}) => {
 
     const address = state.sendAddress.fieldValue
     const amount = state.sendAmount.coins
-    const transactionFee = await wallet.getTxFee(address, amount)
+    const transactionFee = 1 //await wallet.getTxFee(address, amount)
```
mpinter: don't forget about this change

frontend/actions.js
```diff
@@ -26,67 +26,101 @@ export default ({setState, getState}) => {
...
+        try {
+          wallet = await Cardano.CardanoWallet({
+            cryptoProvider: 'trezor',
+            config: CARDANOLITE_CONFIG,
+          })
+        } catch (e) {
+          return setState({
+            loading: false,
+            walletLoagingError:
+              'Trezor authentication rejected! Check your Trezor firmware version.',
```
ddeath: Just wanted  to point out that Trezor popup will show error message if you are using trezor with firmware which is not supported.

screenshot from 2018-06-28 10-39-06

git-jano: ok, then this is temp, 
currently we do not ask for firmware version, since there is no compatible version. 
Communication just fails... . (no info in popup)

PR #88, opened 2018-06-21
load demo wallet button

git-jano:


refi93:
Nice, but I would prefer if after pressing the "Try demo wallet" button the wallet did not load immediately, but just the wallet mnemonic appeared in the input field, for the experience to be more similar to loading the actual wallet.

mpinter:
@refi93 I don't know, is that necessary ? Don't care much either way, but there isn't much experience in the extra button click :)

refi93:
@mpinter - I believe leaving the mnemonic on the screen would give the user a clue where the wallet is loaded from, otherwise he might get confused, since he did not get this information, i.e. experience, from the demo

refi93:
Maybe a flashing "Go" button in the demo mode with the prefilled mnemonic field to make the user click on that button or showing a bubble over it - whatever is simpler, would do it. We can discuss it.

ddeath:
I added pulsing animation to the button when it is active

refi93:
@ddeath the pulsing animation is nice, but I think it should stop if you change the mnemonic in the input field, so I would disable the pulsing on the update of the input field

refi93:
I would add debounced validation of the mnemonic (similar as when validating send amount) and the "go" button would be gray until a valid mnemonic is not inserted

refi93:
I refactored it to extract setState() calls to components.js and the error now shows only on blur and hides as soon as you start to type. Can you please have a look at it, @ddeath ?

frontend/actions.js
```diff
@@ -26,32 +31,71 @@ export default ({setState, getState}) => {
...
+  }
 
-    const walletIsLoaded = true
-    const ownAddresses = await wallet.getOwnAddresses()
-    const transactionHistory = await wallet.getHistory()
-    const balance = await wallet.getBalance()
-    const sendAmount = {fieldValue: ''}
-    const sendAddress = {fieldValue: ''}
-    const sendResponse = ''
+  const updateMnemonic = (state, e) => {
```
ddeath: Instead of `e` I would use value directly. If we change input for something else, then `e.target.value` will not be valid anymore.

frontend/actions.js
```diff
@@ -26,32 +31,71 @@ export default ({setState, getState}) => {
...
+        mnemonicValidationError: {
+          code: 'WalletInitializationError',
+        },
+      })
+    }
+  }
+
+  const loadDemoWallet = (state) => {
+    setState({
+      mnemonic: 'civil void tool perfect avocado sweet immense fluid arrow aerobic boil flash',
```
ddeath: I would change this as you suggested - get it from `CARDANOLITE_CONFIG` so developers can change it during development

frontend/actions.js
refi93: I would keep it simple - it would be cleaner to pass directly the value to the action but I think that making custom events directly in the components isn't too nice either. I think it's one of the tradeoffs we have to deal with for using Preact

PR #90, opened 2018-06-25
Add trezor crypto provider

ddeath:


wallet/cardano-trezor-crypto-provider.js
```diff
@@ -0,0 +1,370 @@
...
+      }
+
+      return false
+    })
+
+    if (cachedAddress) {
+      return true
+    }
+
+    try {
```
refi93: I have a suspicion that this would trigger a popup asking for public key after submitting the transaction since the update of unspent outputs list (updateUtxosFromTxAux) uses the isOwnAddress() call to add and remove addresses from the utxo list. For now I would remove this piece of code, since "own" addresses are now actually only those 10 which we manage and any other address, despite being derivable, should not be touched from our wallet to avoid confusion. Imagine for example sending funds to an address generated by daedalus for a wallet with the same mnemonic - it would mess up our list of unspent outputs vs the balance after performing the transaction.

We should make this modification in mnemonic provider as well, to preserve consistency, since now it considers as "own" also all derivable addresses, as well as the implementation there.

wallet/cardano-trezor-crypto-provider.js
```diff
@@ -0,0 +1,370 @@
...
+    if (!state.rootHdPassphrase) {
+      const rootHdPassphrase = (await deriveTrezorXpub([])).root_hd_passphrase
+      state.rootHdPassphrase = new Buffer(rootHdPassphrase, 'hex')
+    }
+
+    return state.rootHdPassphrase
+  }
+
+  async function getDerivationPathFromAddress(address) {
+    const cachedAddress = Object.values(state.derivedAddresses).find((record) => {
```
refi93: i would simplify this to:`Object.values(state.derivedAddresses).find((record) => record.address === address)`

wallet/cardano-trezor-crypto-provider.js
```diff
@@ -0,0 +1,370 @@
...
+  async function deriveXpubNonHardened(derivationPath) {
+    const parentPath = derivationPath.slice(0, derivationPath.length - 1)
+    const childPath = derivationPath.slice(derivationPath.length - 1, derivationPath.length)
+
+    // this reduce ensures that this would work even for empty derivation path
+    return childPath.reduce(derivePublic, await deriveXpub(parentPath, 'hardened'))
+  }
+
+  async function isOwnAddress(address) {
+    const cachedAddress = Object.values(state.derivedAddresses).find((record) => {
```
refi93: I would simplify this to: `Object.values(state.derivedAddresses).find((record) => record.address === address)`

wallet/cardano-trezor-crypto-provider.js
```diff
@@ -0,0 +1,370 @@
...
+
+    const outputs = []
+    for (const output of unsignedTx.outputs) {
+      const data = await _prepareOutput(output)
+      outputs.push(data)
+    }
+
+    const attributes = []
+
+    return await new Promise((resolve, reject) => {
```
refi93: this await is probably redundant. The calling function should already call await on this promise

refi93:
Seems OK for now, put on hold until cardano support is actually added to trezor.

wallet/cardano-trezor-crypto-provider.js
```diff
@@ -0,0 +1,334 @@
...
+      for (const derivationPath of derivationPaths) {
+        newAddresses.push(await deriveAddress(derivationPath, derivationMode))
+      }
+    }
+
+    return addresses.concat(newAddresses)
+  }
+
+  function fromBip44Path(path) {
+    switch (path.length) {
```
refi93: this can be generalized like this:

```
if ([0,3,5].includes(path.length)) {
    return path.filter((elem, i) => [2,4].includes(i))
} else {
    throw Error(`Unsupported derivation path: ${path}`)
}
```

ddeath: Yes it can, but I am not sure if it is also more readable a easy to understand.

refi93: ok I agree

PR #94, opened 2018-06-29
Fix chunking of fetching utxo

ddeath:


wallet/cardano-wallet.js
```diff
@@ -186,9 +186,21 @@ const CardanoWallet = (mnemonicOrHdNodeString, CARDANOLITE_CONFIG, randomSeed) =
 
     if (state.overallTxCountSinceLastUtxoFetch < currentOverallTxCount) {
       const nonemptyAddresses = await blockchainExplorer.selectNonemptyAddresses(addresses)
-      const response = await blockchainExplorer.fetchUnspentTxOutputs(nonemptyAddresses)
+      const chunks = range(0, Math.ceil(nonemptyAddresses.length / 10))
```
refi93: I would put this chunking logic directly into the fetchUnspentTxOutputs function, as well as the optimization with filtering out empty addresses. The caller (CardanoWallet) should not be aware of this "low level" optimizations since they don't really matter for it.

refi93:
you can also remove the unused libs at the top of the file - "cbor" and "cbor-parsers"

PR #96, opened 2018-07-02
Fix overflows on mobile

ddeath:
Transaction history table was overflowing and forced whole page to scroll... Same thing for addresses in "receive" tab


refi93:
for bigger balances the problem seems to persist. Try a balance for example "12345.123456". I also suggest making the balance displayed on mobile screens smaller, IMHO its too big currently.

ddeath:
Added text wrapping for balance.

PR #97, opened 2018-07-02
Paper wallet support

refi93:
Now you can insert into the input the 27 word paper wallet mnemonic and the wallet will get properly loaded. The proper way of working is shown in the tests (the "mnemonic.js" one). You can try loading the paper wallet string from the tests and see that the wallet was loaded by checking for the presence of the address "DdzFFzCqrhssvx7eAtDT2xbQtVeRVjCktAEg4sc3gmSFE6RFBadLVZX2e8ByfawtASMBeUtNMsyauaSSG2TP3Rgv3V94xLzmt67fJG4z" between the receive addresses. You can also generate the paper wallet yourself in Daedalus and verify that the first receive address to appear in cardanolite matches the one from the PDF generated by Daedalus.

### DIFF
--- a/frontend/components.js
+++ b/frontend/components.js
@@ -275,7 +275,7 @@ class UnlockClass extends Component {
             class: 'styled-input-nodiv styled-unlock-input',
             id: 'mnemonic-submitted',
             name: 'mnemonic-submitted',
-            placeholder: 'Enter twelve-word mnemonic',
+            placeholder: 'Enter 12-word mnemonic or paper wallet mnemonic',
             value: mnemonic,
             onInput: updateMnemonic,
             onBlur: checkForMnemonicValidationError,
--- a/frontend/components.js
+++ b/frontend/components.js
@@ -264,6 +264,11 @@ class UnlockClass extends Component {
       h(
         'div',
         {class: 'auth-section'},
+        h(
+          'div',
+          {class: 'centered-row margin-bottom'},
+          'Enter the 12-word wallet mnemonic or 27-word Daedalus paper wallet mnemonic'
+        ),
         mnemonicValidationError &&
           showMnemonicValidationError &&
           h('p', {class: 'alert error'}, translations[mnemonicValidationError.code]()),
@@ -275,7 +280,7 @@ class UnlockClass extends Component {
             class: 'styled-input-nodiv styled-unlock-input',
             id: 'mnemonic-submitted',
             name: 'mnemonic-submitted',
-            placeholder: 'Enter 12-word mnemonic or paper wallet mnemonic',
+            placeholder: 'Enter wallet mnemonic',
             value: mnemonic,
             onInput: updateMnemonic,
             onBlur: checkForMnemonicValidationError,
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@stablelib/chacha": "^0.7.2",
     "@stablelib/chacha20poly1305": "^0.10.2",
     "bignumber.js": "^6.0.0",
-    "bip39": "^2.5.0",
+    "bip39": "2.5.0",
     "blakejs": "^1.1.0",
     "body-parser": "^1.18.2",
     "bs58": "^4.0.1",
--- a/wallet/cardano-mnemonic-crypto-provider.js
+++ b/wallet/cardano-mnemonic-crypto-provider.js
@@ -12,7 +12,11 @@ const {HdNode, mnemonicToHdNode, hdNodeStringToHdNode} = require('./hd-node')
 const derivePublic = require('./helpers/derivePublic')
 const {packAddress, unpackAddress} = require('./address')
 
-const CardanoMnemonicCryptoProvider = (mnemonicOrHdNodeString, walletState) => {
+const CardanoMnemonicCryptoProvider = (
+  mnemonicOrHdNodeString,
+  walletState,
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
   "dependencies": {
     "@stablelib/chacha20poly1305": "^0.10.2",
     "bignumber.js": "^6.0.0",
-    "bip39": "^2.5.0",
+    "bip39": "2.3.0",
```
ddeath: Why this lib is downgrading?

wallet/mnemonic.js
```diff
@@ -8,13 +12,85 @@ function generateMnemonic() {
...
+  if (isPaperWalletMnemonic(mnemonic)) {
+    mnemonic = decodePaperWalletMnemonic(mnemonic)
+  }
+
+  const ent = Buffer.from(bip39.mnemonicToEntropy(mnemonic), 'hex')
+
+  return cbor.encode(hashBlake2b256(cbor.encode(ent)))
+}
+
+function mnemonicToPaperWalletPasphrase(mnemonic, password) {
```
ddeath: there is missing `s` in `mnemonicToPaperWalletPasphrase`

wallet/mnemonic.js
```diff
@@ -8,13 +12,85 @@ function generateMnemonic() {
...
+  if (paperWalletMnemonicAsList.length !== 27) {
+    throw Error(
+      `Paper Wallet Mnemonic must be 27 words, got ${paperWalletMnemonicAsList.length} instead`
+    )
+  }
+
+  const scrambledPart = paperWalletMnemonicAsList.slice(0, 18).join(' ')
+  const passphrasePart = paperWalletMnemonicAsList.slice(18, 27).join(' ')
+
+  const passphrase = mnemonicToPaperWalletPasphrase(passphrasePart)
```
ddeath: why is there injected `passphrasePart` in the spot where the mnemonic should be? I would rename `passphrasePart` to mnemonic, or if it is not a mnemonic then I would suggest to rename function `mnemonicToPaperWalletPasphrase` so it will be clear what it is for.

wallet/mnemonic.js
```diff
@@ -8,13 +12,85 @@ function generateMnemonic() {
...
+  }
+
+  return bip39.entropyToMnemonic(output)
+}
+
 module.exports = {
   generateMnemonic,
   validateMnemonic,
+  mnemonicToHashSeed,
+  _decodePaperWalletMnemonic: decodePaperWalletMnemonic,
```
ddeath: why it is not exposed? Also there is no UI, but I guess that if there was it would use this function?

ddeath: Ok, I see that support for paper wallet is injected into "normal mnemonic input" - but there is a small catch. The 27 words is a valid mnemonic too if we look at bip39 standard so there may be conflict

wallet/cardano-mnemonic-crypto-provider.js
```diff
@@ -12,7 +12,11 @@ const {HdNode, mnemonicToHdNode, hdNodeStringToHdNode} = require('./hd-node')
 const derivePublic = require('./helpers/derivePublic')
 const {packAddress, unpackAddress} = require('./address')
 
-const CardanoMnemonicCryptoProvider = (mnemonicOrHdNodeString, walletState) => {
+const CardanoMnemonicCryptoProvider = (
+  mnemonicOrHdNodeString,
+  walletState,
+  disableCaching = false
```
ddeath: not sure if we need this. It looks that it is there only because of tests. Would be better to create a new provider in each test case if you do not want to operate on cached keys?

package.json
refi93: because the newer version adds a checksum test to bip39 mnemonics that does not allow the transformation in the first line of unscrambleStrings, since the entropy we get from this scrambled mnemonic does not have to be valid. Daedalus uses this version of bip39, too, probably for the same reason.

wallet/cardano-mnemonic-crypto-provider.js
refi93: yes, I considered also this approach, but I didn't like the need to reinitialize the cryptoproviders on each test. Moreover, it is quite easy to forget about caching when writing tests if you don't explicitly manage it. For example, the test for nonhardened derivation actually was not testing the nonhardened derivation since the result was already cached from the hardened one with the same index, and I sense that there could be more similar cases in the future.

package.json
refi93: Ok, seems that works properly even with the latest version of bip39, so I upgraded it back. I was probably misled by some bug that I had when developing this functionality.

frontend/components.js
```diff
@@ -275,7 +275,7 @@ class UnlockClass extends Component {
             class: 'styled-input-nodiv styled-unlock-input',
             id: 'mnemonic-submitted',
             name: 'mnemonic-submitted',
-            placeholder: 'Enter twelve-word mnemonic',
+            placeholder: 'Enter 12-word mnemonic or paper wallet mnemonic',
```
ddeath: Just wanted to point out that for example is using 12, 18 or 24 words mnemonic.
Also the paper wallet addition makes the placeholder too long and on iphone5 will not be visible and since it is a placeholder you can not even scroll it.

I suggest to put short paragraph above the input with short note same as in the trezor tab.

PR #107, opened 2018-07-03
Add mnemonic generation modal

ddeath:


refi93:
when I shrink the screen - the "Regenerate" and "Confirm" button become weirdly aligned

refi93:
Also the modal is unscrollable if the screen has too little height (try shrinking the screen vertically)

refi93:
I was playing around with it a bit and I came to the conclusion that I would drop the "Regenerate" button - I believe it isn't really needed. Also I put an overflow-y on the modal, so it would be scrollable on very small screens - you can try it by shrinking vertically the browser window. I also unified the size of the "Generate mnemonic" title with the disclaimer title. @ddeath feel free to modify it and merge it if you are OK with it.

PR #108, opened 2018-07-03
json keyfiles import gui

git-jano:
load a wallet form Daedalus compatible json (un)encrypted keyfile


refi93:
The password prompt is too wide imho, I would recommend limiting its width to avoid unnatural stretching on wide screens.

refi93:
Also I would prefer if the password prompt did not close if you insert a wrong password. Uploading the key file again just because you misspeled your password is a bit frustrating.

wallet/keypass-json.js
```diff
@@ -0,0 +1,150 @@
...
+  return [
+    n.toString(),
+    r.toString(),
+    p.toString(),
+    salt.toString('base64'),
+    hash.toString('base64'),
+  ].join('|')
+}
+
+async function encryptOrDecryptWalletSecret(walletSecret, password) {
```
ddeath: This is ambiguous and also in my opinion breaks single responsibility principle. I suggest to make two functions instead of this one.

wallet/keypass-json.js
```diff
@@ -0,0 +1,150 @@
+const scrypt = require('scryptsy')
```
ddeath: I would add some comment with link to the code which was used as a `template`

frontend/components.js
```diff
@@ -4,6 +4,234 @@ import actions from './actions'
...
+              ''
+            ),
+            h(
+              'div',
+              {class: 'margin-1rem'},
+              h(
+                'div',
+                {class: 'intro-input-row margin-top'},
+                h('input', {
+                  type: 'text',
```
ddeath: I suggest to change this to `password`

wallet/keypass-json.js
refi93: The purpose of the name of the function is to make clear that the function is self inverse, but I split it anyway, putting the crypto stuff in the "memoryCombine" function - similarly to the source code in cardano-crypto. I tried making two separate functions but I didn't like it, because at first sight you may not understand if the functions decrypt and encrypt do the same thing, which is, I believe quite an important insight.

wallet/keypass-json.js
refi93: I added a reference to the original github repo above, fixed

wallet/keypass-json.js
refi93: One "hacky" approach I tried is the following:
const [encryptWalletSecret, decryptWalletSecret] = Array(2).fill(async(x,y) => {...})

But I think this goes quite a bit against readability, so I dropped it, but maybe you think otherwise.

frontend/components.js
```diff
@@ -85,25 +316,74 @@ class UnlockClass extends Component {
...
             )
           )
         )
       )
+
+    const LoadByFileSection = () =>
+      h(
+        'div',
+        {class: 'auth-section'},
+        h('div', {class: 'centered-row'}, 'CardanoLite supports key files generated by Daedalus.'),
```
ddeath: I would check with @refi93 what text we want here. Because cardanolite do not operate with same addresses as daedalus.

refi93: I would drop this message. Daedalus currently supports importing/exporting wallet as JSON only in developer mode and the import of a JSON from Daedalus won't work as expected for most users that are not aware of the differences between cardanolite and daedalus.

I would put a similar message only when exporting the wallet and only when the JSON import/export JSON feature makes it to the official version of Daedalus.

ddeath:
Also suggesting to add some padding to the content on mobile as the content of tabs (on "login page") is right on the edge of the screen

ddeath:
@refi93 your suggestions were also implemented

frontend/components.js
```diff
@@ -4,6 +4,239 @@ import actions from './actions'
...
+    this.setState({encrypted: undefined, error: undefined})
+  }
+
+  updatePassword(e) {
+    this.setState({password: e.target.value})
+  }
+
+  unlockKeyfile() {
+    this.props.loadingAction('Unlocking key file')
+    setTimeout(
```
refi93: why setTimeout?

frontend/components.js
```diff
@@ -4,6 +4,239 @@ import actions from './actions'
...
+      error: undefined,
+    })
+
+    const reader = new FileReader()
+    await reader.readAsText(file)
+
+    reader.onload = ((theFile) => {
+      return async (e) => {
+        try {
+          const json = await JSON.parse(e.target.result)
```
refi93: "json" name for variable is imho too generic. Maybe walletExport or something along those lines would be more suitable.

frontend/components.js
```diff
@@ -4,6 +4,239 @@ import actions from './actions'
...
+          class: 'overlay-close-button',
+          onClick: closeGenerateMnemonicDialog,
+        },
+        ''
+      ),
+      h('h4', undefined, 'Generate a Mnemonic Phrase'),
+      h(
+        'h7',
+        undefined,
+        'Write these words down. Do not copy them to your clipboard, or save them anywhere online.'
```
refi93: that comma before "or" is imho redundant.

frontend/components.js
```diff
@@ -85,25 +320,68 @@ class UnlockClass extends Component {
...
+      )
+
+    const LoadByHardwareWalletSection = () =>
+      h(
+        'div',
+        {class: 'auth-section'},
+        h(
+          'div',
+          undefined,
+          'Hardware wallets porvide the best security level for your Cryptocurrency.'
```
refi93: typo: "porvide" -> "provide" I would reword it like this: "Hardware wallets provide the best security level for storing your cryptocurrencies."

git-jano:
screenshot from 2018-07-10 14-10-42
on small mobiles iPhone5 /SE  / Nokia Lumia 520 it shows 2 tabs instead of 3 !

PR #114, opened 2018-07-09
Send all ADA support

jmuravsky:
functionality 99.9% done, needs some CSS work.


jmuravsky:
also refactored calculating fee to not fluctuate for fixed inputs/outputs and take into account possibility of single output transactions

frontend/actions.js
```diff
@@ -215,6 +253,7 @@ export default ({setState, getState}) => {
       ownAddressesWithMeta,
       transactionHistory,
     })
+    checkAllFundsSendable(getState())
```
refi93: I would rather set the "allFundsSendable" key in the setState call above

frontend/actions.js
```diff
@@ -128,6 +130,42 @@ export default ({setState, getState}) => {
     })
   }
 
+  const checkAllFundsSendable = (state) => {
```
refi93: this function should imho return a bool and it should be the responsibility of the calling function to set the state appropriately

frontend/components.js
```diff
@@ -628,6 +630,22 @@ class SendAdaClass extends Component {
             {class: 'validationMsg send-amount-validation-error'},
             translations[sendAmountValidationError.code](sendAmountValidationError.params)
           ),
+        allFundsSendable
+          ? h(
+            'a',
+            {
+              class: 'intro-link fade-in-up', //TODO: fix CSS
```
refi93: todo comment left there

frontend/components.js
```diff
@@ -628,6 +630,22 @@ class SendAdaClass extends Component {
             {class: 'validationMsg send-amount-validation-error'},
             translations[sendAmountValidationError.code](sendAmountValidationError.params)
           ),
+        allFundsSendable
+          ? h(
```
refi93: too much repetition, this could be merged into one element and use the ternary operator only on the second parameter, something like this:
`
h('a', allFundsSendable ? {class: 'intro-link ...', onClick: '...'} : {class: 'disabled-link'}, 'send all funds')`

frontend/actions.js
```diff
@@ -128,6 +130,42 @@ export default ({setState, getState}) => {
...
+    } else {
+      setState({
+        allFundsSendable: true,
+      })
+    }
+  }
+
+  const sendAllFunds = async (state) => {
+    setState({calculatingFee: true})
+    checkAllFundsSendable(state)
```
refi93: this check is probably redundant

frontend/actions.js
```diff
@@ -128,6 +130,42 @@ export default ({setState, getState}) => {
...
+    if (state.allFundsSendable) {
+      let transactionFee = 0
+      if (state.transactionFee) {
+        transactionFee = await wallet.getTxFee('', state.balance - state.transactionFee)
+      } else {
+        const minTxFee = Cardano.txFeeFunction(349) //smallest possible size for one input, 2 outputs transaction
+        transactionFee = await wallet.getTxFee('', state.balance - minTxFee)
```
refi93: I would drop the address from getTxFee if it is not used

wallet/cardano-wallet.js
```diff
@@ -189,7 +196,25 @@ const CardanoWallet = async (options) => {
...
+
+    /*
+    * if they are equal it means there is no change necessary
+    * if they are more the transaction is invalid even with higher fee so we let caller handle it
+    */
+    if (coins + fee >= txInputsCoinsSum) {
+      return fee
+    } else {
+      fee = txFeeFunction(txSizeInBytes + txOutputsSize + deviation) //we try to compute fee for 2 output tx
+      if (coins + fee > txInputsCoinsSum) {
```
refi93: this case should also be handled when preparing the transaction outputs. Currently if coins + fee < tx inputs sum, the final transaction contains two outputs. Maybe this function could output an object instead of a single number to provide additional information to the caller, something like this: 
{fee: 155547, outputsCount: 2, valid: false}

wallet/cardano-wallet.js
```diff
@@ -189,7 +196,25 @@ const CardanoWallet = async (options) => {
     */
     const deviation = 4
 
-    return txFeeFunction(txSizeInBytes + deviation)
+    let fee = txFeeFunction(txSizeInBytes + deviation)
+
+    /*
+    * if they are equal it means there is no change necessary
```
refi93: this comment is unclear, what is "they"?

jmuravsky: i thought about rewording this, but hoped this is clear enough, i can add a `(coin+fee)` there for example or word it completely differently

wallet/cardano-wallet.js
jmuravsky: i think that bloats the scope of this issue a bit, id say make it a separate issue depending on this.

refi93: ok, I misunderstood the code

frontend/actions.js
```diff
@@ -147,6 +149,28 @@ module.exports = ({setState, getState}) => {
     })
   }
 
+  const sendAllFunds = async (state) => {
+    setState({calculatingFee: true})
+
+    const balance = await wallet.getBalance()
+    const allFundsSendable = balance > (await wallet.getAllFundsTxFee())
```
ddeath: I would rewrite it:
```
const txFee = await getAllFundsTxFee()
const allFundsSendable = balance > txFee
```

then you can reuse `txFee` inside if block

wallet/cardano-wallet.js
```diff
@@ -143,24 +155,39 @@ const CardanoWallet = async (options) => {
       return acc + elem.coins
     }, 0)
 
-    const out1coins = coins
-    const out2coinsUpperBound = Math.max(0, txInputsCoinsSum - coins)
+    const oneOutputFee = txFeeFunction(estimateTxSize(txInputs, coins, 1)) //first we try one output transaction
```
ddeath: these lines are too long. Check the linter errors / warnings

frontend/translations.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   SendAmountIsNotPositive: () => 'Invalid format: Amount has to be a positive number',
   SendAmountInsufficientFunds: ({balance}) =>
     `Insufficient funds for the transaction. Your balance is ${printAda(balance)} ADA.`,
+  SendAmountCantSendAllFunds: () => 'Insufficient balance to send all funds',
```
ddeath: Suggesting to rewrite to something like `Sending funds is not possible because there is no balance for paying tx fee`,

PR #116, opened 2018-07-10
Add code splitting for faster load time and fix imports

ddeath:
* unify imports vs require
* move some functions we can better split the code
* add code spliting


frontend/actions.js
```diff
@@ -1,11 +1,11 @@
-import Cardano from '../wallet/cardano-wallet'
-import {CARDANOLITE_CONFIG} from './config'
-import {
+const generateMnemonicWords = require('../wallet/mnemonic').generateMnemonic
```
refi93: "mnemonicWords" is a bit redundant since the mnemonic itself is defined as a set of words. I would name it simply "generateMnemonic"

PR #121, opened 2018-07-11
Add option to export wallet to JSON file

ddeath:
Added button which enable export wallet to json file (when you are not using trezor)


refi93:
I would also add the option to set the wallet export file name and this would also set the wallet name in the json which is now hardcoded.

PR #122, opened 2018-07-11
unified ed25519 cardano implementation with cardano-crypto.js

refi93:
I took out a significant part of crypto from cardanolite and put it in an independent package: https://github.com/vacuumlabs/cardano-crypto.js


refi93:
I would prefer to do the transition iteratively. For now I picked the biggest dependencies, which were the various ed25519 implementations we used.

wallet/constants.js
```diff
@@ -9,4 +9,5 @@ module.exports = {
   * the "5820" part is the CBOR prefix for a hex string
   */
   TX_SIGN_MESSAGE_PREFIX: '011a2d964a095820',
+  CARDANO_KEY_DERIVATION_MODE: 1,
```
ddeath: I would add some comment with link to some doc (if there is some) or some explanation of what possible values are and what is their meaning

PR #123, opened 2018-07-12
Refactor structure of components.js

ddeath:
The `component.js` becomes folder `components` and it is split into page components and their child components. Also common folder was created.


refi93:
Please move the demo wallet warning dialog to the login section. To fit into the concept, it can be triggered right after pressing "Go" on login page and if you click "I understand" it proceeds to the wallet, otherwise you stay on the login page

PR #127, opened 2018-07-12
Refactor estimateTxSize to take into account outgoing address size

jmuravsky:


wallet/cardano-wallet.js
```diff
@@ -82,7 +83,7 @@ const CardanoWallet = async (options) => {
       return acc + elem.coins
     }, 0)
 
-    const fee = computeTxFee(txInputs, coins)
+    const fee = computeTxFee(txInputs, coins, base58.decode(address).length)
```
refi93: imho computeTxFee should receive the address directly, not only its size

wallet/cardano-wallet.js
```diff
@@ -146,7 +149,7 @@ const CardanoWallet = async (options) => {
     return txInputs
   }
 
-  function computeTxFee(txInputs, coins) {
+  function computeTxFee(txInputs, coins, outAddressSize = 76) {
```
refi93: I would not put there a default parameter

wallet/cardano-wallet.js
```diff
@@ -178,19 +181,27 @@ const CardanoWallet = async (options) => {
     }
   }
 
-  function estimateTxSize(txInputs, coins, outputcount) {
+  function estimateTxSize(txInputs, coins, hasChange, outAddressSize) {
```
refi93: I would flip the order of hasChange and outAddress, since I see this order as more intuitive.

wallet/cardano-wallet.js
```diff
@@ -178,19 +181,27 @@ const CardanoWallet = async (options) => {
...
+    const ownAddressSize = 76
+
     /*
     * we assume that at most two outputs (destination and change address) will be present
     * encoded in an indefinite length array
     */
     const maxCborCoinsLen = 9 //length of CBOR encoded 64 bit integer, currently max supported
-    const txOutputSize = (77 + maxCborCoinsLen + 1) * outputcount
+    let txOutputsSize
+    if (hasChange) {
```
refi93: I would reformulate this with ternary operator into a single expression

refi93:
I would also add a testcase for tx fee estimation for a shorter output address, for example this one: Ae2tdPwUPEZ18ZjTLnLVr9CEvUEUX4eW1LBHbxxxJgxdAYHrDeSCSbCxrvx

frontend/actions.js
```diff
@@ -152,8 +152,9 @@ module.exports = ({setState, getState}) => {
   const sendAllFunds = async (state) => {
     setState({calculatingFee: true})
 
+    const address = state.sendAddress.fieldValue
     const balance = await wallet.getBalance()
-    const txFee = await wallet.getAllFundsTxFee()
+    const txFee = await wallet.getAllFundsTxFee(address)
```
refi93: I would rather put the "state.sendAddress.fieldValue" directly as an input of getAllFundsTxFee, since you would need to look at the definition of "address" anyway if you want to understand the call below and it is used only once

PR #135, opened 2018-07-16
fix paper wallet passphrase validation, fix paper wallet mnemonic test

### LINKED ISSUE #134: When i enter invalid mnemonic the wallet freezes
If I enter `civil void tool perfect avocado sweet immense fluid arrow aerobic boil flash civil void tool perfect avocado sweet immense fluid arrow aerobic boil flash boil flash house`, the wallet allows to "login" but it will get stuck on loading screen with error in console `Uncaught (in promise) Error: Invalid mnemonic checks`

git-jano:
closes #134

### DIFF
--- a/frontend/validators.js
+++ b/frontend/validators.js
@@ -1,4 +1,4 @@
-const validateMnemonic = require('../wallet/mnemonic').validateMnemonic
+const {validateMnemonic, validatePaperWalletMnemonic} = require('../wallet/mnemonic')
 const isValidAddress = require('../wallet/address').isValidAddress
 
 const parseCoins = (str) => Math.trunc(parseFloat(str) * 1000000)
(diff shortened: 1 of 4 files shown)


frontend/validators.js
```diff
@@ -44,7 +46,10 @@ const feeValidator = (sendAmount, transactionFee, balance) => {
 const mnemonicValidator = (mnemonic) => {
   let validationError = null
 
-  if (!validateMnemonic(mnemonic)) {
+  if (
+    !validateMnemonic(mnemonic) ||
+    (isPaperWalletMnemonic(mnemonic) && !decodePaperWalletMnemonic(mnemonic))
```
ddeath: maybe would be cleaner to rename `isPaperWalletMnemonic` to something like `isMnemonicInPaperWalletFormat` and create function `validatePaperWalletMnemonic` which would check `isMnemonicInPaperWalletFormat` and would try to `decodePaperWalletMnemonic` so we end up with `!validateMnemonic(mnemonic) && !validatePaperWalletMnemonic(mnemonic)`

frontend/validators.js
```diff
@@ -1,4 +1,6 @@
 const validateMnemonic = require('../wallet/mnemonic').validateMnemonic
```
refi93: I would merge this three lines into one statement:

const {a, b, c } = require('x')

PR #136, opened 2018-07-17
remove various crypto dependencies by importing cardano-crypto.js

refi93:
I removed various dependencies by importing and extending cardano-crypto.js and replacing bip39 with bip39-light. Overall bundle size should be reduced to ~1MB. Everything should work the same as previously.


wallet/helpers/isValidAddress.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(address) {
```
ddeath: It looks like there are missing imports for cbor and base58

wallet/helpers/parseMnemonicOrHdNodeString.js
```diff
@@ -0,0 +1,19 @@
+const CardanoCrypto = require('cardano-crypto.js')
+
+const {isPaperWalletMnemonic, decodePaperWalletMnemonic} = require('../mnemonic')
+
+module.exports = async (mnemonicOrHdNodeString) => {
```
ddeath: I would suggest `const parseMnemonicOrHdNodeString =` and then at the bottom of file `module.exports = parseMnemonicOrHdNodeString`

This way you will be able search and find the function in the code

webpack.build.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   },
   externals: {
     'trezor-connect': 'TrezorConnect',
+    'crypto': 'crypto',
```
ddeath: this looks like forgotten line

webpack.test.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
   },
   externals: {
     'trezor-connect': 'TrezorConnect',
+    'crypto': 'crypto',
```
ddeath: same here

PR #138, opened 2018-07-18
reorganized folder structure

refi93:
check if new folder structure makes sense and if .gitignore, linting, running server, is set right


ddeath:
I would move `svg.js` into `components/common` folder

git-jano:
/navbar/exportWalletDialog.js might go up, it is triggered by a button in navbar, but the action is not related to navbar

git-jano:
why is landing page at top level and not in client? i suggest put it in frontend

git-jano:
what is /public/js for? semms empty even after build

ddeath:
looks like `frontend/middleware.js` is unused so I suggest to delete it

ddeath:
It seem that files `printAda.js`, `validators.js`, `debugLog.js` are just helpers so I suggest to move it to `wallet/helper` also maybe rename it to `utils`

ddeath:
I would move `initialState` and `createStore` from `walletApp.js` to `store.js` - if you see in folder `actions.js` then you will probably also looking for `store.js`

PR #153, opened 2018-07-26
fix: wrong mnemonic prevent new one or demo 

### LINKED ISSUE #144: Make new mnemonic after generating it directly copyable from the modal
It can make the use of the wallet for newcomers easier

### LINKED ISSUE #145: When I click to generate a new mnemonic and the input field contains invalid mnemonic, an error is displayed
The expected behavior is that the mnemonic in the input field would be overwritten even if not valid

git-jano:
closes #144 fix: enable selecting text in generateNewMnemonic dialog
closes #145 fix: wrong mnemonic prevent new one or demo -> err moves layout

### DIFF
--- a/app/frontend/components/pages/login/mnemonicAuth.js
+++ b/app/frontend/components/pages/login/mnemonicAuth.js
@@ -13,8 +13,13 @@ const LoadByMenmonicSection = ({
   showGenerateMnemonicDialog,
   loadDemoWallet,
   showMnemonicValidationError,
-}) =>
-  h(
+}) => {
+  const isLeftClick = (e, action) => {
+    if (e.button === 0) {
+      action()
+    }
+  }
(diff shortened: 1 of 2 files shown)


app/frontend/components/pages/login/mnemonicAuth.js
```diff
@@ -59,7 +64,7 @@ const LoadByMenmonicSection = ({
       'a',
       {
         class: 'intro-link fade-in-up',
-        onClick: openGenerateMnemonicDialog,
+        onMouseDown: (e) => isLeftClick(e, openGenerateMnemonicDialog),
```
git-jano: this is to prevent mnemonic field onBlur happen before onClick (validator will show err, which moves layout, so click might end outside the button)

PR #154, opened 2018-07-27
show spinner until transaction appears in explorer

refi93:
The spinner now should be displayed until the transaction actually appears in the blockchain explorer, so the user won't be confused.

If the transaction fails to appear within ~100 seconds (20 retries * 5s sleep), an error should be displayed where the user is notified about the transaction id, so he can check it later.

I also refactored and fixed error handling when submitting transactions. I added NamedError helper. Most of the errors were not displayed/handled propery, since they did not make it to the "sendResponse" state variable.

I also refactored translations so they have a nicer API and fail gracefully if translation code is not present

### DIFF
--- a/app/frontend/actions.js
+++ b/app/frontend/actions.js
@@ -9,6 +9,7 @@ const {
 } = require('./helpers/validators')
 const printAda = require('./helpers/printAda')
 const debugLog = require('./helpers/debugLog')
+const sleep = require('./helpers/sleep')
 
 let wallet = null
 
@@ -228,7 +229,7 @@ module.exports = ({setState, getState}) => {
     const ownAddressesWithMeta = await wallet.getOwnAddressesWithMeta()
     const transactionHistory = await wallet.getHistory()
 
-    // timeout setting loading state, so that loading shows even if everything was cashed
+    // timeout setting loading state, so that loading shows even if everything was cached
     setTimeout(() => setState({loading: false}), 500)
     setState({
       balance,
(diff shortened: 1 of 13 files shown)


app/frontend/actions.js
```diff
@@ -9,6 +9,8 @@ const {
 } = require('./helpers/validators')
 const printAda = require('./helpers/printAda')
 const debugLog = require('./helpers/debugLog')
+const sleep = require('./helpers/sleep')
+const NamedError = require('./helpers/NamedError')
```
ddeath: this one is unused

app/frontend/actions.js
```diff
@@ -357,29 +359,53 @@ module.exports = ({setState, getState}) => {
...
       if (sendResponse.success) {
-        setTimeout(() => setState({sendResponse: ''}), 4000)
+        loadingAction(state, 'Transaction submitted - syncing wallet...')
+
+        const maxRetries = 20
+        for (let pollingCounter = 0; pollingCounter < maxRetries; pollingCounter++) {
+          if ((await wallet.fetchTxInfo(sendResponse.txHash)) !== undefined) {
+            /*
+            * theoretically we should clear the request cache of the wallet
+            * but submitting the transaction and syncing of  the explorer
```
ddeath: nitpick: there are two spaces after  `of`

app/frontend/actions.js
```diff
@@ -357,29 +359,53 @@ module.exports = ({setState, getState}) => {
...
+    loadingAction(state, 'Submitting transaction...')
     try {
       const address = state.sendAddress.fieldValue
       const amount = state.sendAmount.coins
-      const sendResponse = await wallet.sendAda(address, amount)
-      const updatedBalance = await wallet.getBalance()
+      let sendResponse = await wallet.sendAda(address, amount)
       if (sendResponse.success) {
-        setTimeout(() => setState({sendResponse: ''}), 4000)
+        loadingAction(state, 'Transaction submitted - syncing wallet...')
```
ddeath: I would refactor this to the function something like `waitForBlockchain()` it will make code simpler and more easy to understand

app/frontend/actions.js
```diff
@@ -357,29 +359,53 @@ module.exports = ({setState, getState}) => {
...
+            sendResponse = Object.assign(sendResponse, {
+              success: false,
+              error: 'TransactionNotFoundInBlockchainAfterSubmission',
+            })
+            break
+          }
+
+          await sleep(5000)
+        }
+
```
ddeath: I would refactor this set state to the `resetSendForm(sendResponse)`

app/frontend/actions.js
```diff
@@ -357,29 +359,53 @@ module.exports = ({setState, getState}) => {
...
+            /*
+            * theoretically we should clear the request cache of the wallet
+            * but submitting the transaction and syncing of  the explorer
+            * should take more time than the TTL of the cache
+            */
+            await reloadWalletInfo(state)
+            break
+          }
+
+          if (pollingCounter === maxRetries - 1) {
```
ddeath: I suggest to change it:
```
for (let pollingCounter = 0; pollingCounter < maxRetries; pollingCounter++) {
    const txInfo = await wallet.fetchTxInfo(sendResponse.txHash)
    if (txInfo !== undefined) {
        await reloadWalletInfo(state)
        resetSendForm(state, sendResponse)
        return
    }

    if (pollingCounter < maxRetries - 1) {
        await sleep(5000)
    }
}
```

git-jano:
When offline during waiting for tx to apperar in blockchain explorer, I get `Transaction failed: TypeError` on UI, and ```
debugLog.js:6 TypeError: Failed to fetch
debugLog.js:6 Translation for TypeError not found!
```

PR #159, opened 2018-07-27
Donation urls in footer

### LINKED ISSUE #150: Add donations address
it/they can be displayed in github as well as in the app

git-jano:
closes #150


README.md
```diff
@@ -66,4 +66,10 @@ yarn test
 
 Open `app/tests/index.html` in browser
 
-Check us out at https://www.cardanolite.com
\ No newline at end of file
+Check us out at https://www.cardanolite.com
+
+---
+Donations are really appreciated!
+[BTC](https://www.blockchain.com/btc/address/1DECAF2uSpFTP4L1fAHR8GCLrPqdwdLse9) /
```
refi93: I would put the address in the link title - it would be then look like this:
BTC: [0x...](https://explorer.com/0x....)

This applies to ETH and ADA as well

app/frontend/components/common/footer.js
```diff
@@ -23,19 +33,78 @@ const Footer = () =>
...
       ),
       '/',
       h(
         'small',
         {class: 'contact-link'},
-        h('a', {href: 'https://twitter.com/hashtag/cardanolite'}, '#cardanolite')
+        h('a', {href: 'https://twitter.com/hashtag/cardanolite', target: '_blank'}, '#cardanolite')
+      )
+    ),
+    (donateBtcUrl || donateEthUrl || donateAdaAddress) &&
```
refi93: originally I thought that this check would make sense, but now I think that it just bring unnecesary corner cases with displaying the slashes right to the code. We already know what currencies we want. Therefore I would remove the check for the presence of BTC, ADA, ETH donation address for the sake of simplicity.

app/frontend/store.js
```diff
@@ -30,6 +30,15 @@ const initialState = {
   validationMsg: undefined,
   authMethod: 'mnemonic',
   enableTrezor: CARDANOLITE_CONFIG.CARDANOLITE_ENABLE_TREZOR,
+  donateBtcUrl: CARDANOLITE_CONFIG.CARDANOLITE_DONATE_BTC_ADDRESS
+    ? `https://www.blockchain.com/btc/address/${CARDANOLITE_CONFIG.CARDANOLITE_DONATE_BTC_ADDRESS}`
```
refi93: I would make this URL a named constant and put it in constants.js. Same for ETH.

app/frontend/store.js
```diff
@@ -30,6 +31,15 @@ const initialState = {
   validationMsg: undefined,
   authMethod: 'mnemonic',
   enableTrezor: CARDANOLITE_CONFIG.CARDANOLITE_ENABLE_TREZOR,
+  donateBtcUrl: CARDANOLITE_CONFIG.CARDANOLITE_DONATE_BTC_ADDRESS
+    ? BTC_BLOCKCHAIN_EXPLORER_URL + CARDANOLITE_CONFIG.CARDANOLITE_DONATE_BTC_ADDRESS
```
refi93: I would leave out the ternary operator. I think it is safe to assume that the addresses would always be available

app/frontend/components/common/footer.js
```diff
@@ -23,19 +33,75 @@ const Footer = () =>
...
+      )
+    ),
+    h(
+      'p',
+      undefined,
+      h('div', undefined, 'Donations are really appreciated! '),
+      h(
+        'div',
+        undefined,
+        donateBtcUrl &&
```
refi93: I would remove this &&, same for remaining coins. I would assume that the urls and addresses are available

app/frontend/components/common/footer.js
```diff
@@ -23,19 +36,74 @@ const Footer = () =>
           'View on Github'
         )
       ),
-      '/',
+      ' /',
```
refi93: this space is probably redundant since other links do not have this space before the slash

refi93:
approved, waiting for actual addresses to merge

PR #162, opened 2018-07-31
added mocking of transaction submission for easier development

refi93:
usage explained in README.

- check that the explanation is clear.
- check that tx submission mocking actually works as described


server/transactionSubmitter.js
```diff
@@ -24,7 +25,15 @@ module.exports = function(app, env) {
       return res.status(500).send('bad request format')
     }
 
-    // return res.end(JSON.stringify({success: true, txHash}))
+    if (process.env.CARDANOLITE_MOCK_TX_SUBMISSION === 'true') {
+      await sleep(5000)
+      return res.end(
+        JSON.stringify({
+          success: process.env.CARDANOLITE_MOCK_TX_SUBMISSION_RESPONSE === 'true',
```
git-jano: shouldn't it be CARDANOLITE_MOCK_TX_SUBMISSION_SUCCESS ?

server/blockchainExplorerMock.js
```diff
@@ -16,7 +21,7 @@ module.exports = function(app, env) {
     const txId = req.params.txId
 
     const response = await request(
-      `${process.env.CARDANOLITE_BLOCKCHAIN_EXPLORER_PROXY_TARGET}/api/txs/raw/${txId}`
+      `${process.env.CARDANOLITE_MOCK_BLOCKCHAIN_EXPLORER_TARGET}/api/txs/raw/${txId}`
```
git-jano: misleading naming if the target is not mocked

server/helpers/loadFrontendConfig.js
```diff
@@ -1,8 +1,17 @@
 // config intended to be passed on to frontend
 // keys in process .env must be referencesd explicitly
 // so dotenv-webpack is able to include them in the test bundle for Mocha tests
+
+function getBlockchainExplorerUrlForFrontend() {
+  if (process.env.CARDANOLITE_ENABLE_MOCK_SERVER === 'true') {
+    return `http://localhost:${process.env.PORT}`
+  } else {
```
git-jano: what about ternary operator? is it short enough to use it here?

app/frontend/helpers/debugLog.js
```diff
@@ -1,7 +1,7 @@
 const CARDANOLITE_CONFIG = require('../config').CARDANOLITE_CONFIG
 
 function debugLog(message) {
-  if (CARDANOLITE_CONFIG.CARDANOLITE_ENABLE_DEBUGGING) {
+  if (CARDANOLITE_CONFIG && CARDANOLITE_CONFIG.CARDANOLITE_ENABLE_DEBUGGING) {
```
git-jano: is this useful? who would bebug without config?

app/frontend/wallet/cardano-wallet.js
```diff
@@ -62,13 +62,13 @@ const CardanoWallet = async (options) => {
...
       throw NamedError('TransactionRejectedByTrezor')
     })
+
     const response = await blockchainExplorer
       .submitTxRaw(signedTx.txHash, signedTx.txBody)
       .catch((e) => {
         debugLog(e)
         throw NamedError('TransactionRejectedByNetwork')
       })
-    updateUtxosFromTxAux(txAux)
```
git-jano: how is this replaced?

refi93: thanks for pointing out, I forgot to reverse this change after debugging the tests

app/frontend/helpers/debugLog.js
refi93: the tests fail without this. There is already an issue to refactor debug logging: https://github.com/vacuumlabs/cardanolite/issues/161

refi93: but it fails only if something is logged which isn't if everything goes as expected. Ok, dropping this change for now.

PR #170, opened 2018-08-01
Issue 129 tidy up css

### LINKED ISSUE #126: Add unspent outputs integrity check also to mnemonic crypto provider
Trezor currently checks transactions leading to the unspent outputs referenced in the submitted transaction to avoid losing funds to unexpectedly high fees if the sender manages to "lie" about the actual amounts of coins in the inputs. Mnemonic crypto provider should do the same check.

git-jano:
closes #126

### DIFF
--- a/app/public/css/styles.css
+++ b/app/public/css/styles.css
@@ -900,13 +900,12 @@ div.box.box-auto {
 /*LANDING*/
 .intro-wrapper {
   margin: auto;
-  margin-top: 1.5rem;
   padding: 0 1em;
 }
 .intro-header {
   font-size: 2.3rem;
   font-weight: 600;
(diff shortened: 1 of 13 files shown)


app/public/css/styles.css
```diff
@@ -82,8 +115,10 @@ p {
 path {
   fill: currentColor;
 }
+
+
 /* BUTTON */
-button, .button-like {
```
refi93: button-like class appears still in styles.css despite I can't find it in the code. Are you sure this change won't mess up the styles. Moreover, btn as an element probably does not exist, you probably wanted to put the class ".btn", but this is imho too generic. Was there any good reason to replace "button" selector with "btn"?

PR #173, opened 2018-08-07
Added fitler to drop utxos too small to add value to transaction

jmuravsky:
Since adding an input to transaction increases its cost, some utxos are too small to add value to transaction, because the fee increase to add them is more than their value. Fixed this issue by dropping such utxos.


app/frontend/wallet/cardano-wallet.js
```diff
@@ -162,29 +162,37 @@ const CardanoWallet = async (options) => {
     return await blockchainExplorer.fetchTxInfo(txHash)
   }
 
+  function dropUnprofitableUtxos(utxos) {
+    const profitableUtxos = []
+    const witnessSize = 139
```
refi93: this should be moved probably to constants.js and I would call it "txWitnessByteSize"

app/frontend/wallet/cardano-wallet.js
```diff
@@ -162,29 +162,37 @@ const CardanoWallet = async (options) => {
...
+    let addedCost = 0
```
refi93: I would decalre the variables inputSize and addedCost inside the for loop with const instead of declaring them with "let" outside of the for loop

app/frontend/wallet/cardano-wallet.js
```diff
@@ -162,29 +162,37 @@ const CardanoWallet = async (options) => {
...
+    let inputSize = 0
+    for (const utxo of utxos) {
```
refi93: this can be rewritten like this:
```
return utxos.filter((utxo) => {
    const inputSize = cbor.encode(TxInputFromUtxo(utxo)).length
    const addedCost = txFeeFunction(inputSize + witnessSize) - txFeeFunction(0)

    return utxo.coins > addedCost
})
```

app/frontend/wallet/cardano-wallet.js
```diff
@@ -162,29 +162,37 @@ const CardanoWallet = async (options) => {
...
+      }
+    }
+    return profitableUtxos
+  }
+
   async function prepareTxInputs(coins, address) {
     // we do it pseudorandomly to guarantee fee computation stability
     const randomGenerator = PseudoRandom(state.randomSeed)
     const utxos = shuffleArray(await getUnspentTxOutputs(), randomGenerator)
+    const profitableUtxos = dropUnprofitableUtxos(utxos)
```
refi93: I suggest making a function for evaluating if a single utxo is profitable and ussing array filter there. Something like this:
```
function isUtxoProfitable(utxo) {
    const inputSize = cbor.encode(TxInputFromUtxo(utxo)).length
    const addedCost = txFeeFunction(inputSize + witnessSize) - txFeeFunction(0)

    return utxo.coins > addedCost
}
...
const profitableUtxos = utxos.filter(isUtxoProfitable)
```

PR #174, opened 2018-08-07
added transactions counter to stats

refi93:
Added counter of transaction submissions whether they are successful or not. The results should appear in the stats page as soon as you attempt to submit a transaction, if you have set the REDIS_URL properly in .env

I initially tried to filter out unsuccessful tx submission attempts, but I did not find a nice way to intercept the content of the response, maybe you can suggest something.

### DIFF
--- a/server/middlewares/stats.js
+++ b/server/middlewares/stats.js
@@ -1,36 +1,47 @@
 const redis = require('redis')
 const device = require('device')
 const client = redis.createClient(process.env.REDIS_URL)
+const mung = require('express-mung')
 
 const knownIps = new Set()
 
-const monthNames = [
-  'January',
-  'February',
-  'March',
-  'April',
-  'May',
-  'June',
-  'July',
-  'August',
-  'September',
-  'October',
-  'November',
-  'December',
-]
+function getSlicedDate() {
+  return new Date()
+    .toISOString()
+    .split('T')[0]
+    .split('-')
+}
 
-module.exports = (req, res, next) => {
+const trackVisits = (req, res, next) => {
   // 'x-forwarded-for' due to internal heroku routing
-  const ip = req.headers['x-forwarded-for']
+  const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress
 
   const mydevice = device(req.headers['user-agent'])
 
   if (!knownIps.has(ip) && !mydevice.is('bot')) {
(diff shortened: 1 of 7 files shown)


server/statsPage.js
```diff
@@ -58,7 +70,7 @@ module.exports = function(app, env) {
           </head>
       
           <body>
-            Estimated number of unique IPs per day, aggregate for day/month/year
+            Estimated number of unique IPs per day, aggregate for day/month/year and tx submission stats
```
ddeath: I would better explained what the numbers are.

`Estimated number of unique IPs visits and tx submission stats per day and also total number`

ddeath: Ok this would not work as I see that there are also months...

ddeath: Maybe:
`Estimated number of unique IPs visits and tx submission stats per day, per month and for whole time..`

server/middlewares/stats.js
```diff
@@ -24,13 +24,20 @@ module.exports = (req, res, next) => {
...
     knownIps.add(ip)
     client.incr('total')
     client.incr(`${monthNames[date.getMonth()]}-${date.getFullYear()}`)
     client.incr(`${date.getDate()}.${date.getMonth() + 1}.${date.getFullYear()}`)
   }
+
+  if (req.originalUrl === '/api/txs/submit' && req.method === 'POST') {
+    client.incr('total tx submissions')
+    client.incr(`${monthNames[date.getMonth()]}-${date.getFullYear()} (txs submitted)`)
+    client.incr(`${date.getDate()}.${date.getMonth() + 1}.${date.getFullYear()} (txs submitted)`)
```
ddeath: I think it would be better to use some kind of code words so it will be more easy to parse later

server/middlewares/stats.js
```diff
@@ -1,36 +1,46 @@
...
+    client.incr(`visits:daily:${slicedDate[0]}-${slicedDate[1]}-${slicedDate[2]}`)
   }
+
   next()
 }
+
+const trackTxSubmissionCount = mung.json((body, req, res) => {
+  if (req.originalUrl === '/api/txs/submit' && req.method === 'POST') {
+    const slicedDate = new Date()
+      .toISOString()
```
ddeath: I would use `get` methods of date instead of parsing the date string

PR #176, opened 2018-08-10
Improve stats formatting

ddeath:
Reformatted into lists

### DIFF
--- a/server/statsPage.js
+++ b/server/statsPage.js
@@ -2,13 +2,25 @@ const redis = require('redis')
 const redisScan = require('redisscan')
 const client = redis.createClient(process.env.REDIS_URL)
 
-const getData = async () => {
-  const result = []
+const getStats = async () => {
+  const stats = {
+    visits: {total: [], monthly: [], daily: []},
+    txSubmissions: {
+      'successful:total': [],
+      'successful:monthly': [],
+      'successful:daily': [],
+      'unsuccessful:total': [],
+      'unsuccessful:monthly': [],
+      'unsuccessful:daily': [],
+    },
+  }
+
+  const response = []
   await new Promise((resolve, reject) => {
     redisScan({
       redis: client,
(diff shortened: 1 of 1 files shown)


server/statsPage.js
```diff
@@ -76,13 +119,24 @@ module.exports = function(app, env) {
             <meta charset="utf-8"/>
             <title>CardanoLite Wallet Stats</title>
             <link rel="icon" type="image/ico" href="assets/favicon.ico">
+            <style>
```
git-jano: this is useless, as we do not use table

server/statsPage.js
```diff
@@ -58,16 +58,59 @@ const getData = async () => {
...
+    .map(
+      (period) => `
+            <ul>
+              <li>${period}</li>
+              <ul>
+                ${subjects[subject][period]
+    .map(
+      (item) => `
+                  <li>
+                    ${item[0]}: ${item[1]}
```
git-jano: hightlight the vaules,  e.g.  in row `2018-08-08: 0` it is hard to spot, what is the actual value, use bold and/or wider space between  `item[0]` and `item[1]`

server/statsPage.js
```diff
@@ -58,16 +58,59 @@ const getData = async () => {
...
+      const stats = Object.keys(subjects)
         .map(
-          (v) => `
-        <tr>
-          <td>${v[0]}</td>
-          <td>${v[1]}</td>
-        </tr>
-      `
+          (subject) => `
+          <b>${subject}:</b>
```
git-jano: as the data looks like almost column, it would be nice to show visits and (un) successful submisions side by side for visual comparison  as columns as they have similar rows

git-jano: if you prefer structured lists to table, it would be nice to have daily reports collapsable (but show a few most recent)

PR #177, opened 2018-08-15
Issue 175 keyboard shortcuts

### LINKED ISSUE #175: Optimize page navigation with keyboard
Since we use custom elements, navigation with Tab may not always work properly. For example, when importing JSON keyfile, enter for password prompt does not work. It would be nice to check other elements, too.

git-jano:
closes #175


refi93:
The tabs in the initial page (Access Cardano Wallet via) aren't reachable by tabs and enter. Also the "select key file" button in the "Key File" tab is not reachable. It would be nice if they were if we want our site to be actually navigable by keyboard.

app/frontend/components/pages/login/loginPage.js
```diff
@@ -78,14 +63,10 @@ class LoginPage extends Component {
 
 module.exports = connect(
   (state) => ({
-    mnemonic: state.mnemonic,
```
refi93: why those properties were removed from there. It isn't clear for me at first sight

app/frontend/actions.js
```diff
@@ -141,6 +141,7 @@ module.exports = ({setState, getState}) => {
       mnemonic: CARDANOLITE_CONFIG.CARDANOLITE_DEMO_WALLET_MNEMONIC,
       mnemonicValidationError: undefined,
       walletLoadingError: undefined,
+      focus: 'goBtn',
```
refi93: the name 'focus' created the impression that the element is focused while this variable is set to 'goBtn' but actually goBtn remains focused even when focus is undefined, after calling clearFocus in components.js. Therefore I would choose a less generic name for this property, perhaps "focusTo"?

app/frontend/components/pages/login/mnemonicAuth.js
```diff
@@ -1,95 +1,144 @@
...
-    if (e.button === 0) {
-      action()
+const isLeftClick = (e, action) => {
+  if (e.button === 0) {
+    e.preventDefault()
+    action()
+  }
+}
+
+class LoadByMenmonicSectionClass extends Component {
```
refi93: typo: Menmonic

refi93: the "class" suffix is imho redundant

app/frontend/components/pages/receive_ada/receiveAdaPage.js
```diff
@@ -11,8 +9,8 @@ const OwnAddressesList = connect('ownAddressesWithMeta')(({ownAddressesWithMeta}
     h('h2', undefined, 'My Addresses'),
     ...ownAddressesWithMeta.map((adr) =>
       h(AddressContainer, {address: adr.address, bip32path: adr.bip32StringPath})
-    ),
-    h(AddressDetailDialog)
+    )
+    // h(AddressDetailDialog) not needed anymore, as it is invoked from footer
```
refi93: then it would be probably more suitable to move the AddressDetailDialog to some higher-level scope, like components/app.js and have this dialog between "common" components

app/frontend/components/pages/receive_ada/addressDetailDialog.js
```diff
@@ -1,96 +1,125 @@
...
-      {class: 'overlay fade-in-up'},
-      !showVerification &&
-        h('div', {
-          class: 'overlay-close-layer',
-          onClick: closeAddressDetail,
-        }),
+class AddressDetailDialogClass extends Component {
+  componentDidUpdate() {
+    this.props.showDetail && this.copyBtn && this.copyBtn.focus()
+    // known bug: Trezor emulator steals the focus
```
refi93: is it fixable?

app/frontend/components/pages/receive_ada/addressDetailDialog.js
```diff
@@ -1,96 +1,125 @@
...
-                    {
-                      href: `https://cardanoexplorer.com/address/${showDetail.address}`,
-                      target: '_blank',
-                      class: 'address-link margin-1rem centered-row',
+                  'div',
+                  {class: 'centered-row'},
+                  h(CopyOnClick, {
+                    value: showDetail.address,
+                    tabIndex: 1,
+                    reference: (element) => {
```
refi93: I would rename this to copyBtnRef

refi93: I would try assigning this "h(CopyOnClick...)" to a variable, let's say copyOnClick and maybe then you would be able to call focus on it, without this weird passing of reference

git-jano:
After a discussion, we've decided to keep keyboard support only in modals and forms to avoid complications in furee sustaining of the project

app/frontend/actions.js
```diff
@@ -201,12 +201,12 @@ module.exports = ({setState, getState}) => {
   }
 
   const openAddressDetail = (state, {address, bip32path}) => {
-    const showAddressVerification = state.usingTrezor
+    const showAddressVerification = state.usingTrezor && bip32path
```
refi93: I would explain in a comment that this is done to avoid  trezor checking for donation address

app/frontend/actions.js
```diff
@@ -201,12 +201,12 @@ module.exports = ({setState, getState}) => {
...
 
   const openAddressDetail = (state, {address, bip32path}) => {
-    const showAddressVerification = state.usingTrezor
+    const showAddressVerification = state.usingTrezor && bip32path //due to donation address
     setState({
       showAddressDetail: {address, bip32path},
       showAddressVerification,
     })
-    if (showAddressVerification) {
+    if (showAddressVerification && bip32path) {
```
refi93: this && bip32Path is redundant

PR #181, opened 2018-08-23
Issue 180 implement derivation scheme v2

refi93:
implementation of derivation scheme v2, waiting for refactor and for adding Trezor support


app/frontend/wallet/constants.js
```diff
@@ -10,13 +10,33 @@ module.exports = {
   */
   TX_SIGN_MESSAGE_PREFIX: '011a2d964a095820',
```
ddeath: is this still in use? Also this is hardcoded in trezor should we change that?

app/frontend/wallet/helpers/isValidAddress.js
```diff
@@ -8,7 +8,9 @@ const isValidAddress = (address) => {
     const addressAsBuffer = cbor.decode(base58.decode(address))[0].value
     const addressData = cbor.decode(addressAsBuffer)
     const addressAttributes = addressData[1]
-    cbor.decode(addressAttributes.get(1))
+    if (addressAttributes[1]) {
```
ddeath: Here should be used validation function from cardano-crypto.js

PR #201, opened 2018-11-02
show dollar/euro equivalent

### LINKED ISSUE #199: Show dollar/euro equivalent for amounts

mirecmrozek:
closes #199
getting conversion rates from [CoinGecko](https://www.coingecko.com)

### DIFF
--- a/app/frontend/actions.js
+++ b/app/frontend/actions.js
@@ -10,6 +10,7 @@ const {
 } = require('./helpers/validators')
 const printAda = require('./helpers/printAda')
 const debugLog = require('./helpers/debugLog')
+const getConversionRates = require('./helpers/getConversionRates')
 const sleep = require('./helpers/sleep')
 const {ADA_DONATION_ADDRESS} = require('./wallet/constants')
 const NamedError = require('./helpers/NamedError')
@@ -109,6 +110,7 @@ module.exports = ({setState, getState}) => {
       const ownAddressesWithMeta = await wallet.getOwnAddressesWithMeta()
       const transactionHistory = await wallet.getHistory()
       const balance = await wallet.getBalance()
+      const conversionRates = getConversionRates(state)
       const sendAmount = {fieldValue: ''}
       const sendAddress = {fieldValue: ''}
       const sendResponse = ''
@@ -129,6 +131,16 @@ module.exports = ({setState, getState}) => {
         showDemoWalletWarningDialog: isDemoWallet,
         showGenerateMnemonicDialog: false,
       })
+      try {
+        setState({
+          conversionRates: await conversionRates,
+        })
+      } catch (e) {
+        debugLog('Could not fetch conversion rates.')
+        setState({
+          conversionRates: null,
+        })
+      }
     } catch (e) {
       debugLog(e.toString())
       setState({
@@ -232,6 +244,7 @@ module.exports = ({setState, getState}) => {
     const balance = await wallet.getBalance()
     const ownAddressesWithMeta = await wallet.getOwnAddressesWithMeta()
     const transactionHistory = await wallet.getHistory()
+    const conversionRates = getConversionRates(state)
 
     // timeout setting loading state, so that loading shows even if everything was cached
     setTimeout(() => setState({loading: false}), 500)
@@ -240,6 +253,16 @@ module.exports = ({setState, getState}) => {
       ownAddressesWithMeta,
       transactionHistory,
     })
+    try {
+      setState({
+        conversionRates: await conversionRates,
+      })
+    } catch (e) {
+      debugLog('Could not fetch conversion rates.')
+      setState({
+        conversionRates: null,
+      })
+    }
   }
 
   const toggleAboutOverlay = (state, rememberInStorage) => {
--- a/app/frontend/components/common/balance.js
+++ b/app/frontend/components/common/balance.js
@@ -1,5 +1,6 @@
 const {h} = require('preact')
 const printAda = require('../../helpers/printAda')
+const printConversionRates = require('../../helpers/printConversionRates')
 const {AdaIcon, RefreshIcon} = require('./svg')
 
 const truncatePrintAda = (amount, maxDigits) => {
@@ -13,7 +14,7 @@ const truncatePrintAda = (amount, maxDigits) => {
   return `${Math.trunc(amountString / 1000)}K`
 }
 
-const Balance = ({balance, reloadWalletInfo}) =>
+const Balance = ({balance, reloadWalletInfo, conversionRates}) =>
   h(
     'div',
     {class: 'balance-block'},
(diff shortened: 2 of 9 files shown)


app/frontend/helpers/getConversionRates.js
```diff
@@ -0,0 +1,10 @@
+const request = require('../wallet/helpers/request')
+
+const getConversionRates = async function getConversionRates() {
```
refi93: I would add caching so each refresh of balance would not trigger an API fetch. I would create a private fetchConversionRates() method that performs the actual fetch and expose getConversionRates() which would then cache the result for let's say 5 minutes. You can inspire by how getAddressInfo and fetchAddressInfo are made in blockchain-explorer.js

app/frontend/actions.js
```diff
@@ -233,13 +236,15 @@ module.exports = ({setState, getState}) => {
     const balance = await wallet.getBalance()
     const ownAddressesWithMeta = await wallet.getOwnAddressesWithMeta()
     const transactionHistory = await wallet.getHistory()
+    const conversionRates = await getConversionRates()
```
refi93: This is probably a bit dangerous since you don't catch possible exceptions if fetching of the rates fails. I would wrap this into a try-catch block and if fetching fails, I would set the rates to null. If the rates are null, they would not simply display, since it is not critical.

ppershing: Yep. We even might want to fetch them in parallel so that slow request doesn't block the main wallet

app/frontend/components/common/balance.js
```diff
@@ -13,7 +13,15 @@ const truncatePrintAda = (amount, maxDigits) => {
   return `${Math.trunc(amountString / 1000)}K`
 }
 
-const Balance = ({balance, reloadWalletInfo}) =>
+const printConversionRates = (amount, conversionRates, maxDigits) =>
+  isNaN(Number(amount))
+    ? ''
+    : maxDigits
+      ? `(${truncatePrintAda(printAda(conversionRates.usd * amount), 12)} $,
```
ppershing: I am pretty sure `truncatePrintAda(printAda(` combo is not what you want for formatting of fiat currencies. maybe create a separate `formatFiat` function that takes in as an agrument number of decimals?

app/frontend/helpers/getConversionRates.js
```diff
@@ -0,0 +1,10 @@
...
+  const conversionRates = await request(
+    'https://api.coingecko.com/api/v3/coins/cardano?tickers=false&community_data=false&developer_data=false&sparkline=false'
```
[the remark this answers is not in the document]
refi93: perhaps this endpoint is more suitable: https://api.coingecko.com/api/v3/coins/markets?vs_currency=usd&ids=cardano and https://api.coingecko.com/api/v3/coins/markets?vs_currency=eur&ids=cardano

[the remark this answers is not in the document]
mirecmrozek: Thank you for pointing it out!
I found a better Api endpoint: https://min-api.cryptocompare.com/data/price?fsym=USD&tsyms=USD,EUR
Only 29B.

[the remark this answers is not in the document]
refi93: The cryptocompare api looks nice, but @ppershing raised a good point about the terms of service: https://www.cryptocompare.com/api/ - in the case of cryptocompare, if I understand right, we should put a reference to it on our site, which is nevertheless good to do to be transparent about what data we use. I see two suitable locations for the link - either next to the balance or in the footer of the page.

app/frontend/components/pages/txHistory/transactionHistory.js
```diff
@@ -62,6 +66,30 @@ const TransactionHistory = ({transactionHistory}) =>
                 {class: 'align-right'},
                 transaction.effect < 0 ? h('pre', undefined, printAda(transaction.fee)) : ''
               ),
+              h(
```
refi93: adding those fields to the transaction history makes it too cluttered imho. I would rather keep the original structure and have the EUR/USD amounts as titles/tooltips of the the amounts in ADA triggered on hover and on click, so it works properly on mobiles, too. There is already a tooltip element for that in "common" folder

app/frontend/helpers/getConversionRates.js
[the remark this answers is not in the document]
mirecmrozek: Right, going to add it to the Footer.
I think "Conversion rates from (link)" would be better as it is shorter and just as clear. What do you think?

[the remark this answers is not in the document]
refi93: looks good to me

mirecmrozek:
This should solve most comments.
I added `padding-top: 1.5em` to `transaction-history-wrapper`, because tooltips on the first line of table were somehow behind the heading. 
Should i still add parallelism to the fetching of conversion rates, now that the request is much smaller?
Also, how do i pass the ci test? It says `Cannot read property 'name' of null` but i don't access `name` anywhere in this PR.

ppershing:
I would keep paralellism, mainly because even if the external service is down we want to keep running (without conversion rates)

mirecmrozek:
Added parallelism to fetching conversion rates.

refi93:
I would round the displayed amounts of USD/EUR to two decimals. Displaying it rounded to 6 decimals seems unnatural

app/frontend/actions.js
```diff
@@ -233,14 +245,26 @@ module.exports = ({setState, getState}) => {
...
     const transactionHistory = await wallet.getHistory()
+    const conversionRates = getConversionRates(state)
 
     // timeout setting loading state, so that loading shows even if everything was cached
     setTimeout(() => setState({loading: false}), 500)
     setState({
       balance,
       ownAddressesWithMeta,
       transactionHistory,
+      conversionRates,
```
refi93: this is a promise if I understand right, I'm not sure if you want to have a promise there. I think it would be safer to not set the conversionRates there at all and you could move the "const conversionRates = getConversionRates(state)" inside of the try-catch block below then.

I see one particular upside in the approach I suggest - when fetching fails, it does not reset your conversion rates to null when they were set previously. It probably makes sense to keep an older rate rather than delete it completely.

app/frontend/actions.js
```diff
@@ -233,14 +245,26 @@ module.exports = ({setState, getState}) => {
...
     })
+    try {
+      setState({
+        conversionRates: await conversionRates,
```
refi93: you are doing an  await on an object, not on a promise. I suggest rewriting it like this:

`
let conversionRates = null
try {
    conversionRates = await getConversionRates(state)
} catch (e) {
    debugLog(`Could not fetch conversion rates:  ${e}`)
}
setState({
   balance,
   state,
   etc...
`

Issue #204, opened 2018-11-06
Remove Alpha version warning from site


MichalPetro:
As we are already stable and fully featured we can remove all mentions of Alpha stage.

- [x] Add explicit versioning 1.0.0 (small font under the logo)

- [x] Change "Try the Alpha now" buttons on the landing page to "Enter wallet" (2 times)

- [x] Remove Alpha letter from name on https://adalite.io/app

- [x] Change "AdaLite Wallet" name on https://adalite.io/app to AdaLite

- [x] On Landing page, change Try Now in top right corner to Wallet

MichalPetro:
As we are already stable and fully featured we can remove all mentions of Alpha stage.

- [ ] Add explicit versioning 1.0.0 (small font under the logo)

- [ ] Change "Try the Alpha now" buttons on the landing page to "Enter wallet" (2 times)

- [ ] Remove Alpha letter from name on https://adalite.io/app

- [ ] Change "AdaLite Wallet" name on https://adalite.io/app to AdaLite

PR #212, opened 2018-11-07
Remove alpha

mirecmrozek:
Closes https://github.com/vacuumlabs/adalite/issues/204.
I used inline styles for styling the version number on landing page, because i didn't know, where to put them.

### DIFF
--- a/app/frontend/components/common/navbar/navbarAuth.js
+++ b/app/frontend/components/common/navbar/navbarAuth.js
@@ -2,6 +2,7 @@ const {h} = require('preact')
 const connect = require('unistore/preact').connect
 const actions = require('../../../actions')
 const {ExitIcon, AdaLiteLogo, MenuIcon} = require('../svg')
+const APP_VERSION = require('../../../config').ADALITE_CONFIG.ADALITE_APP_VERSION
 
 const LoginStatus = connect(
   (state) => ({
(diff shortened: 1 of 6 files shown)


app/frontend/components/common/navbar/navbarAuth.js
```diff
@@ -51,11 +51,10 @@ const NavbarAuth = connect(
           onClick: () => window.history.pushState({}, 'txHistory', 'txHistory'),
         },
         h(AdaLiteLogo),
-        h('span', undefined, 'AdaLite'),
-
-        h('sup', undefined, '⍺')
+        h('span', undefined, 'AdaLite')
       ),
       isDemoWallet ? h('div', {class: 'public-wallet-badge pulse'}, 'DEMO WALLET') : null,
+      h('div', {class: 'logo-version-number'}, '1.0.0'),
```
refi93: I would fetch the app version from package.json. I suggest using the one in the root directory. I would remove the "version" field from the second package.json (the one in the "app" directory) to avoid confusion.

The helper to get the app version could be placed in app/helpers directory and could look something like this (tested and working):

```
const {version: appVersion} = require('../../../package.json')

export default appVersion
```

I'm aware that this approach would not work for the landing page, but I see no point in displaying the version there, so I would simply remove it from there

landing_page/index.html
```diff
@@ -73,7 +73,10 @@
                     aria-controls="navbar">
                     <i></i>
                 </a>
-                <a class="navbar-brand" href="index.html">AdaLite Wallet</a>
+                <div style="display: flex; flex-direction: column;">
+                  <a class="navbar-brand" style="padding-bottom: 2px;" href="index.html">AdaLite Wallet</a>
+                  <span class="navbar-brand" style="font-size: 18px; cursor: default; padding: 0; margin: 0;height: 10px; bottom: 14px; position: relative;">1.0.0</span>
```
refi93: As I wrote above, I would remove the version from there. It's unnecessary and easy to forget to maintain.

PR #216, opened 2018-11-08
Facelift

PeterPsota:
addresses points 3 - 8 from #208 :

- Logging in(defaulting to) `TxHistoryPage` updates url
- Scrollbar issues resolved with `calc(100vw - 100%)` from https://stackoverflow.com/a/30293718
- Consistent spacing between footer items' texts
- Narrower(`360px`) inputs in `/exportWallet`
- Disclaimer title now separate, centered and bold
- `fade-in-up` removed from modals, kept in `LoginPage` title and `MnemoticAuth` 'generate new' link

### DIFF
--- a/app/frontend/components/common/addressDetailDialog.js
+++ b/app/frontend/components/common/addressDetailDialog.js
@@ -19,7 +19,7 @@ class AddressDetailDialogClass extends Component {
       h(
         'div',
         {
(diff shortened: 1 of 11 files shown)


refi93:
I would remove the fade-in-up completely

PeterPsota:
Also removed `rise-up` as it was only used in `fade-in-up`

PR #219, opened 2018-11-08
Fix trezor version check

### LINKED ISSUE #217: Fix trezor firmware version check
It incorrectly identifies version 2.0.10 as obsolete, since it does an alphhabetical comparison.

refi93:
closes #217 

the problem of the previous check was that it failed to recognize version 2.0.10 (not released yet) as higher since there was a simple alphabetical comparison of version strings so I replaced it with a more robust one.

### DIFF
--- a/app/frontend/wallet/helpers/alertIfUnsupportedTrezorFwVersion.js
+++ b/app/frontend/wallet/helpers/alertIfUnsupportedTrezorFwVersion.js
@@ -9,7 +9,7 @@ async function alertIfUnsupportedTrezorFwVersion() {
     const minSupportedFwVersion = '2.0.8'
     const {model, fwVersion} = await getTrezorInfo()
 
-    if (model === 'T' && fwVersion < minSupportedFwVersion) {
+    if (model === 'T' && versionCompare(fwVersion, minSupportedFwVersion) < 0) {
       // eslint-disable-next-line no-alert
       alert(`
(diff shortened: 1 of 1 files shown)


app/frontend/wallet/helpers/alertIfUnsupportedTrezorFwVersion.js
```diff
@@ -9,7 +9,7 @@ async function alertIfUnsupportedTrezorFwVersion() {
     const minSupportedFwVersion = '2.0.8'
     const {model, fwVersion} = await getTrezorInfo()
 
-    if (model === 'T' && fwVersion < minSupportedFwVersion) {
+    if (model === 'T' && versionCompare(fwVersion, minSupportedFwVersion) < 0) {
```
PeterPsota: `versionCompare` can return NaN, making `Nan < 0` resolve as false and not display the alert even tho something probably went wrong(?)

refi93: fair point, and I would say that's the desired behavior in that case, so I would leave it as-is. This whole code is meant to cause as little hassle as possible if something goes wrong since it is basically a hotfix to notify people that their trezor version does not match which the trezor interface currently fails to state explicitly, only suggesting you to update your firmware without further explanation.

PR #221, opened 2018-11-08
Auto logout

PeterPsota:
Didn't find anything that looked good as is, this solution is based on https://stackoverflow.com/a/10126042
Made component that resets 15minute timer on common actions and logs out user after 15m of inaction.
Also works on mobile (tested from different browser tab and app, and while screen off).


app/frontend/components/autoLogout.js
```diff
@@ -0,0 +1,72 @@
...
+  logout() {
+    this.props.closeDemoWalletWarningDialog()
+    this.props.setLogoutNotificationOpen(true)
+    this.props.logout()
+  }
+
+  resetTimer() {
+    if (this.props.walletIsLoaded) {
+      clearTimeout(this.timer)
+      this.timer = setTimeout(this.logout, 15 * 60 * 1000)
```
refi93: this should probably be tweakable in .env by something like ADALITE_LOGOUT_AFTER and it could be the amount of seconds. If it is not set, it could be interpreted as no timeout at all.

PeterPsota:
Anything falsey or numbers <=0 will disable the feature completely

app/frontend/components/app.js
```diff
@@ -7,6 +7,8 @@ const LoadingOverlay = require('./common/loadingOverlay')
 const NavbarAuth = require('./common/navbar/navbarAuth')
 const NavbarUnauth = require('./common/navbar/navbarUnauth')
 const AddressDetailDialog = require('./common/addressDetailDialog')
+const AutoLogout = require('./autoLogout')
+const ADALITE_LOGOUT_AFTER = require('../config').ADALITE_CONFIG.ADALITE_LOGOUT_AFTER
```
refi93: I would use
'''
const {ADALITE_LOGOUT_AFTER}
'''

server/helpers/loadFrontendConfig.js
```diff
@@ -20,4 +20,5 @@ module.exports = {
   ADALITE_ENABLE_DEBUGGING: process.env.ADALITE_ENABLE_DEBUGGING === 'true',
   ADALITE_ENABLE_TREZOR: !!process.env.ADALITE_ENABLE_TREZOR,
   ADALITE_APP_VERSION: appVersion,
+  ADALITE_LOGOUT_AFTER: process.env.ADALITE_LOGOUT_AFTER * 1000,
```
refi93: I would rather multiply it by 1000 directly in setTimeout. Modifying the value there seems a bit confusing to me. I would use parseInt for the sake of consistency and check ADALITE_LOGOUT_AFTER for falsiness rather than for >0 in app.js

Issue #203, opened 2018-11-05
Align transaction times


ppershing:
![image](https://user-images.githubusercontent.com/837681/48030040-26d61e00-e150-11e8-8f52-02ba768ed1bd.png)

PR #224, opened 2018-11-09
format date in TransactionHistory to have same length

mirecmrozek:
image
Closes https://github.com/vacuumlabs/adalite/issues/203

### DIFF
--- a/app/frontend/helpers/formatDate.js
+++ b/app/frontend/helpers/formatDate.js
@@ -1,0 +1,12 @@
+const formatDate = (date) =>
+  new Date(date * 1000).toLocaleString(undefined, {
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    day: '2-digit',
+    month: '2-digit',
+    year: 'numeric',
+    hour12: false,
+  })
+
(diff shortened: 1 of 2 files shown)


app/frontend/helpers/formatDate.js
```diff
@@ -0,0 +1,10 @@
+const formatDate = (date) => {
+  const timeArray = new Date(date * 1000).toLocaleString().split(',')
+  const numberOfSpaces = 10 - timeArray[0].length
+  const numberOfSpaces2 = 12 - timeArray[1].length
```
refi93: this may fail for locales different that en-US, let's say the slovak one

refi93: other solution could be to force en-US locale but I would prefer to use browser's locale for optimal UX

refi93:
btw, my browser has a non-monospace font for sans-serif and it ends up looking like this:
screenshot from 2018-11-09 13-21-31

app/frontend/helpers/formatDate.js
mirecmrozek: Yup, totally agree, and it even looks better.

PR #236, opened 2018-11-09
Added content security policy middleware

### LINKED ISSUE #183: Research & Implement Content Security Policy
MyEtherWallet may be a good inspiration. For example, they block developer console, if you open chrome devtools. Other things may be relevant to us.

PeterPsota:
closes #183 
Added Content-Security-Policy header with middleware

the directives could be made more secure by:
- restricting `connect-src` to only sources we use (impact on maintainability)
- refactoring app in pursuit of removing `unsafe-eval` and maybe scripts `unsafe-inline` (eval seems more complex then just refactoring `init.js`, the inline scripts might be possible, inline css probably isn't)

### DIFF
--- a/app/public/js/init.js
+++ b/app/public/js/init.js
@@ -1,34 +1,58 @@
-(function() {
-  try {
-    // test for JS features needed to load the main script
-    eval( // eslint-disable-line no-eval
-      '(function() {' +
-        // destructuring binding
-        '({a, b} = {a: 10, b: 20, c: 30, d: 40});' +
-        // const, let
-        'const c1 = 5; let var1 = 5;' +
-        // async functions
-        'async function f1 (){return 9};' +
-        // arrow functions
-        '() => {return 4 + 5};' +
-        // template literal syntax
-        '`${"aaa"}`;' + // eslint-disable-line no-template-curly-in-string
-        // default parameters in function
-        'function f2 (a = 3, b, c=false) {return a}' +
-        // classes
-        'class MyClass {};' +
-        // assign object keys from variable
-        'let objKey = "a"; let obj = {[objKey]: 3};' +
-      '})()'
-    )
+// source: https://gist.github.com/DaBs/89ccc2ffd1d435efdacff05248514f38#gistcomment-1976435
+function playsNice(cb) {
+  const testScript =
+    '(function() {' +
+      // destructuring binding
+      '({a, b} = {a: 10, b: 20, c: 30, d: 40});' +
+      // const, let
+      'const c1 = 5; let var1 = 5;' +
+      // async functions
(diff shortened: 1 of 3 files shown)


refi93:
perhaps in this PR I address most of the problem with unsafe eval: https://github.com/vacuumlabs/adalite/pull/237 by removing it from the bundle, could you take a look at init.js and how to avoid eval there?

PeterPsota:
I found eval free CSP friendly feature detection method and removed the `unsafe-eval` directive.

I also apparently didn't rebase correctly(i think ill keep merge), should I make a new, cleaner PR?

refi93:
I would just force push the cleaned version to this PR.

PeterPsota:
Ok, I think I figured it out.

app/public/js/init.js
```diff
@@ -1,34 +1,58 @@
...
+    cb(!(event && event.error))
+  }
+
+  window.addEventListener('error', n.ondone, true)
+  n.appendChild(document.createTextNode(x))
+  document.head.appendChild(n)
+  setTimeout(n.ondone, 50)
+}
 
+playsNice((bool) => {
```
refi93: I would rename "bool" to "success", to make it clearer

app/public/js/init.js
```diff
@@ -1,34 +1,58 @@
...
-        'function f2 (a = 3, b, c=false) {return a}' +
-        // classes
-        'class MyClass {};' +
-        // assign object keys from variable
-        'let objKey = "a"; let obj = {[objKey]: 3};' +
-      '})()'
-    )
+// source: https://gist.github.com/DaBs/89ccc2ffd1d435efdacff05248514f38#gistcomment-1976435
+function playsNice(cb) {
+  const x =
```
refi93: I would rename "x" to "testScript"

app/public/js/init.js
refi93: arrow notation would fail on legacy browsers, tried on internet explorer 11

app/public/js/init.js
```diff
@@ -1,34 +1,58 @@
...
+    s = this
+    if (s.done) {
+      window.removeEventListener('error', s.ondone, true)
+      if (s.parentNode) {
+        s.parentNode.removeChild(s)
+      }
+      return
+    }
+    this.done = 1
+    cb(!(event && event.error))
```
refi93: I would change the condition to 
```!(event && (event.error || event.type === 'error' || event.lineno))```
I tried it in IE11 and the returned ErrorEvent has the error property set to undefined :D but there are additional properties provided, like "type" and "lineno" which give away the error, so let's keep it broader.

PeterPsota:
Thanks for the review, should be ok now.

PR #239, opened 2018-11-13
don't display public key on trezor when exporting it

### LINKED ISSUE #238: Don't display public key on trezor when exporting it

refi93:
closes #238


refi93:
Just a little UX improvement. I think there is need to display the public key on the device since the user has no practical way of verifying it, nor is it a gain in security afaik. Neither myetherwallet does it.

PR #243, opened 2018-11-14
Fix footer style

### LINKED ISSUE #234: Rework footer
Update footer

- [x] so that it is always on the bottom of the page (unless the content pushes it down) - like on main page https://adalite.io/

- [x] change the twitter hyperlink name to "Twitter"

- [x] Increase size of the font of social media hyperlinks

- [x] Decrease the font of Crypto compare line

- [x] Make sure that cryptocompare line displays in footer only on history screen

PeterPsota:
closes #234 
Footer sticks to bottom of page
Twitter link now 'Twitter'
Social media link bigger- no longer in `<small>`
Conversion rates line 15% smaller and only visible when on history page

### DIFF
--- a/app/frontend/components/common/footer.js
+++ b/app/frontend/components/common/footer.js
@@ -10,10 +10,14 @@ const {
 } = require('../../wallet/constants')
 const {VacuumlabsLogo} = require('./svg')
 
+const showRatesOn = ['/txHistory', '/send']
+
 const Footer = connect(
-  {},
+  (state) => ({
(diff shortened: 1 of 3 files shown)


app/frontend/components/common/footer.js
```diff
@@ -11,9 +11,11 @@ const {
 const {VacuumlabsLogo} = require('./svg')
 
 const Footer = connect(
-  {},
+  (state) => ({
+    showConversionRates: state.router.pathname === '/txHistory' && state.walletIsLoaded,
```
refi93: the "send" screen also contains the coversion rates

PR #246, opened 2018-11-15
Update landing

mirecmrozek:
Now that section with stats is gone, there are two sections (More to Come, Contact) next to each other with white background so I'm still looking into the css.


refi93:
@mirecmrozek so it's still WIP, right? Btw I took a look again at the landing page and I realized, in the "More to come" section has also outdated info. I would just ditch it, so perhaps it even solves your whitespace problem :smile:

mirecmrozek:
I does! :D

PR #247, opened 2018-11-16
Only add usage from adalite.com

Siegrift:
I don't know hot to test this on production, though :/ 

I believe host should be 'https://adalite.io/', but I don't know for sure.
@refi93 any ideas? 

Also another approach could be to use environment variables, and don't count usage if the app is running in dev mode.

### DIFF
--- a/server/middlewares/stats.js
+++ b/server/middlewares/stats.js
@@ -2,6 +2,7 @@ const redis = require('redis')
 const device = require('device')
 const client = redis.createClient(process.env.REDIS_URL)
 const mung = require('express-mung')
+const normalizeUrl = require('normalize-url')
 
 const knownIps = new Set()
 
@@ -10,6 +11,13 @@ function getSlicedDate() {
     .toISOString()
     .split('T')[0]
     .split('-')
+}
+
+const isSameOrigin = (urlString1, urlString2) => {
+  return (
+    normalizeUrl(urlString1, {stripProtocol: true}) ===
+    normalizeUrl(urlString2, {stripProtocol: true})
+  )
 }
 
 const trackVisits = (req, res, next) => {
@@ -33,7 +41,11 @@ const trackVisits = (req, res, next) => {
 const trackTxSubmissionCount = mung.json((body, req, res) => {
   if (req.originalUrl === '/api/txs/submit' && req.method === 'POST') {
     const [year, month, day] = getSlicedDate()
-    const key = `txSubmissions:${body.Right ? 'successful' : 'unsuccessful'}`
+    const key = `${
+      isSameOrigin(req.get('origin'), process.env.ADALITE_SERVER_URL)
+        ? 'txSubmissions'
+        : 'otherTxSubmissions'
+    }:${body.Right ? 'successful' : 'unsuccessful'}`
 
     client.incr(`${key}:total`)
     client.incr(`${key}:monthly:${year}-${month}`)
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "express-basic-auth": "^1.1.5",
     "express-mung": "^0.5.1",
     "isomorphic-fetch": "^2.2.1",
+    "normalize-url": "^4.0.0",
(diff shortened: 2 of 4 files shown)


server/middlewares/stats.js
```diff
@@ -31,7 +31,11 @@ const trackVisits = (req, res, next) => {
 }
 
 const trackTxSubmissionCount = mung.json((body, req, res) => {
-  if (req.originalUrl === '/api/txs/submit' && req.method === 'POST') {
+  if (
+    req.header('host') === 'https://adalite.io' &&
```
refi93: I think you should check the 'origin' header

refi93:
@Siegrift you can take the "ADALITE_SERVER_URL" value from .env which is the url adalite is hosted on

server/middlewares/stats.js
```diff
@@ -31,7 +31,11 @@ const trackVisits = (req, res, next) => {
...
+    req.originalUrl === '/api/txs/submit' &&
+    req.method === 'POST'
+  ) {
```
refi93: The purpose is not to ignore requests coming from other sources, just record them apart under the key, let's say `otherTxSubmissions` with the same subkeys (successful/unsuccessful, daily/monthly/total)

server/middlewares/stats.js
refi93: And I would rather make a helper function `isSameOrigin(url1, url2)` or something like that, which would take the urls as strings, transform the to URL objects (https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) which does the parsing for you and you just compare the `origin` attributes of them which are stripped of any redundant slashes which may cause the comparison to return false negatives.

refi93: I would also use some url normalization, to strip "www" from the compared urls: https://www.npmjs.com/package/normalize-url Feel free to add it as a server dependency

Siegrift: Is it really needed? I only see the point if comparing URL's with params, but comparing url origins (after wrapping them in URL object) should be sufficient.

refi93: I think so, the 'www' prefix would pass as a false negative, here is a code sample showing what I mean:
```
new URL('https://www.adalite.io').origin
-> "https://www.adalite.io"
new URL('https://adalite.io').origin
-> "https://adalite.io"
```

server/middlewares/stats.js
```diff
@@ -12,6 +12,12 @@ function getSlicedDate() {
     .split('-')
 }
 
+const isSameOrigin = (urlString1, urlString2) => {
+  const url1 = new URL(urlString1)
+  const url2 = new URL(urlString2)
+  return url1.origin === url2.origin
```
refi93: I think the `host` attribute is safer since it strips the `https://` part which isn't really a problem now, since we do https redirect, but I would rather ignore it in the comparison to stay on the safe side.

refi93: although well, calling a function `isSameOrigin` and comparing `host`s seems a bit fishy to me, I leave it up to your taste, perhaps adding a comment explaining the rationale behind using `host` instead of `origin` would do it

Siegrift: I would rather not use external library for this so I chose the first option.

server/middlewares/stats.js
Siegrift: I see.

server/middlewares/stats.js
refi93: but this still has the problem that if somebody accesses adalite.io through www.adalite.io and makes a transaction, the backend would wrongly recognize it as a foreign transaction

refi93: but a regex would do it as well, of course: https://stackoverflow.com/questions/3738282/javascript-strip-words-from-url

Siegrift: Thanks for the example, I see now why you wanted to normalize the URL's. Probably the easiest solution is to use the library.

PR #249, opened 2018-11-16
refactor overlay into general Modal component

mirecmrozek:
Created a general Modal component to make code more clear and adding Modals easier.
Refactored every component using css class 'overlay' to use Modal, except `LoadingOverlay', because it's too different from a typical modal.

### DIFF
--- a/app/frontend/components/pages/login/aboutOverlay.js
+++ b/app/frontend/components/pages/login/aboutOverlay.js
@@ -1,6 +1,7 @@
 const {h, Component} = require('preact')
 const connect = require('unistore/preact').connect
 const actions = require('../../../actions')
+const Modal = require('../../common/modal')
 
 class AboutOverlayClass extends Component {
   constructor(props) {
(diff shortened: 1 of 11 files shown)


app/frontend/components/pages/login/keyFileAuth.js
```diff
@@ -158,67 +158,49 @@ class LoadKeyFileClass extends Component {
...
         encrypted &&
           h(
-            'div',
+            Modal,
             {
-              class: 'overlay',
-              onKeyDown: (e) => {
-                e.key === 'Escape' && this.closePasswordModal()
-              },
+              closeHandler: this.closePasswordModal,
```
refi93: the password modal when logging in by json now does not close by pressing esc button

refi93:
when I click on the disclaimer on the "I understand the risk ..." checkbox, it does not persist after closing and the modal is shown again when I reload the page. Imho it should be stored in local storage after pressing the "close" button (not esc or clicking the "x")

mirecmrozek:
Thank you, had to move the event check to the body div.
When I checked www.adalite.io, `Esc` only closed the password modal for logging with JSON only when the submit button had focus, so now it works with the button. But because input's `onKeyDown` (or other key events) wasn't firing for `Escape` key, I set a listener, that closes the modal when `Esc` is pressed twice (once to unfocus input, second to close modal).
I also changed where the `showAboutOverlay` value is checked, to be more like other components.
Right now, the Disclaimer modal is also not immediately closable by `Esc` and close button does not have focus, but I think that's a good thing, as we want the user to read the disclaimer.

app/frontend/components/pages/login/aboutOverlay.js
```diff
@@ -20,152 +21,129 @@ class AboutOverlayClass extends Component {
...
+          'I understand the risk and do not wish to be shown this screen again'
+        )
+      ),
+      h(
+        'span',
+        {class: 'centered-row'},
+        h(
+          'button',
+          {
+            onClick: toggleAboutOverlay,
```
refi93: this should be this.toggleAboutOverlay, otherwise the checkbox won't be saved

mirecmrozek: This, and I also accidentally deleted the line, that changed the checkbox state...

PR #253, opened 2018-11-19
Disable overlay over transaction screen

Siegrift:
UX: Instead of showing a modal dialog, show an inline spinner (same as in calculate fee) and display a transparent overlay to disable UI interaction.

### DIFF
--- a/app/frontend/actions.js
+++ b/app/frontend/actions.js
@@ -413,13 +413,23 @@ module.exports = ({setState, getState}) => {
   }
 
   const submitTransaction = async (state) => {
-    loadingAction(state, 'Submitting transaction...')
+    if (state.usingTrezor) {
+      setState({waitingForTrezor: true})
+    } else {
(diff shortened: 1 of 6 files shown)


refi93:
the modal dialog is essential since it lets you review the transaction before submitting it, so I would not remove it. I would do the spinning animation on the button in the modal. And in case of trezor it would probably be more appropriate to give the button the title "Waiting" until the user confirms the tx in the device.

refi93:
the "waiting for trezor" text shows even if you are logged in by mnemonic

refi93:
closes #256

refi93:
@Siegrift seems ok, I just wanted to point out that it would be more appropriate to show the "submitting transaction" overlay once the signed transaction is returned from trezor but I realized that this would involve splitting the "sendAda" function into building the signed tx and submitting it.

So I did the refactor and changed the UI as I suggested above. Could you please take a look at the modifications I made?

PR #262, opened 2018-12-11
Seo optimization

JMatej:
Changed html tags, added 301 redirect and other improvements for SEO


refi93:
I would rename the svg files to "css-case" - i.e. no capital letters and "-" instead of them, treating "adalite" as a single word. i.e. renaming "AdaLiteIcon.svg" to "adalite-icon.svg" etc

CamelCase for image files is not standard, imho

server/index.js
```diff
@@ -10,10 +10,7 @@ app.use(bodyParser.json())
 app.use(cors())
 app.use(compression())
 
-// must be before every other route to guarantee the redirect!
-if (process.env.ADALITE_FORCE_HTTPS === 'true') {
-  app.use(require('./middlewares/forceHttps'))
-}
+app.use(require('./middlewares/seoRedirect'))
```
refi93: you need to add the following line, so it works on heroku:
```
app.enable('trust proxy') // to get right request protocol on heroku
```
I tested the redirect on heroku and it did not work (infinite redirect loop), since it failed to pick the actual protocol of the request, it was picking the http protocol of the reverse proxy behind which our app is hosted, not the actual https protocol.

refi93: also, I would rename the middleware from seoRedirect to redirectToBaseUrl or something like that. SEO is not the only purpose of this redirect, so it would be a misleading name for it.

PR #276, opened 2018-12-14
Adding create wallet button to the toolbar

dpribula:
https://github.com/vacuumlabs/adalite/issues/270


app/frontend/components/common/navbar/navbarUnauth.js
```diff
@@ -1,7 +1,19 @@
 const {h} = require('preact')
 const APP_VERSION = require('../../../config').ADALITE_CONFIG.ADALITE_APP_VERSION
+const connect = require('unistore/preact').connect
+const actions = require('../../../actions')
 
-const NavbarUnauth = () =>
+const isLeftClick = (e, action) => {
```
refi93: I would move the "isLeftClick" function to frontend/helpers, since it's quite generic and it's used more than once.

PR #282, opened 2018-12-14
JTMarnie webfonts

### LINKED ISSUE #281: Use JTMarnie font

dejvid15:
Closes #281 

Exported & used JTMarnie webfonts


app/public/css/styles.css
```diff
@@ -38,7 +54,7 @@ html {
 body {
   margin: 0;
   padding: 0;
-  font-family: sans-serif;
+  font-family: 'JTMarnie', sans-serif, arial;
```
darjanin: The font family `sans-serif` must be the last. This way having `arial` at the end will not help.

darjanin:
```suggestion
  font-family: JTMarnie, Arial, Helvetica, sans-serif;
```

darjanin: You don't need to use quotes around name in this case.

https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#Valid_family_names

PR #303, opened 2018-12-21
Style inputs

### LINKED ISSUE #300: Styling for inputs

dejvid15:
Closes #300 

Styled `input`, `:focus` state & `error` state. Didn't change the input occurrences throughout the project as it will be changed incrementally. Checkbox is in another PR.


app/public/css/styles.css
```diff
@@ -1149,6 +1129,42 @@ p {
...
 }
 
+/* INPUTS */
+
+.input {
+  border-radius: 4px;
+  border: 1px solid var(--color-border);
+  padding: 12px 16px;
+  font-size: 14px;
+  min-height: 58px;
```
darjanin: Why `min-height`? Should it have more than one line?

dejvid15: Replaced with the `line-height` solution as discussed

dejvid15:
@darjanin Implemented your review comment and changed the class on the `mnemonicAuth` input so we have one input using the new `.input` class

PR #306, opened 2019-01-03
Cleanup old style variables

### LINKED ISSUE #305: Cleanup old variables

dejvid15:
Closes #305 

Removed old variables from the start of the file, kept only the `width, transition-easing & transition-time` and refactored the new styles to use the new variables. `width` also changed to `1288px` as in the new design.


darjanin:
Please rebase and merge.

PR #309, opened 2019-01-04
Style sidebar

### LINKED ISSUE #272: Login wrapper - Side info panel

dejvid15:
Closes #272 

Created `loginPageSidebar` component that handles the content displayed right to the access method card. Created main login page layout - left the access card inners intact so we have manageable PR sizes. Used the new type of `info Alert` - this has already a PR so after rebasing will work as expected. Also added basic styles for `.card`.

### DIFF
--- a/app/frontend/components/pages/login/loginPageSidebar.js
+++ b/app/frontend/components/pages/login/loginPageSidebar.js
@@ -1,0 +1,183 @@
+const {h, Component} = require('preact')
+const connect = require('unistore/preact').connect
+
+const Alert = require('../../../components/common/alert')
+
+class LoginPageSidebar extends Component {
+  render({authMethod}) {
+    const initialContent = () =>
+      h(
+        'div',
+        {class: 'sidebar-content'},
+        h(
+          'p',
+          {class: 'sidebar-paragraph'},
+          'AdaLite supports 3 means of accessing your wallet. For enhanced security, we recommend you to use a ',
+          h('strong', undefined, 'hardware wallet.')
+        ),
+        h(
+          'a',
+          {
+            class: 'sidebar-link',
+            href: 'https://github.com/vacuumlabs/adalite/wiki/AdaLite-FAQ#hardware-wallets',
+            rel: 'noopener',
+          },
+          'What is a Hardware Wallet'
+        ),
+        h(
+          'a',
+          {
+            class: 'sidebar-link',
+            href: 'https://wiki.trezor.io/Cardano_(ADA)',
+            rel: 'noopener',
+          },
+          'How to use Trezor T with AdaLite'
+        ),
+        h(
+          'p',
+          {class: 'sidebar-paragraph'},
+          'If you want to purchase Trezor, please consider supporting us by using our ',
+          h(
+            'a',
+            {
+              class: 'sidebar-link',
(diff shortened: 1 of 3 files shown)


app/frontend/components/pages/login/loginPageSidebar.js
```diff
@@ -0,0 +1,183 @@
+const {h, Component} = require('preact')
+const connect = require('unistore/preact').connect
+
+const Alert = require('../../../components/common/alert')
```
darjanin:
```suggestion
const Alert = require('../../common/alert')
```

app/frontend/components/pages/login/loginPageSidebar.js
```diff
@@ -0,0 +1,183 @@
...
+
+class LoginPageSidebar extends Component {
+  render({authMethod}) {
+    const initialContent = () =>
```
darjanin: Please, define these components outside of the render method for `LoginPageSidebar`. They should be components define still in this same file but render method for `LoginPageSidebar` should be:

```js
return h(
      'aside',
      {class: 'sidebar'},
      authMethod === '' && h(InitialContent),
      authMethod === 'mnemonic' && h(MnemonicContent),
      authMethod === 'trezor' && h(WalletContent),
      authMethod === 'file' && h(FileContent)
    )
```

app/public/css/styles.css
```diff
@@ -1429,3 +1429,66 @@ p {
...
+
+.page-wrapper {
+  display: flex;
+  height: 100%;
+  width: var(--width);
+  margin: 0 auto;
+}
+
+.page-main {
+  flex: 0 0 66%;
```
darjanin: This can be replaced with `width: 66%`. It's more readable and you'll have the same result.

app/public/css/styles.css
```diff
@@ -1429,3 +1429,66 @@ p {
...
+
+/* SIDEBAR */
+
+.sidebar {
+  flex: 1;
+  margin-left: 5%;
+  padding-left: 40px;
+  padding-top: 216px;
+}
+
```
darjanin: A bottom margin between items can be achieved with the `.sidebar-item` class that will wrap each section in the sidebar.

app/public/css/styles.css
```diff
@@ -1429,3 +1429,66 @@ p {
...
+.sidebar-link {
+  margin-bottom: 24px;
+  display: block;
+}
+
+.sidebar .alert {
```
darjanin: I don't want to look for styling of the alert outside of the `Alert` part of styles. So if you have a version without border and different paddings, add new modifying class to `Alert`. You can have multiple classes on `.alert`.

app/public/css/styles.css
```diff
@@ -1429,3 +1429,66 @@ p {
...
+
+.sidebar .alert:before {
+  left: 0;
+}
+
+.sidebar-paragraph {
+  margin-bottom: 64px;
+}
+
+.sidebar *:last-child {
```
darjanin: This is not needed.

app/public/css/styles.css
dejvid15: Will do

darjanin:
Please rebase and merge.

Issue #363, opened 2019-01-25
Make use of new backend submit tx endpoint


refi93:
Currently we are using a custom endpoint to submit transactions (https://github.com/vacuumlabs/adalite/blob/develop/server/transactionSubmitter.js) which does the raw tcp communication. This can be replaced with this endpoint: https://github.com/vacuumlabs/adalite-backend-service/blob/76c38ad021204b242745c1362c12c2ccf94791d4/src/routes.js#L209

j-matus:
Done in #365

PR #365, opened 2019-01-29
Redirect submits to the new backend service

j-matus:
Redirect transaction submits to the new backend service. Its endpoint
takes the body in base64 format and neither takes nor returns the hash,
therefore the body needs to be reformatted and when checking whether
the transaction was submitted successfully, the hash from signed
transaction is used instead of the hash from response. The old submitter
was not deleted yet, since it can serve as a fallback option. #363

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 ADALITE_ENABLE_DEBUGGING=false
 ADALITE_SERVER_URL = 'http://localhost:3000'
-ADALITE_BLOCKCHAIN_EXPLORER_URL = 'https://explorer.adalite.io'
+ADALITE_BLOCKCHAIN_EXPLORER_URL = 'https://explorer2.adalite.io'
 ADALITE_WALLET_ADDRESS_LIMIT = 10
 REDIS_URL = ''
 STATS_PWD =admin
--- a/server/transactionSubmitter.js
+++ b/server/transactionSubmitter.js
@@ -1,24 +1,13 @@
-const net = require('net')
-
-const cmdtable =
-  '0000040000000113841a2d964a0983000100b3048200d8184105058200d8184104068200d818410' +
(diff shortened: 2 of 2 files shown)


refi93:
I would remove the "transactionSubmitter" file completely and replace it with the "newTransactionSubmitter". In case of need, the original transactionSubmitter is in the old commits.

server/newTransactionSubmitter.js
```diff
@@ -0,0 +1,48 @@
+require('isomorphic-fetch')
+
+async function request(url, method = 'get', body = null, headers = {}) {
```
refi93: no need for this function - it is used only once, so I would prefer using fetch directly

server/newTransactionSubmitter.js
```diff
@@ -0,0 +1,48 @@
...
+        'POST',
+        JSON.stringify(signedBody),
+        {
+          'Content-Type': 'application/json; charset=utf-8',
+        }
+      )
+      return res.json(fetched)
+    } catch (err) {
+      return res.json({
+        Left: `Error: ${err}`,
```
refi93: returning the error message as-is may be too revealing. I would prefer logging it and returning just "An unexpected error has happened" - same as in the old transaction submitter.

server/index.js
```diff
@@ -29,7 +29,7 @@ app.use(require('./middlewares/csp'))
 if (process.env.ADALITE_ENABLE_SERVER_MOCKING_MODE === 'true') {
   require('./mocking')(app)
 } else {
-  require('./transactionSubmitter')(app)
+  require('./newTransactionSubmitter')(app)
```
refi93: please rename to transactionSubmitter

refi93:
please update default blockchain explorer in .env.example to explorer2.adalite.io, since the old one lacks this functionality

PR #377, opened 2019-01-31
refactor deriveXpub to eliminate repetition from crypto providers

refi93:
The purpose of this refactor is to extract the deriveXpub caching logic to a single place and to make clear the (potential) asynchronicity of the operation. I changed the name of the trezorVerifyAddress function as well, to be more abstract, since we may support more hw wallets in the future.

### DIFF
--- a/app/frontend/wallet/cardano-trezor-crypto-provider.js
+++ b/app/frontend/wallet/cardano-trezor-crypto-provider.js
@@ -1,11 +1,9 @@
-const {derivePublic} = require('cardano-crypto.js')
-const indexIsHardened = require('./helpers/indexIsHardened')
 // eslint-disable-next-line import/no-unresolved
 const {default: TrezorConnect} = require('trezor-connect')
+const CachedDeriveXpubFactory = require('./helpers/CachedDeriveXpubFactory')
 
 const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
   const state = Object.assign(walletState, {
-    derivedXpubs: {},
     rootHdPassphrase: null,
     derivedAddresses: {},
   })
@@ -14,10 +12,29 @@ const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
     throw new Error(`Unsupported derivation scheme: ${state.derivationScheme.type}`)
   }
 
-  async function trezorDeriveAddress(absDerivationPath, displayConfirmation) {
+  const deriveXpub = CachedDeriveXpubFactory(state.derivationScheme, async (absDerivationPath) => {
+    const response = await TrezorConnect.cardanoGetPublicKey({
+      path: absDerivationPath,
+      showOnTrezor: false,
+    })
+    if (response.error || !response.success) {
+      throw new Error('public key retrieval from Trezor failed')
+    }
+    return Buffer.from(response.payload.publicKey, 'hex')
+  })
+
+  function deriveHdNode(childIndex) {
+    throw new Error('This operation is not supported on TrezorCryptoProvider!')
+  }
+
+  function sign(message, absDerivationPath) {
+    throw new Error('Not supported')
+  }
+
+  async function displayAddressForPath(absDerivationPath) {
     const response = await TrezorConnect.cardanoGetAddress({
       path: absDerivationPath,
-      showOnTrezor: displayConfirmation,
+      showOnTrezor: true,
     })
 
     if (response.success) {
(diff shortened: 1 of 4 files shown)


app/frontend/wallet/cardano-trezor-crypto-provider.js
```diff
@@ -14,6 +12,30 @@ const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
     throw new Error(`Unsupported derivation scheme: ${state.derivationScheme.type}`)
   }
 
+  const deriveXpub = generateDeriveXpubFn(async (absDerivationPath) => {
+    const response = await TrezorConnect.cardanoGetPublicKey({
+      path: absDerivationPath,
+      showOnTrezor: false,
+    })
+    if (response.error || !response.success) {
+      return Promise.reject(response || 'operation failed')
```
ppershing: Why not throw instead?

app/frontend/wallet/cardano-trezor-crypto-provider.js
```diff
@@ -14,6 +12,30 @@ const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
...
+    }
+    return Buffer.from(response.payload.publicKey, 'hex')
+  }, state.derivationScheme)
```
ppershing: I wonder whether we shouldn't switch the order of arguments. This line is quite confusing

app/frontend/wallet/cardano-trezor-crypto-provider.js
```diff
@@ -14,6 +12,30 @@ const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
...
+
+  function deriveHdNode(childIndex) {
+    throw new Error('This operation is not supported on TrezorCryptoProvider!')
+  }
+
+  function sign(message, absDerivationPath) {
+    throw new Error('Not supported')
+  }
+
+  async function verifyAddressOnDevice(address, addressToAbsPathMapper) {
```
ppershing: I don't think that crypto provider should be asked to verify *address*, wouldn't path be better?
Basically, `displayAddressForPath` or something like that

app/frontend/wallet/helpers/generateDeriveXpubFn.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    return await derivedXpubs[memoKey]
+  }
+
+  async function deriveXpubNonhardenedFn(derivationPath) {
+    const lastIndex = derivationPath[derivationPath.length - 1]
+    const parentXpub = await deriveXpub(derivationPath.slice(0, derivationPath.length - 1))
+
+    return derivePublic(parentXpub, lastIndex, derivationScheme.number)
```
ppershing: Hmm, I don't like `derivePublic` name as it works only one step. What about `deriveChildPubkey` or something like that?

app/frontend/wallet/helpers/generateDeriveXpubFn.js
```diff
@@ -0,0 +1,38 @@
...
+      const deriveHardened =
+        absDerivationPath.length === 0 ||
+        indexIsHardened(absDerivationPath[absDerivationPath.length - 1])
+
+      /*
+      * TODO - reset cache if the promise fails, for now it does not matter
+      * since a failure (e.g. rejection by user) there leads to
+      * the creation of a fresh wallet and crypto provider instance
+      */
+      derivedXpubs[memoKey] = deriveHardened
```
ppershing: Why do we actually put *promise* into the map? If we do this because of concurrent requests, please add a comment there

app/frontend/wallet/helpers/generateDeriveXpubFn.js
```diff
@@ -0,0 +1,38 @@
...
+function generateDeriveXpubFn(deriveXpubHardenedFn, derivationScheme) {
+  const derivedXpubs = {}
+
+  async function deriveXpub(absDerivationPath) {
+    const memoKey = JSON.stringify(absDerivationPath)
+
+    if (!derivedXpubs[memoKey]) {
+      const deriveHardened =
+        absDerivationPath.length === 0 ||
+        indexIsHardened(absDerivationPath[absDerivationPath.length - 1])
```
ppershing: Do we include lodash? If yes, use `_.last()`, if no, add `last` funciton to utilitities

app/frontend/wallet/helpers/generateDeriveXpubFn.js
```diff
@@ -0,0 +1,38 @@
+const indexIsHardened = require('./indexIsHardened')
+const {derivePublic} = require('cardano-crypto.js')
+
+function generateDeriveXpubFn(deriveXpubHardenedFn, derivationScheme) {
```
ppershing: I don't like the name. Maybe `cachedDeriveXpubFactory` ?

app/frontend/wallet/helpers/generateDeriveXpubFn.js
ppershing: Hmm, more general question -- why does this function accept `derivationScheme.number` ? I thought that HDnode derivation is same between schemes, it is just the seed that is different

refi93: https://github.com/input-output-hk/cardano-crypto/blob/master/cbits/encrypted_sign.c#L362 - the derivation step also depends on the scheme

app/frontend/wallet/cardano-trezor-crypto-provider.js
refi93: if you just do `throw new Error(response || 'operation failed'), the "response" object gets transformed to string "[object Object]" and when it's printed to the console, the information about the exact response is lost. But I deem this redundant and the failure should be debuggable even without explicitly returning the response, so I replaced it with a throw

app/frontend/wallet/helpers/generateDeriveXpubFn.js
refi93: yes, otherwise you would end up firing multiple requests at Trezor when deriving multiple addresses.

app/frontend/wallet/helpers/generateDeriveXpubFn.js
refi93: no, we don't,I used slice instead, although this can be considered hacky

PR #378, opened 2019-02-01
Ledger integration

bynjamin:


app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
+const TransportU2F = require('@ledgerhq/hw-transport-u2f').default // for browser
+const Ada = require('ledgerhq/hw-app-ada').default // temporary hack, should be @ledgerhq
```
refi93: I would call it "LedgerAda" instead of simply "Ada"

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+      throw new Error(err)
+    }
+  }
+
+  return {
+    getWalletSecret,
+    signTx,
+    displayAddressForPath,
+    deriveXpub,
+    _sign: sign,
```
refi93: no need to export these. This exports are done only for automatized testing in order to be able to access those method, but no automatized tests are currently written for Ledger nor Trezor.

app/package.json
```diff
@@ -23,10 +23,14 @@
     "test-chrome-headless": "yarn test && node node_modules/.bin/mocha-headless-chrome -f tests/index.html"
   },
   "dependencies": {
+    "@babel/polyfill": "^7.2.5",
+    "@ledgerhq/hw-transport-u2f": "^4.7.3",
+    "base-x": "^3.0.5",
```
refi93: this isn't needed - even if you needed it, I recommed using the "base58" library exported by cardano-crypto.js https://github.com/vacuumlabs/cardano-crypto.js/blob/master/index.js#L692 to avoid adding unnecessary dependencies.

app/frontend/actions.js
```diff
@@ -104,6 +112,7 @@ module.exports = ({setState, getState}) => {
       const sendAddress = {fieldValue: ''}
       const sendResponse = ''
       const usingTrezor = cryptoProvider === 'trezor'
```
refi93: I would rather generalize this to "usingHwWallet", so no additional "or"-ing would be needed in the code below

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+      data.address58 = output.address
+    }
+
+    return data
+  }
+
+  async function prepareWitnesses(witnesses) {
+    const data = []
+    for (const witness of witnesses) {
```
refi93: I would rewrite this using Array.map function:
```
   witnesses.map((witness) => {
      ...
   })
```
It is a good practice to replace "for" cycles with map, reduce and similar functions in js, because it improves readability - it makes clearer the purpose of the iteration and it also eliminates the need to have a "data" array aside - it directly returns it, so you can assign the result to data

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+    if (output.isChange) {
+      data.path = addressToAbsPathMapper(output.address)
+    } else {
+      data.address58 = output.address
+    }
+
+    return data
+  }
+
+  async function prepareWitnesses(witnesses) {
```
refi93: I would name it "prepareWitness" to comply with the rest of the functions - prepareOutput and prepareInput, and therefore I would move the iteration to "signTx" function

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+
+  async function signTx(unsignedTx, rawInputTxs, addressToAbsPathMapper) {
+    const transactions = rawInputTxs.map((tx) => tx.toString('hex'))
+
+    const inputs = []
+    for (let i = 0; i < unsignedTx.inputs.length; i++) {
+      inputs.push(await prepareInput(unsignedTx.inputs[i], addressToAbsPathMapper, transactions[i]))
+    }
+
+    const outputs = []
```
refi93: this can be rewritten similarly, too

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+      throw new Error('Ledger operation failed!')
+    }
+  }
+
+  function getWalletSecret() {
+    throw new Error('Unsupported operation!')
+  }
+
+  function prepareInput(input, addressToAbsPathMapper, txDataHex) {
+    const data = {
```
refi93: I would return the object right away, without assigning it to a variable

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+      txDataHex,
+      outputIndex: input.outputIndex,
+      path: addressToAbsPathMapper(input.utxo.address),
+    }
+
+    return data
+  }
+
+  function prepareOutput(output, addressToAbsPathMapper) {
+    const data = {
```
refi93: "data" is too generic, I would call it "result" to make clear we are going to return it

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+      const data = await prepareOutput(output, addressToAbsPathMapper)
+      outputs.push(data)
+    }
+
+    try {
+      const response = await ada.signTransaction(inputs, outputs)
+
+      if (response.txHashHex !== unsignedTx.getId()) {
+        throw new Error('Ledger: Transaction was corrupted')
+      } else {
```
refi93: no need for "else" if you throw in the first branch of the if

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+        const txWitnesses = await prepareWitnesses(response.witnesses)
+        return {
+          txHash: response.txHashHex,
+          txBody: prepareBody(unsignedTx, txWitnesses),
+        }
+      }
+    } catch (err) {
```
refi93: This try-catch is redundant, probably remains of debugging

app/frontend/actions.js
```diff
@@ -66,33 +66,55 @@ module.exports = ({setState, getState}) => {
...
-        })
-        break
-      default:
-        return setState({
-          loading: false,
-          walletLoadingError: {
-            code: 'UnknownCryptoProvider',
-            params: {cryptoProvider},
+    try {
+      switch (cryptoProvider) {
```
ppershing: this switch should be probably extracted into some async CryptoProviderFactory

app/frontend/components/common/addressDetailDialog.js
```diff
@@ -13,7 +13,19 @@ class AddressDetailDialogClass extends Component {
...
+    showDetail,
+    showVerification,
+    error,
+    verifyAddress,
+    closeAddressDetail,
+    usingTrezor,
+    usingLedger,
+  }) {
+    let device = null
+    if (usingTrezor) device = 'Trezor'
```
ppershing: I guess device should be part of crypto provider

app/frontend/wallet/cardano-ledger-crypto-provider.js
ppershing: Maybe even Ledger would do

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+  if (state.derivationScheme.type !== 'v2') {
+    throw new Error(`Unsupported derivation scheme: ${state.derivationScheme.type}`)
+  }
+
+  const deriveXpub = CachedDeriveXpubFactory(state.derivationScheme, async (absDerivationPath) => {
+    try {
+      const response = await ada.getExtendedPublicKey(absDerivationPath)
+      return Buffer.from(response.publicKeyHex.concat(response.chainCodeHex), 'hex')
+    } catch (err) {
+      throw new Error('public key retrieval from Ledger failed')
```
ppershing: Be consistent about error messages, other start capitalized ...

app/frontend/wallet/cardano-ledger-crypto-provider.js
ppershing: See comment on line 90

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,128 @@
...
+    for (const output of unsignedTx.outputs) {
+      const data = await prepareOutput(output, addressToAbsPathMapper)
+      outputs.push(data)
+    }
+
+    try {
+      const response = await ada.signTransaction(inputs, outputs)
+
+      if (response.txHashHex !== unsignedTx.getId()) {
+        throw new Error('Ledger: Transaction was corrupted')
```
ppershing: I would prefer better error. Something like "Tx serialization mismatch between Ledger and Adalite"

server/index.js
```diff
@@ -69,7 +76,18 @@ app.get('*', (req, res) => {
...
+********************************************************************************/
+if (process.env.ADALITE_ENABLE_HTTPS === 'true') {
+  https.createServer(options, app).listen(process.env.PORT, () => {
+    // eslint-disable-next-line no-console
+    console.log(`Cardano wallet app listening on secure port ${process.env.PORT}!`)
+  })
+} else {
+  app.listen(process.env.PORT, () => {
+    // eslint-disable-next-line no-console
+    console.log(`Cardano wallet app listening on ${process.env.PORT}!`)
```
ppershing: Iguess we should change the text to "Server is lisening on port" ... as "cardano wallet app" is a bit misleading name

server/index.js
```diff
@@ -69,7 +76,18 @@ app.get('*', (req, res) => {
     `)
 })
 
-app.listen(process.env.PORT, () => {
-  // eslint-disable-next-line no-console
-  console.log(`Cardano wallet app listening on ${process.env.PORT}!`)
-})
+/*******************************************************************************
```
refi93: the convention for multiline comments here is rather:
```
/*
* some comment
* xxxx
*/
```

app/frontend/wallet/cardano-ledger-crypto-provider.js
bynjamin: It is the same as in trezor-crypto-provider. Should i change also these?

app/frontend/wallet/cardano-ledger-crypto-provider.js
bynjamin: It is the same as in trezor-crypto-provider. Should i change also these?

app/frontend/wallet/cardano-ledger-crypto-provider.js
bynjamin: I don't see any comment

bynjamin:
> Tried verifying address and the dialog disappears from the screen of the computer roughly after a minute, probably due to some timeout on Ledgerjs side, which is imho too soon. There should be a way to increase that timeout (let's say 15 minutes) or turn it off completely. Same for transaction signing.


This issue is happening in u2f-api node module. The timeout can be for example modified in TransportU2F.js (row:54 - class from ledgerjs) where u2f-api sign function is called. This is not an AdaLite issue but ledgerjs issue.

app/frontend/actions.js
```diff
@@ -66,33 +65,30 @@ module.exports = ({setState, getState}) => {
...
-          walletLoadingError: {
-            code: 'UnknownCryptoProvider',
-            params: {cryptoProvider},
-          },
-        })
+    try {
+      wallet = await CryptoProviderFactory.getWallet(cryptoProvider, secret)
+    } catch (e) {
+      debugLog(e)
+      const code = e.name === 'UnknownCryptoProviderError' ? 'UnknownCryptoProviderError' : 'WalletInitializationError'
```
ppershing: Shouldn't this be set in following switch as well? Seems a bit ad-hoc to have this logic here.

app/frontend/actions.js
```diff
@@ -103,7 +99,8 @@ module.exports = ({setState, getState}) => {
       const sendAmount = {fieldValue: ''}
       const sendAddress = {fieldValue: ''}
       const sendResponse = ''
-      const usingTrezor = cryptoProvider === 'trezor'
+      const usingHwWallet = wallet.isHwWallet()
+      const hwWalletName = wallet.getHwWalletName()
```
ppershing: shouldn't this be called only if it is hw wallet?

app/frontend/actions.js
```diff
@@ -200,20 +198,30 @@ module.exports = ({setState, getState}) => {
...
   const verifyAddress = async (address) => {
     const state = getState()
-    if (state.usingTrezor && state.showAddressDetail) {
+    if (state.usingHwWallet && state.showAddressDetail) {
       try {
+        setState({waitingForHwWallet: true})
         await wallet.verifyAddress(state.showAddressDetail.address)
-        setState({showAddressVerification: false})
+        setState({
+          showAddressVerification: false,
```
ppershing: I wonder why do we manipulate different parts of the state
1) shouldnt' first setState alse reset `showAddressVerification` and `showAddressDetail` ?
2) why do we actually have these two parts of the state? Are they different?
3) we might prefer `finally` to clean up the state, what do you thing @refi93 ?

server/index.js
```diff
@@ -69,7 +76,18 @@ app.get('*', (req, res) => {
...
 
-app.listen(process.env.PORT, () => {
-  // eslint-disable-next-line no-console
-  console.log(`Cardano wallet app listening on ${process.env.PORT}!`)
-})
+/*
+* To run server in secure mode, you need to set ADALITE_ENABLE_HTTPS to true and
+* ADALITE_SERVER_URL to 'https://localhost:3000'
+*/
+if (process.env.ADALITE_ENABLE_HTTPS === 'true') {
```
ppershing: why not simply `if (process.env...) app = https.createServer(options, app)` ? This would save us logic duplication.

server/index.js
```diff
@@ -4,8 +4,15 @@ const bodyParser = require('body-parser')
 const express = require('express')
 const compression = require('compression')
 const frontendConfig = require('./helpers/loadFrontendConfig')
+const fs = require('fs')
+const https = require('https')
 const app = express()
 
+const options = {
+  cert: fs.readFileSync('server.cert'),
```
ppershing: Does this read these files even if the server is not https?

app/frontend/wallet/crypto-provider-factory.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const UnknownCryptoProviderError = (message) => {
+    const error = new Error(message)
+    error.name = 'UnknownCryptoProviderError'
+    return error
+  }
+
+  const getWallet = async (cryptoProvider, secret) => {
+    switch (cryptoProvider) {
+      case TREZOR:
```
ppershing: I would prefix these constants. Either by string or wrap them enum-style, e.g.
`case CRYPTO_PROVIDER.TREZOR:`

app/frontend/wallet/crypto-provider-factory.js
```diff
@@ -0,0 +1,52 @@
...
+const CryptoProviderFactory = (() => {
+
+  const UnknownCryptoProviderError = (message) => {
+    const error = new Error(message)
+    error.name = 'UnknownCryptoProviderError'
+    return error
+  }
+
+  const getWallet = async (cryptoProvider, secret) => {
+    switch (cryptoProvider) {
```
ppershing: Note, we usually try to avoid switch. One idea might be

const configFactory = {
  [TREZOR]: () => {....},
  [LEDGER]: ...
}
const config = configFactory[provider]()
return await Cardano.CardanoWallet(config)
```
Looking at this, the whole function can be renamed to `walletConfigFactory` and would return only config, caller would do Cardano.CardanoWallet ...

File: app/frontend/wallet/cardano-ledger-crypto-provider.js [hunk 25]
```diff
@@ -0,0 +1,124 @@
...
+    throw new Error(`Unsupported derivation scheme: ${state.derivationScheme.type}`)
+  }
+
+  const isHwWallet = () => true
+  const getHwWalletName = () => 'Ledger'
+
+  const deriveXpub = CachedDeriveXpubFactory(state.derivationScheme, async (absDerivationPath) => {
+    try {
+      const response = await ledger.getExtendedPublicKey(absDerivationPath)
+      return Buffer.from(response.publicKeyHex.concat(response.chainCodeHex), 'hex')
```
ppershing: Personally, I am not big fan of "concat" call as it looks asymmetric.
Why not `const xpubHex = response.publicKeyHex + response.chainCodeHex` and then `Buffer.from(xpubHex)` ?

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,124 @@
...
+const CachedDeriveXpubFactory = require('./helpers/CachedDeriveXpubFactory')
+const cbor = require('borc')
+const {TxWitness, SignedTransactionStructured} = require('./transaction')
+
+const CardanoLedgerCryptoProvider = async (ADALITE_CONFIG, walletState) => {
+  const transport = await TransportU2F.create()
+  transport.setExchangeTimeout(15 * 60000) // 60000ms = 1minute
+  const ledger = new Ledger(transport)
+
+  const state = Object.assign(walletState, {
```
ppershing: is this `Object.assign` intentional? Shouldn't this be using `setState` ?

app/frontend/wallet/cardano-ledger-crypto-provider.js
```diff
@@ -0,0 +1,124 @@
+const TransportU2F = require('@ledgerhq/hw-transport-u2f').default // for browser
+const Ledger = require('ledgerhq/hw-app-ada').default // temporary hack, should be @ledgerhq
+const CachedDeriveXpubFactory = require('./helpers/CachedDeriveXpubFactory')
+const cbor = require('borc')
+const {TxWitness, SignedTransactionStructured} = require('./transaction')
+
+const CardanoLedgerCryptoProvider = async (ADALITE_CONFIG, walletState) => {
+  const transport = await TransportU2F.create()
+  transport.setExchangeTimeout(15 * 60000) // 60000ms = 1minute
```
ppershing: move to constants?

app/frontend/components/pages/login/mnemonicAuth.js
```diff
@@ -75,7 +76,7 @@ class LoadByMenmonicSectionClass extends Component {
                 mnemonic && !mnemonicValidationError ? 'pulse' : ''
               }`,
               disabled: !mnemonic || mnemonicValidationError,
-              onClick: () => loadWallet({cryptoProvider: 'mnemonic', secret: mnemonic}),
+              onClick: () => loadWallet({cryptoProvider: MNEMONIC, secret: mnemonic}),
```
ppershing: I am not sure `MNEMONIC` is a good name of `cryptoProvider` as mnemonic does not provide crypto. @refi93 any suggestions on naming?

refi93: to keep it consistent with module names, I would name it WALLET_SECRET

app/frontend/wallet/cardano-ledger-crypto-provider.js
refi93: I think this is Ledger specific, but perhaps the "LOGOUT_AFTER" constant could be reused, which makes sense to me.

app/frontend/components/pages/login/mnemonicAuth.js
refi93: and I agree with namespacing those constants as you pointed out below in crypto-provider-factory.js

app/frontend/actions.js
bynjamin: I just wanted to maintain simplicity of code, avoid redundant ifs and uniformity of all crypto providers. For mnemonic crypto provider the function returns undefined. But you're right. It's a bit tricky.

app/frontend/actions.js
bynjamin: **showAddressDetail** is flag to show parent modal
**showAddressVerification** is flag to show child div. If we are using hw wallet and when it's true, address and pink button with "verify on <hw-wallet>" is displayed in modal. If it's false, QR code with address and path is displayed.
**waitingForHwWallet** is defined for click feedback. When it's true, the spinner is shown on button.

bynjamin: Maybe it's misleading that this state.usingHwWallet && state.showAddressDetail don't include also this `&& state.showAddressVerification`

app/frontend/wallet/crypto-provider-factory.js
bynjamin: With this approach, we lose ability to send secret parameter, when the crypto provider is mnemonic, don't we?

app/frontend/wallet/cardano-ledger-crypto-provider.js
bynjamin: It was inspired by trezor crypto provider, but i don't understand why it should by using setState. It seems like constant read-only state used only by this class.

refi93: at cardano wallet level, there is no "setState" function for changing the state, this is used only on the preact layer. This state isn't the global app state, it's just the state of the cardano wallet provider.

app/frontend/wallet/crypto-provider-factory.js
refi93: not really, functions have access to the outer scope in js

app/frontend/wallet/cardano-ledger-crypto-provider.js
ppershing: Then why dont' we manipulate walletState directly?

refi93: Yes, we can, it would be more straightforward. I'm not sure at this moment why we used Object.assign, it was probably a half-hearted effort to make a "local" state, but it's probably quite misleading, since this "local" state has also access to the global one.

refi93: But probably I would not go for refactoring it now, and keep the pattern for the sake of consistency - I created a separate issue for that: https://github.com/vacuumlabs/adalite/issues/383

PR #382, opened 2019-02-13
Yoroi compatibility

bynjamin:


app/frontend/actions.js
```diff
@@ -82,7 +82,7 @@ module.exports = ({setState, getState}) => {
           mnemonicOrHdNodeString: secret,
           config: ADALITE_CONFIG,
           network: 'mainnet',
-          derivationScheme: derivationSchemes.v1,
+          //derivationScheme: derivationSchemes.v1,
```
refi93: remove this line please

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
 
-  function discoverAddresses() {
+  async function discoverAddresses() {
+    switch (derivationScheme.type) {
+      case 'v1':
+        return await discoverAddressesV1()
+      case 'v2':
+        return await discoverAddressesV2()
+      default:
+        return null
```
refi93: this is dangerous, we should throw an exception if an unknown scheme is supplied

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+      if (!isGapBlock || blockItterator === 0) addresses = [...addresses, ...block]
+
+      childIndexBegin += blockSize
+      blockItterator++
+    }
+    return await reduceAddresses(addresses)
+  }
+
+  /*
+   * Function which discovers addresses from begin index to end index
```
refi93: this comment is redundant, and I would rather name the function to "getAbsDerivationPaths(...)" and return only the absolute derivation paths, since that's the trickiest part which is worth encapsulating.

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+   * To achieve compatibility with Yoroi wallet, we have to discover all addresses
+   * until gap limit is reached. We discover addresses by blocks of 20. This should
+   * be more foolproof (v2 scheme).
+   */
+  async function discoverAddressesV2() {
+    const blockSize = 20
+    let addresses = []
+    let isGapBlock = false
+    let childIndexBegin = derivationScheme.startAddressIndex
+    let blockItterator = 0
```
refi93: typo: Itterator -> Iterator

refi93: and I would name it rather addressBlockIndex

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+    let childIndexEnd, block
```
refi93: I would rename block to currentAddressBlock

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+
+    while (!isGapBlock) {
```
refi93: I would rewrite this with a do { } while cycle, so you don't need declaring "isGapBlock" outside. Also childIndexEnd makes more sense to declare it inside the cycle

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+    return discoverAddressesBlock(childIndexBegin, childIndexEnd)
+  }
+
+  /*
+   * To achieve compatibility with Yoroi wallet, we have to discover all addresses
+   * until gap limit is reached. We discover addresses by blocks of 20. This should
+   * be more foolproof (v2 scheme).
+   */
+  async function discoverAddressesV2() {
+    const blockSize = 20
```
refi93: I would name this "gapLimit" which is a standard name for that

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
     const absDerivationPaths = range(childIndexBegin, childIndexEnd)
       .map((i) => [accountIndex, isChange ? 1 : 0, i])
       .map(derivationScheme.toAbsoluteDerivationPath)
 
     return deriveAddresses(absDerivationPaths)
   }
 
+  /*
+   * To maintain default behavior of AdaLite, we reduce addreses
+   * to default 10 if it's possible
```
refi93: "if it's possible" - I would elaborate a bit more on that - perhaps saying "if no address beyond the first 10 are used"

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+   */
+  async function reduceAddresses(addresses) {
+    if (addresses.length <= 20) {
+      const isUnused = !(await blockchainExplorer.isSomeAddressUsed(addresses.slice(10)))
+      if (isUnused) return addresses.slice(0, 10)
```
refi93: I would replace 10 with addressLimit variable

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
+
+    while (!isGapBlock) {
+      childIndexEnd = childIndexBegin + blockSize
+      block = await discoverAddressesBlock(childIndexBegin, childIndexEnd)
+
+      isGapBlock = !(await blockchainExplorer.isSomeAddressUsed(block))
+      if (!isGapBlock || blockItterator === 0) addresses = [...addresses, ...block]
+
+      childIndexBegin += blockSize
+      blockItterator++
```
ppershing: typo

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
     return deriveAddresses(absDerivationPaths)
   }
 
+  /*
+   * To maintain default behavior of AdaLite, we reduce addreses
+   * to default 10 if it's possible
+   */
+  async function reduceAddresses(addresses) {
+    if (addresses.length <= 20) {
+      const isUnused = !(await blockchainExplorer.isSomeAddressUsed(addresses.slice(10)))
```
refi93: I would replace 10 with the addressLimit variable

app/frontend/wallet/address-manager.js
```diff
@@ -9,21 +9,81 @@ const AddressManager = ({
...
 
     return deriveAddresses(absDerivationPaths)
   }
 
+  /*
+   * To maintain default behavior of AdaLite, we reduce addreses
+   * to default 10 if it's possible
+   */
+  async function reduceAddresses(addresses) {
+    if (addresses.length <= 20) {
```
refi93: this condition is redundant

app/frontend/wallet/cardano-wallet.js
```diff
@@ -285,7 +287,7 @@ const CardanoWallet = async (options) => {
 
   async function getChangeAddress() {
     // if we used all available addresses return random address from the available ones
```
refi93: remove please this old comment, it's misleading since now we always return a random address

app/frontend/wallet/address-manager.js
bynjamin: thanks :smile:

app/frontend/wallet/address-manager.js
bynjamin: Okay, but isn't it little bit confusing?

PR #384, opened 2019-02-15
Bulk request optimalization

bynjamin:
DEPENDS ON yoroi-compatibility BRANCH!!!


app/frontend/helpers/hashCode.js
```diff
@@ -0,0 +1,19 @@
+const hashCode = (s) => {
```
refi93: please cite source if you found it somewhere. Is it some standard hash algorithm? If so, please call the function to reflect that.

refi93:
you probably forgot to delete  `app/tests/src/common/singleAddressesMock.js`

app/tests/src/common/bulkSummaryRequests.js
```diff
@@ -0,0 +1,366 @@
...
+    'Ae2tdPwUPEZLYjtsKf6xjC5TCmD3RKS5BXMY3bhPv9P8X7QATo3dLzVLU2m',
+    'Ae2tdPwUPEZEwDwRoBzJt3WwhpdEZXzQZTMx7xEjUGRqN2TziD7dj87bTnz',
+    'Ae2tdPwUPEZL7TdN2Ugr3isi3m5UoH6nXnMZA63VuYBBsbTiqALv3UMM2bS',
+    'Ae2tdPwUPEZ1j7ShGjWmWJNmtN9neEYgMcixuguum8cidaRnGERGh8QMDGb',
+    'Ae2tdPwUPEZHszSLNV6L8qAFE1Jh7Q8u5z3h4NU3NRReAn4e47zhDgJTnRC',
+    'Ae2tdPwUPEZN3GEFq8FoMzsSchMRwjcwzUoHsy3ARHUHQQHZxYzeDXtAkir',
+    'Ae2tdPwUPEZD9a9iktQ4YRVU8tSUE5KgUWJAWcVhNrh8G3xEXbd2ZJBH7Jx',
+  ],
+  [
+    'Ae2tdPwUPEZ6RUCnjGHFqi59k5WZLiv3HoCCNGCW8SYc5H9srdTzn1bec4W',
```
refi93: this seems weird - why is there a request for so many addresses? I would expect that each of the summary requests happens in a bulk of `GAP_LIMIT` addresses, but this is far more than 20

app/frontend/helpers/hashCode.js
bynjamin: https://gist.github.com/iperelivskiy/4110988

PR #394, opened 2019-02-27
Design 2.0

refi93:


refi93:
Issues found:
[] "raw unsigned transaction" link is not disabled when the address/amount/fee are missing
[] "max" button is clickable when the address is missing, but max amount cannot be computed without knowing the output address (which influences the final transaction size and therefore the fee)
[] when I click on an address from "My Addresses" that takes more than one line, only the line I clicked on is selected
[] When verifying an address on trezor, the "Verify on trezor" link on the modal should be disabled, ideally being replaced by a text "Waiting...", until the user actually confirms the address. Clicking again on the "Verify on Trezor" link results in closing the modal which is not desired.
[] When I confirm the transaction, the screen gets stuck on the transaction summary modal until the tranaction is submitted. The expected behaviour after confirming the transaction is seeing a spinner describing the state of the transaction being submitted.

d-novak:
Updated the checklist - the checked off ones were resolved in PR #406

refi93:
closing this one to store the original state of redesign before rebase to latest develop

PR #400, opened 2019-03-01
Export import wallet

bynjamin:


app/frontend/wallet/helpers/mnemonicToWalletSecret.js
```diff
@@ -0,0 +1,33 @@
...
+
+const _mnemonicToWalletSecret = async (mnemonic, derivationScheme) => {
+  if (await isMnemonicInPaperWalletFormat(mnemonic)) {
+    mnemonic = await decodePaperWalletMnemonic(mnemonic)
+  }
+
+  return mnemonicToRootKeypair(mnemonic, derivationScheme.number)
+}
+
+const mnemonicToWalletSecret = async (mnemonic) => {
```
refi93: I would name it `mnemonicToWalletSecretDef` to keep consistent naming with the output structure which is called `walletSecretDef` in the code

app/frontend/wallet/keypass-json.js
```diff
@@ -102,23 +110,29 @@ async function isWalletExportEncrypted(walletExportObj) {
 }
 
 async function importWalletSecret(walletExportObj, password) {
-  const {passwordHash, walletSecret} = parseWalletExportObj(walletExportObj)
+  const {passwordHash, walletSecretDef} = parseWalletExportObj(walletExportObj)
```
refi93: also I would call this function `importWalletSecretDef` for the sake of consistency

app/frontend/wallet/keypass-json.js
```diff
@@ -102,23 +110,29 @@ async function isWalletExportEncrypted(walletExportObj) {
...
   }
 
-  return decryptWalletSecret(walletSecret, password)
+  walletSecretDef.rootSecret = decryptWalletSecret(walletSecretDef.rootSecret, password)
+
+  return walletSecretDef
 }
 
-async function exportWalletSecret(walletSecret, password, walletName) {
+async function exportWalletSecret(walletSecret, password, walletName, derivationScheme) {
```
refi93: I would refactor this function to take a `walletSecretDef` object and rename it `exportWalletSecretDef` for the sake of consistency

app/frontend/wallet/cardano-wallet.js
```diff
@@ -26,15 +25,14 @@ function txFeeFunction(txSizeInBytes) {
...
-
+  const {walletSecretDef, config, randomSeed, network} = options
   const state = {
     randomSeed: randomSeed || Math.floor(Math.random() * MAX_INT32),
     ownUtxos: {},
     overallTxCountSinceLastUtxoFetch: 0,
     accountIndex: HARDENED_THRESHOLD,
     network,
-    derivationScheme: options.derivationScheme || derivationSchemes.v2,
+    derivationScheme: null, //is set in cryptoproviders
```
refi93: I would not set any derivationscheme there. I would maintain the pattern that derivation scheme should be managed solely by crypto providers. So let's say the address manager below should not receive the derivation scheme as a parameter, but rather ask the cryptoprovider about the derivation scheme. This will avoid mistakenly setting one derivation scheme in one module and another scheme in other module.

app/frontend/actions.js
```diff
@@ -103,7 +100,10 @@ module.exports = ({setState, getState}) => {
       const sendAddress = {fieldValue: ''}
       const sendResponse = ''
       const usingTrezor = cryptoProvider === 'trezor'
-      const isDemoWallet = secret === ADALITE_CONFIG.ADALITE_DEMO_WALLET_MNEMONIC
+      const isDemoWallet =
+        walletSecretDef && // we can also use !usingTrezor
```
refi93: I don't understand the "we can also use !usingTrezor" comment - either remove it or  provide a more elaborate explanation

app/frontend/wallet/cardano-trezor-crypto-provider.js
```diff
@@ -122,8 +120,13 @@ const CardanoTrezorCryptoProvider = (ADALITE_CONFIG, walletState) => {
     throw new Error('Unsupported operation!')
   }
 
+  function getWalletDerivationScheme() {
```
refi93: I would just call it `getDerivationScheme` - the crypto provider does not have a notion of "wallet"

refi93:
I would also add a test of exporting/importing a v2 mnemonic

app/frontend/actions.js
bynjamin: When we merge it with ledger integration, there will be usingHwWallet boolean. In my opinion, using this boolean would be more descriptive than checking just if walletSecretDef is present.

PR #405, opened 2019-03-06
Shown addresses reduction

bynjamin:


app/frontend/wallet/constants.js
```diff
@@ -19,4 +19,5 @@ module.exports = {
     },
   },
   DELAY_AFTER_TOO_MANY_REQUESTS: 2000,
+  DEFAULT_ADDRESS_COUNT: 10,
```
refi93: isn't this fetched from .env? I think this should not be needed anymore and the value should be fetched from the environment

PR #409, opened 2019-03-08
[WIP] check-types added

refi93:


server/helpers/loadConfig.js
```diff
@@ -0,0 +1,86 @@
+// config intended to be passed on to frontend
+// keys in process .env must be referencesd explicitly
+// so dotenv-webpack is able to include them in the test bundle for Mocha tests
+const check = require('check-types')
+
+// process.env.npm_package_version is undefined on Heroku
+const appVersion = require('../../package.json').version
+
+const {env} = process
+const booleans = ['true', 'false']
```
refi93: I would directly declare a helper function "isBoolStr" and use that instead of repeatedly calling includes

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,86 @@
...
+
+check.assert(
```
refi93: I would rather use check.map

```
check.map({
    foo: 2,
    bar: { baz: 'qux' }
}, {
    foo: check.odd,
    bar: { baz: check.nonEmptyString }
});
```
Then check with `check.all` if everything is true and if not, I would just print the result of check.map from which it should be obvious which value was invalid without the need for such eloquent error messages.

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,98 @@
...
+  ADALITE_MOCK_TX_SUBMISSION_SUCCESS,
+  ADALITE_MOCK_TX_SUMMARY_SUCCESS,
+  TREZOR_CONNECT_URL,
+  REDIS_URL,
+  STATS_PWD,
+  PORT,
+} = process.env
+
+const checkMap = check.map(
+  {
```
refi93: I would put there directly process.env and where integers need to be validated I would do something like this:
```
   ADALITE_DEFAULT_ADDRESS_COUNT: (x) => check.positive(parseInt(x, 10))
```
This should vastly eliminate the duplicity

PR #410, opened 2019-03-11
Relocate Create new wallet functionality

### LINKED ISSUE #402: Relocate Create new Wallet functionality

d-novak:
Closes #402 

Relocated the `Create New Wallet` functionality from the navigation to the Mnemonic login page.


refi93:
please remove `isLeftClick` require from `navbarUnauth.js` since it's unused now

d-novak:
@refi93 Sorry missed that, removed

refi93:
after discussing it with Michal we came to the conclusion that it would be better to have both the "create new wallet" button on the navbar and in the mnemonic tab. The reason is that without the button the landing page would lack any call to action.

d-novak:
@refi93 Added it back

PR #414, opened 2019-03-13
Empty config detection (check-types)

### LINKED ISSUE #327: Avoid running server with empty config
It causes weird behaviour (redirect loops since base url is not set, ...)

It would be nice to detect a missing config (e.g. non-presence of the  and kill the app with a comprehensive message. The options I see are either checking for presence of certain variables (a bit hacky approach), or to load the environment variables to a js object (see e.g. https://www.npmjs.com/package/config) and detect such required variables (failing if they are not available). Feel free to research best practices.

bynjamin:
closes #327


server/helpers/loadConfig.js
```diff
@@ -0,0 +1,98 @@
...
+  ADALITE_MOCK_TX_SUBMISSION_SUCCESS,
+  ADALITE_MOCK_TX_SUMMARY_SUCCESS,
+  TREZOR_CONNECT_URL,
+  REDIS_URL,
+  STATS_PWD,
+  PORT,
+} = process.env
+
+const checkMap = check.map(
+  {
```
refi93: I would put there directly process.env and where integers need to be validated I would do something like this:
```
   ADALITE_DEFAULT_ADDRESS_COUNT: (x) => check.positive(parseInt(x, 10))
```
This should vastly eliminate the duplicity

server/helpers/isBoolStr.js
```diff
@@ -0,0 +1,5 @@
+const boolStrings = ['true', 'false']
```
refi93: I would not put the helper in a separate file since it is used only in one place and has quite a specific usecase

package.json
```diff
@@ -41,6 +41,7 @@
   },
   "dependencies": {
     "body-parser": "^1.18.2",
+    "check-types": "^8.0.1",
```
refi93: until the fix is launched please revert to version 7.4.0

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,85 @@
+// config intended to be passed on to frontend
+// keys in process .env must be referencesd explicitly
```
refi93: typo: referencesd

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,85 @@
...
+  REDIS_URL,
+  STATS_PWD,
+  PORT,
+} = process.env
+
+const checkMap = check.map(process.env, {
+  ADALITE_ENABLE_DEBUGGING: isBoolString,
+  ADALITE_SERVER_URL: check.nonEmptyString,
+  ADALITE_BLOCKCHAIN_EXPLORER_URL: check.nonEmptyString,
+  ADALITE_DEFAULT_ADDRESS_COUNT: (x) => check.positive(parseInt(x, 10)),
```
refi93: I would extract that check to isPositiveIntString similarly as isBoolString

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,85 @@
...
+  ADALITE_GAP_LIMIT: (x) => check.positive(parseInt(x, 10)),
+  ADALITE_DEMO_WALLET_MNEMONIC: check.nonEmptyString,
+  ADALITE_ENABLE_TREZOR: isBoolString,
+  ADALITE_LOGOUT_AFTER: (x) => check.positive(parseInt(x, 10)),
+  ADALITE_ENABLE_SERVER_MOCKING_MODE: isBoolString,
+  ADALITE_MOCK_TX_SUBMISSION_SUCCESS: isBoolString,
+  ADALITE_MOCK_TX_SUMMARY_SUCCESS: isBoolString,
+})
```
refi93: I would validate PORT variable as well, since there is no default value for it

server/helpers/loadConfig.js
```diff
@@ -0,0 +1,85 @@
...
+  TREZOR_CONNECT_URL,
+  REDIS_URL,
+  STATS_PWD,
+  PORT,
+} = process.env
+
+const checkMap = check.map(process.env, {
+  ADALITE_ENABLE_DEBUGGING: isBoolString,
+  ADALITE_SERVER_URL: check.nonEmptyString,
+  ADALITE_BLOCKCHAIN_EXPLORER_URL: check.nonEmptyString,
```
refi93: I would check whether it is an url, there is a validator for it in check-types AFAIK

bynjamin: webUrl predicate was removed in check-types version 3

PR #415, opened 2019-03-13
address numbers added

### LINKED ISSUE #413: Add address number next to address in My Addresses

bynjamin:
closes #413

### DIFF
--- a/app/frontend/components/pages/receiveAda/addressItem.js
+++ b/app/frontend/components/pages/receiveAda/addressItem.js
@@ -11,7 +11,12 @@ const AddressItem = connect(
   h(
     'div',
     {class: 'address'},
-    h('p', {class: 'address-value one-click-select'}, address),
+    h(
+      'div',
+      {class: 'address-value'},
+      h('p', {class: 'address-number no-select'}, `/${bip32path.split('/').pop()}`),
+      h('p', {class: 'one-click-select'}, address)
+    ),
     h(
       'div',
       {class: 'address-links'},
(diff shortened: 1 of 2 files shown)


app/public/css/styles.css
```diff
@@ -1690,6 +1690,18 @@ button[data-balloon] {
...
   margin-bottom: 16px;
+  display: flex;
+}
+
+.address-number {
+  font-weight: bold;
+  word-break: normal;
+  margin-right: 15px;
+  user-select: none;
+  -moz-user-select: none;
```
refi93: I would rather make a separate class `unselectable` for that since it's a generic property other elements may have and the classname would make clear in the html that the element is indeed unselectable

app/public/css/styles.css
```diff
@@ -1690,6 +1690,21 @@ button[data-balloon] {
   color: var(--color-grey);
   word-break: break-all;
   margin-bottom: 16px;
+  display: flex;
+}
+
+.address-number {
+  font-weight: bold;
+  word-break: normal;
+  margin-right: 15px;
```
d-novak: Please change to `margin-right: 16px`, we're using increments of 8 for spacing.

app/frontend/components/pages/receiveAda/addressItem.js
```diff
@@ -11,7 +11,12 @@ const AddressItem = connect(
   h(
     'div',
     {class: 'address'},
-    h('p', {class: 'address-value one-click-select'}, address),
+    h(
+      'div',
+      {class: 'address-value'},
+      h('p', {class: 'address-number unselectable'}, `/${bip32path.split('/').pop()}`),
```
d-novak: I'd prefer a class `no-select`, it's easily readable.

PR #417, opened 2019-03-13
generated mnemonic autocopy disabled & unlock button glitch fixed

### LINKED ISSUE #416: Disable autocopying mnemonic when generating a new one

bynjamin:
closes #416

### DIFF
--- a/app/frontend/components/pages/login/generateMnemonicDialog.js
+++ b/app/frontend/components/pages/login/generateMnemonicDialog.js
@@ -4,7 +4,7 @@ const actions = require('../../../actions')
 const Modal = require('../../common/modal')
 
 class GenerateMnemonicDialogClass {
-  render({confirmGenerateMnemonicDialog, mnemonic, closeGenerateMnemonicDialog}) {
+  render({confirmGenerateMnemonicDialog, newWalletMnemonic, closeGenerateMnemonicDialog}) {
     return h(
       Modal,
       {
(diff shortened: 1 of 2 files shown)


app/frontend/components/pages/login/generateMnemonicDialog.js
```diff
@@ -18,7 +19,7 @@ class GenerateMnemonicDialogClass {
         'The new wallet is created together with a mnemonic phrase. Write the mnemonic phrase down, you will need it to access your wallet. ',
         h('strong', undefined, 'Don’t copy it to your clipboard or save it anywhere online.')
       ),
-      h('div', {class: 'modal-mnemonic one-click-select'}, mnemonic),
+      h('div', {class: 'modal-mnemonic one-click-select'}, generateMnemonic()),
```
refi93: this isn't a good idea since re-rendering the component would trigger mnemonic regeneration. I would rather keep the previous logic but wipe the generated mnemonic from the state once the user closes the generate mnemonic dialog.

bynjamin: Ok , but it has to be different state parameter to avoid mnemonic field to be filled out on background and than cleared when user closes the window. It looks glitchy. For example generatedMnemonic.

refi93:
When I fill some random characters in the mnemonic input and then I click "create new wallet" and the modal shows, the "Unlock" button becomes active (not greyed out) and it stays like that even after exiting the generate mnemonic modal. It may not be directly related to this PR but I would rather have it fixed as well.

PR #420, opened 2019-03-14
Tidy up frontend dependencies

### LINKED ISSUE #418: Tidy up frontend dependencies
- webpack should be dev dependency

- file-saver and unistore are quite small libs - I'd copy them to the project as they are (only the relevant parts of them)

- we're having multiple versions of babel-polyfill and babel-runtime installed (because some of our libs depend on it by themselves). I'd download the polyfill from CDN and configure it as external (thus forcing the same version for everyone). (see current webpack config for example of external dependencies).

Alternatively, you can keep using npmregistry and just force the one version such as:
https://stackoverflow.com/questions/40884485/force-dependency-to-use-specific-sub-dependency-version.

refi93:
- [X] moved webpack to dev dependencies
- [X] moved unistore and file-saver to the repo as hard-linked libraries
- [X] enforced version of babel-polyfill and babel-runtime (@babel/runtime) through [resolutions](https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-flat) and enforced the use of @babel-runtime over babel-runtime in webpack config by aliasing it (ledgerjs uses babel-runtime and trezor-connect @babel/runtime). npm shrikwrap seems not to be compatible with yarn

Along the way I aliased in webpack the hardlinked libraries for better developer experience and so it would not be needed to refactor the imports.

closes #418


app/frontend/libs/file-saver.js
```diff
@@ -0,0 +1,218 @@
+/* eslint-disable */
```
tomaskulich: Cannot we use not-transpiled version of the dependency, so that the result is more readable? 

https://github.com/eligrey/FileSaver.js/blob/master/src/FileSaver.js

app/frontend/libs/unistore/index.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable */
```
tomaskulich: Same here, why not go for readable version?

PR #432, opened 2019-03-15
home link added to export page in mobile view

bynjamin:


refi93:
I thought about it a bit more and I would rather remove the export button from navbar totally in favor of the "about" link which can be on the navbar both in mobile and desktop mode. i.e. solve this issue: https://github.com/vacuumlabs/adalite/issues/433 and close this one

PR #440, opened 2019-03-22
verify address error handle

### LINKED ISSUE #436: Handle error when verifying address
When hw wallet fails for some reason (e.g. ledger gets locked after 15 mins), the modal to verify the address just hides without any error message. An error popup describing the error should show.

bynjamin:
closes #436


bynjamin:
Screenshot from 2019-03-22 12-48-45

PR #445, opened 2019-04-01
Add Android + ledger option

tomaskulich:


app/frontend/components/pages/login/hardwareAuth.js
```diff
@@ -47,6 +47,7 @@ const LoadByHardwareWalletSection = ({loadWallet}) => {
         alt: 'Ledger Nano S',
       }),
       h('div', {class: 'authentication-paragraph'}, 'Ledger Nano S'),
+      h('div', {class: 'authentication-paragraph small'}, 'also with Android device'),
```
j-matus: This causes the Unlock with Trezor / Unlock with Ledger buttons to be no longer aligned, are we OK with that?

tomaskulich: Probably not. Please think about it and come with some solution.

j-matus: Adding a small authentication paragraph above the Unlock with Trezor button and setting its visibility to hidden is not very elegant but seems to work. Another option is to position the Trezor button by changing the `.button.trezor` style.

tomaskulich: I fixed it somehow. It'll do IMO.

PR #446, opened 2019-04-01
Polish env

tomaskulich:


.env.example
```diff
@@ -1,18 +1,19 @@
...
-REDIS_URL = ''
-STATS_PWD =admin
-ADALITE_DEMO_WALLET_MNEMONIC = 'civil void tool perfect avocado sweet immense fluid arrow aerobic boil flash'
-ADALITE_LOGOUT_AFTER = 900
-ADALITE_DISABLE_LEDGER = true
+ADALITE_SERVER_URL='http://localhost:3000'
+ADALITE_BLOCKCHAIN_EXPLORER_URL='https://explorer2.adalite.io'
+ADALITE_DEFAULT_ADDRESS_COUNT=10
+ADALITE_GAP_LIMIT=20
+ADALITE_REDIS_URL=''
```
refi93: `REDIS_URL` is in production injected by heroku, so that would not very likely work in production. Same for `PORT` environment variable. The remaining ones should be ok. Perhaps some aliasing could be done with `PORT` and `REDIS_URL` but I don't see a nice solution for that, so I would just leave those two as-is probably.

PR #457, opened 2019-04-12
Keep transaction form filled after user rejection (from hw wallet)

### LINKED ISSUE #453: Don't clear transaction if it is rejected
E.g. when signing the tx on ledger and the user rejects it, it would be nice if the inputs did not clear

bynjamin:
closes #453

### DIFF
--- a/app/frontend/actions.js
+++ b/app/frontend/actions.js
@@ -424,15 +424,20 @@ module.exports = ({setState, getState}) => {
     }
   }
 
-  const resetSendForm = (state) => {
+  const resetSendFormState = (state) => {
     setState({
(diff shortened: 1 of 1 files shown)


app/frontend/actions.js
```diff
@@ -423,18 +423,23 @@ module.exports = ({setState, getState}) => {
     }
   }
 
-  const resetSendForm = (state) => {
+  const resetSendFormState = () => {
```
j-matus: Even though the value of `state` will not be read, it makes sense to keep `state` as argument to signify that this function mutates it and to keep the style consistent.

PR #461, opened 2019-05-21
Replace AdaScan with Seiza

j-matus:
Replace AdaScan links with Seiza links.

### DIFF
--- a/app/frontend/components/common/footer.js
+++ b/app/frontend/components/common/footer.js
@@ -103,7 +103,7 @@ const Footer = connect(
             'a',
             {
               class: 'donations-item ada',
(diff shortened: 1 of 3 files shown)


refi93:
@ppershing what do you mean by having both? Will there be two links? I'm not sure if I understand how that would look like.

refi93:
Yes, we can, at least for the address and transaction tiles. A possible approach without taking up too much space could be having it as `View on <href>Seiza</> <href>AdaScan</>`. Could you test it please @j-matus whether it works for different resolutions and if so, make it this way?

j-matus:
@refi93 It looks like we'd have to change the components a bit to support two links in the 'My Addresses' part.

PR #477, opened 2019-06-07
Temporarily add AdaScan links

j-matus:
Temporarily add AdaScan links for transaction history and addresses on
top of Seiza links.


j-matus:
@refi93 This solution is pretty ugly but gets the job done and if we remove AdaScan links again in the future we won't need to change `styles.css`.

PR #478, opened 2019-06-13
Add contact form modal

### LINKED ISSUE #473: Add a support contact form on adalite

j-matus:
Add a contact form modal which can be opened by a link in the footer.
The modal contains some hints that can make self-diagnosis easier.
The user can also send a message to the support email via Formspree.
The support email address is stored in .env. Closes #473

### DIFF
--- a/app/frontend/components/common/contactForm.js
+++ b/app/frontend/components/common/contactForm.js
@@ -1,0 +1,133 @@
+const {h, Component} = require('preact')
+const connect = require('unistore/preact').connect
+const actions = require('../../actions')
+const Modal = require('./modal')
+const {ADALITE_CONFIG} = require('../../config')
+
+class ContactForm extends Component {
+  constructor(props) {
+    super(props)
+    this.closeContactFormModal = this.closeContactFormModal.bind(this)
+  }
+
+  closeContactFormModal() {
+    this.props.closeContactFormModal()
+  }
+
+  render() {
+    return h(
+      Modal,
+      {
+        closeHandler: this.closeContactFormModal,
+      },
+      h(
+        'main',
+        {class: 'contact-modal'},
+        h('h2', {class: 'export-title'}, 'Contact Us'),
+        h(
+          'p',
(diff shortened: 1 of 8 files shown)


martinmaricak:
I sent a couple of notes in Slack to @ladomyriak regarding this:

- the copy under the heading `consider the following…` doesn’t seem right. Did @MichalPetro provide it? If so, I’ll talk to him about it after he’ll be back from vacation.
- “Your issue” copy also doesn’t seem right. If it is a `contact form`, we should go with stuff like `Your message` instead. If it is `report an issue form` we have the modal hading wrong.
- `Your issue` input placeholder uses a different font than others. We should make it the same font as in `Your email` input.
- I would move the `Back to AdaLite` button next to the `Submit` button
- Let’s replace `Back to AdaLite` button with a gray button style (just like `Logout` or `Refresh Balance` buttons, but without an icon).

@d-novak @refi93

refi93:
@martinmaricak 
* the text and checklist is just a very rough draft, they were not reviewed yet. They will probably be replaced by a single sentence and link to troubleshooting/issue reporting guideline.
* I agree, we should replace the copy as suggested. The form is meant mainly to report issues but not necessarily.
* ok, let's add back the "back to adalite" button athe way you suggested
* the email input is a HTML 5 "email" element, i.e. it's natively validated

app/frontend/components/common/contactForm.js
```diff
@@ -0,0 +1,104 @@
...
+            type: 'hidden',
+            name: '_subject',
+            value: 'Support Contact Form',
+          }),
+          h('input', {
+            type: 'text',
+            name: '_gotcha',
+            style: 'display:none',
+          }),
+          h('br', {}),
```
refi93: not sure if this is clean. Shouldn't the spacing rather be achieved with an appropriate padding?

ladomyriak: `br` removed.

PR #495, opened 2019-07-15
I 487 migrate statistics from redis to google universal analytics

### LINKED ISSUE #487: Migrate statistics from Redis to Google Universal Analytics
Currently the counters are stored in redis which seems cumbersome. A better suited tool seems to be google universal analytics where we would be increasing just one counter and the counts over days/months/total would be derived automatically. Of course, this tracking will still remain on the backend side and no IPs should be permanently stored.

xdzurman:
closes #487


package.json
```diff
@@ -53,7 +53,8 @@
     "isomorphic-fetch": "^2.2.1",
     "normalize-url": "^4.0.0",
     "redis": "^2.8.0",
-    "redisscan": "^2.0.0"
+    "redisscan": "^2.0.0",
```
refi93: this package is now redundant, please remove redis-related packages

server/middlewares/stats.js
```diff
@@ -4,8 +4,11 @@ const client = redis.createClient(process.env.REDIS_URL)
 const mung = require('express-mung')
 const normalizeUrl = require('normalize-url')
 const cbor = require('borc')
+const ua = require('universal-analytics')
 
 const knownIps = new Set()
+const adaliteTrackingID = 'UA-143401606-1'
```
refi93: this should not be hardcoded but provided from .env, see how other environment variables are provided. This property should be optional and if no "adaliteTrackingID" is defined, the middleware should not try to submit anything anywhere

server/middlewares/stats.js
```diff
@@ -57,6 +90,7 @@ const trackVisits = (req, res, next) => {
   if (!knownIps.has(ip) && !mydevice.is('bot')) {
     knownIps.add(ip)
     incrCountersBy('visits', 1)
```
refi93: this is now redundant if I understand right

server/middlewares/stats.js
```diff
@@ -78,6 +112,18 @@ const trackTxSubmissions = mung.json((body, req, res) => {
       const txOutAmount = parseTxBodyOutAmount(txBody)
 
       incrCountersBy(`${txSubmissionType}:sentOut`, txOutAmount)
```
refi93: this is now redundant if I understand right

server/middlewares/stats.js
```diff
@@ -48,6 +51,36 @@ const incrCountersBy = (key, value) => {
...
 
+const initVisitor = (trackingID, userID) => {
+  return ua({
+    tid: trackingID,
+    cid: userID,
+    strictCidFormat: false,
+  })
+}
+
+const logActivity = (type, ...args) => {
```
refi93: I would split this into two separate functions - trackPageView and trackEvent (and rename `log` to `track` since `log` may be misleading) and I would move the error handling to trackVisits/trackTxSubmissions. In trackVisits we probably do not want to add the ip to `knownIps` if tracking the visit fails, so it makes sense to handle the error there.

In general, it makes sense to handle the error as high as possible.

server/middlewares/stats.js
```diff
@@ -78,6 +112,18 @@ const trackTxSubmissions = mung.json((body, req, res) => {
       const txOutAmount = parseTxBodyOutAmount(txBody)
 
       incrCountersBy(`${txSubmissionType}:sentOut`, txOutAmount)
+      logActivity('event', {
+        ec: 'Payments',
```
refi93: I would rather pass an object with human readable keys and reassign those keys to the google analytics object inside `logActivity`. In general, you want to keep the internals of the function from the caller, since it may change at any time.

server/middlewares/stats.js
```diff
@@ -4,8 +4,11 @@ const client = redis.createClient(process.env.REDIS_URL)
 const mung = require('express-mung')
 const normalizeUrl = require('normalize-url')
 const cbor = require('borc')
+const ua = require('universal-analytics')
 
 const knownIps = new Set()
+const adaliteTrackingID = 'UA-143401606-1'
+const anonymousUserID = 555
```
refi93: I would not declare this as a variable but rather hardcode it into "initVisitor" and leave there the comment "google analytics anonymous visitor". We are not dealing with visitors, it's just a google analytics specific, so there is no need to expose it.

server/middlewares/stats.js
```diff
@@ -48,6 +51,36 @@ const incrCountersBy = (key, value) => {
   client.incrby(`${key}:daily:${year}-${month}-${day}`, value)
 }
 
+const initVisitor = (trackingID, userID) => {
+  return ua({
+    tid: trackingID,
+    cid: userID,
+    strictCidFormat: false,
```
refi93: if this is really needed, it would be nice to put there a comment and explain the reason behind why this should be false, for future contributors

server/middlewares/stats.js
refi93: perhaps that's google analytics default behavior if the tracking id is undefined

server/helpers/loadConfig.js
```diff
@@ -43,6 +43,7 @@ const {
   ADALITE_MOCK_TX_SUMMARY_SUCCESS,
   ADALITE_TREZOR_CONNECT_URL,
   ADALITE_SUPPORT_EMAIL,
+  ADALITE_TRACKING_ID,
```
refi93: I would name this `ADALITE_GA_TRACKING_ID` to make clear to what service it is related

server/middlewares/stats.js
```diff
@@ -41,44 +41,81 @@ const parseTxBodyOutAmount = (txBody) => {
...
+          ? 'txSubmissions'
+          : 'otherTxSubmissions'
+      const txSubmissionSuccess = body.Right ? 'successful' : 'unsuccessful'
+
+      if (txSubmissionSuccess === 'successful') {
+        const {txBody} = req.body
+        const txOutAmount = parseTxBodyOutAmount(txBody)
+
+        trackEvent({
+          ec: 'Payments',
```
refi93: please use human readable key names

xdzurman: working on it atm

xdzurman: done

server/middlewares/stats.js
```diff
@@ -41,44 +41,81 @@ const parseTxBodyOutAmount = (txBody) => {
...
-
-    if (txSubmissionSuccess === 'successful') {
-      const {txBody} = req.body
-      const txOutAmount = parseTxBodyOutAmount(txBody)
+  if (!backendConfig.ADALITE_TRACKING_ID) {
+    return
+  }
 
-      incrCountersBy(`${txSubmissionType}:sentOut`, txOutAmount)
+  try {
```
refi93: the try-catch imho belongs inside the `if` block, I would wrap only the `if (txSubmissionSuccess === 'successful') {` block with it.

server/middlewares/stats.js
```diff
@@ -41,44 +41,97 @@ const parseTxBodyOutAmount = (txBody) => {
...
+    ea: argsObj.eventAction,
+    el: argsObj.eventLabel,
+    ev: argsObj.eventValue,
+  }
+
+  visitor.event(args).send()
 }
 
 const trackVisits = (req, res, next) => {
+  if (!backendConfig.ADALITE_TRACKING_ID) {
```
refi93: this was originally handled in `index.js`, see `if (backendConfig.REDIS_URL)`. Also, please remove the dead code, i.e. middlewares/basicAuth and statsPage since they are no longer needed. Also, please update .env.example

xdzurman: should I add '&& backendConfig.ADALITE_GA_TRACKING_ID' to the if?

server/middlewares/stats.js
```diff
@@ -41,44 +41,97 @@ const parseTxBodyOutAmount = (txBody) => {
...
+    ea: argsObj.eventAction,
+    el: argsObj.eventLabel,
+    ev: argsObj.eventValue,
+  }
+
+  visitor.event(args).send()
 }
 
 const trackVisits = (req, res, next) => {
+  if (!backendConfig.ADALITE_GA_TRACKING_ID) {
```
refi93: this was handled here: https://github.com/vacuumlabs/adalite/blob/2a7d7e9ea211beaadd947012f26bd8b652925798/server/index.js#L17 and I would rather keep it like that.

package.json
```diff
@@ -52,8 +52,7 @@
     "express-mung": "^0.5.1",
     "isomorphic-fetch": "^2.2.1",
     "normalize-url": "^4.0.0",
-    "redis": "^2.8.0",
```
refi93: please remove the packages via `yarn uninstall`, otherwise the package stays and `yarn.lock` and it isn't really removed

server/middlewares/stats.js
xdzurman: should I specify our tracking id to example or leave it blank?

server/middlewares/stats.js
```diff
@@ -71,13 +110,27 @@ const trackTxSubmissions = mung.json((body, req, res) => {
...
+      } else {
+        trackEvent({
+          eventCategory: 'Payments',
+          eventAction: `${txSubmissionType}:sentOutFail`,
+          eventLabel: 'unsuccessful payment',
+        })
+      }
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(`Tracking event failed - ${err.message}`)
```
refi93: I would rather log the whole error payload. The error may not always have the message property

server/middlewares/stats.js
```diff
@@ -71,13 +110,27 @@ const trackTxSubmissions = mung.json((body, req, res) => {
...
+
+        trackEvent({
+          eventCategory: 'Payments',
+          eventAction: `${txSubmissionType}:sentOutSuccess`,
+          eventLabel: 'successful payment',
+          eventValue: txOutAmount,
+        })
+      } else {
+        trackEvent({
+          eventCategory: 'Payments',
```
refi93: prefixing the key names with "event" seems redundant, if the function is called `trackEvent`

refi93:
please rebase to current dev, so CI tests pass

server/middlewares/stats.js
```diff
@@ -41,28 +41,67 @@ const parseTxBodyOutAmount = (txBody) => {
...
+      // eslint-disable-next-line no-console
+      console.error(`Tracking page view failed - ${err.message}`)
+    }
   }
 
   next()
 }
 
 const trackTxSubmissions = mung.json((body, req, res) => {
+  if (!backendConfig.ADALITE_GA_TRACKING_ID) {
```
refi93: this logic is redundant

server/middlewares/stats.js
```diff
@@ -41,22 +41,52 @@ const parseTxBodyOutAmount = (txBody) => {
...
   const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress
-
   const mydevice = device(req.headers['user-agent'])
 
   if (!knownIps.has(ip) && !mydevice.is('bot')) {
-    knownIps.add(ip)
-    incrCountersBy('visits', 1)
+    try {
+      trackPageView({
+        path: '/',
```
refi93: this function takes also `hostname` and `title` - I would either provide them explicitly here or set them explicitly (if needed) in the `trackPageView` function

server/middlewares/stats.js
```diff
@@ -41,22 +41,52 @@ const parseTxBodyOutAmount = (txBody) => {
...
+  const args = {
+    dp: argsObj.path,
+    dt: argsObj.hostname,
+    dh: argsObj.title,
+  }
+
+  visitor.pageview(args).send()
+}
+
+const trackEvent = (argsObj) => {
```
refi93: `argsObj` seems quite cryptic, you can make this more readable by destructuring the object in the function signature: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_fields_from_objects_passed_as_function_paramete

i.e. it can be
```
const trackEvent = ({category, action, label, value}) =>
...
```

server/middlewares/stats.js
xdzurman: they are optional, what hostname does production have? and usually people specify the title of the page the user is on, what should I do here since its a single page?

server/middlewares/stats.js
```diff
@@ -41,22 +41,52 @@ const parseTxBodyOutAmount = (txBody) => {
...
+  const visitor = initVisitor(backendConfig.ADALITE_GA_TRACKING_ID)
+  const args = {
```
refi93: I would not declare a separate `args` variable but I would directly pass the object to `visitor.event`

yarn.lock
```diff
@@ -303,6 +303,26 @@ ajv@^6.10.0:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
```
refi93: I see the `universal-analytics` module introduced quite a lot of new dependencies. Can you please try this one: https://www.npmjs.com/package/universal-ga which seems to have no dependencies?

server/middlewares/stats.js
refi93: the hostname can be extracted from the request, I would not set the title

yarn.lock
```diff
@@ -303,6 +303,26 @@ ajv@^6.10.0:
...
 
+ajv@^6.5.5:
```
refi93: I see the universal-analytics module introduced quite a lot of new dependencies. Can you please try this one: https://www.npmjs.com/package/universal-ga which seems to have no dependencies?

server/middlewares/stats.js
```diff
@@ -71,13 +106,40 @@ const trackTxSubmissions = mung.json((body, req, res) => {
...
+        if (txOutAmount > 0) {
+          trackEvent({
+            category: 'Payments',
+            action: `${txSubmissionType}:sentOutSuccess`,
+            label: 'successful payment',
+            value: txOutAmount,
+          })
+        }
+
+        const txChangeAmount = parseTxBodyChangeAmount(decodedTxBody)
```
refi93: we are not really interested in the change amount but in the total amount being sent

xdzurman: So that would be sum of sentAmount and changeAmount?

xdzurman: Do you want to track both sentAmount and totalAmount separately? Or should we track only totalAmount?

refi93: yes, that's the sum of all outputs, I would suggest using array.map for that. So I would track total amount and sentOut amount separately, both provide meaningful information - the total amount may serve to determine how is adalite daily volume related to Cardano total daily volume and the sent out amount serves to tell how big transactions are we actually dealing with from the point of view of the user.

server/middlewares/stats.js
```diff
@@ -23,40 +16,82 @@ const isSameOrigin = (urlString1, urlString2) => {
...
 const tokenMatches = (token) => token === process.env.ADALITE_BACKEND_TOKEN
 
-const parseTxBodyOutAmount = (txBody) => {
+const decodeTxBody = (txBody) => {
+  try {
+    return cbor.decode(txBody)
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e)
+    return []
```
refi93: this seems pretty dangerous, I would rather catch in `parseTxBodyOutAmount` and `parseTxBodyTotalAmount` and I would use `cbor.decode(txBody)` directly - extracting it to a function creates only an additional layer of indirection and saves a few characters at best

xdzurman: fixed, please check

xdzurman: the parsing functions now only have 1 line of code because of the change which seems way too thin but I kept them that way for documentation reasons (the comments attached to them)

PR #497, opened 2019-07-17
Add staked page and page banner

### LINKED ISSUE #496: Add staking page

ladomyriak:
Submitting form logic should be implemented

closes #496


app/public/css/styles.css
```diff
@@ -889,6 +891,45 @@ button[data-balloon] {
...
+  height: 24px;
+  display: inline-block;
+  margin-right: 16px;
+  vertical-align: middle;
+  background: url('../assets/icon-star.svg') no-repeat;
+  background-position: center center;
+}
+
+.banner-close {
+  float: right;
```
d-novak: The `float` is not needed, if you want to make extra sure it stays stuck to the right side you can add `flex: 1` to `.banner-text`

app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,33 @@
+const {h} = require('preact')
+
+const StakingPage = () =>
+  h(
+    'div',
+    {class: 'staking-wrapper'},
+    h('div', {class: 'staking-label'}, 'Upcoming'),
```
d-novak: The icon seems slightly off vertical center

app/frontend/components/common/footer.js
```diff
@@ -32,9 +32,10 @@ const Footer = connect(
           'div',
           {class: 'social'},
           h(
-            'a',
+            'button',
```
d-novak: Any particular reason why this is the only link that's actually a `<button>`? The `<a>` tag has also a permitted value for `role="button"` if you'd like to keep that.

d-novak:
@ladomyriak I suppose we don't have tablet / mobile designs but the contact modal has too thick of a padding in my opinion - I'd advise to reduce it. I don't mind the thick padding on iPads but the mobile version has too much padding for my taste. Also the buttons might look better if they were centered or stacked on top of each other.

Screenshot 2019-07-17 at 15 46 39

app/public/css/styles.css
ladomyriak: Great idea, thanks! Fixed

app/frontend/components/pages/staking/stakingPage.js
ladomyriak: Fixed, but check on the retina, please.

app/frontend/components/common/footer.js
ladomyriak: Discussed personally. Solved.

ladomyriak:
> @ladomyriak I suppose we don't have tablet / mobile designs but the contact modal has too thick of a padding in my opinion - I'd advise to reduce it. I don't mind the thick padding on iPads but the mobile version has too much padding for my taste. Also the buttons might look better if they were centered or stacked on top of each other.
> 
> Screenshot 2019-07-17 at 15 46 39


It was the bug in this PR. I fixed and it remains like it was. About buttons, I think it could be done in separate PR, after discussing with a designer.  Now it looks like on screenshot:

Screenshot from 2019-07-18 11-23-40

d-novak:
@ladomyriak Yep I agree, let's leave it as it is now and discuss with @martinmaricak about the design decision on those bottom buttons.

app/frontend/components/common/banner.js
```diff
@@ -0,0 +1,32 @@
+const {h} = require('preact')
+
+const Banner = ({closeBanner}) =>
```
refi93: I would not make a generic banner element if it is used only once. If it is part of the login page, I would rename it to `StakingBanner` and put it in `components/login`.

app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,41 @@
+const {h} = require('preact')
```
refi93: I miss there some button to be able to return back to AdaLite, perhaps, the `Staking` link in the navbar could become `Wallet` when you enter the staking page

ladomyriak: I think the back link would be more user-friendly. @martinmaricak what do you think about it (Link back from staking page)?

martinmaricak: I would leave `Subscribe` as the only action. User is able to navigate via clicking on the logo. The idea of improving the navigation sounds good, but what we actually want is to gather emails. So we would be able to reach out to a lot of users once we launch staking.

app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,33 @@
...
+
+const StakingPage = () =>
+  h(
+    'div',
+    {class: 'staking-wrapper'},
+    h('div', {class: 'staking-label'}, 'Upcoming'),
+    h('h2', {class: 'staking-title'}, 'Staking delegation and staking pool is comming to AdaLite'),
```
xdzurman: typo - coming*

app/frontend/components/common/banner.js
ladomyriak: Fixed

app/frontend/components/pages/staking/stakingPage.js
ladomyriak: Thanks, fixed!

refi93:
the "create new wallet button" does nothing when I'm on the staking page

martinmaricak:
**Buttons in the responsive contact form**

refi93:
and what will we display when the subscription is successful? A modal or the submit button changes text/color? cc @martinmaricak

martinmaricak:
@refi93 somethilng like that (I was not designing the modal, so this is just a very rough sketch). 
- Just put a green text with the same size as the text below "Contact Us" headline. 
- Make the paddings larger than they are at the draft I've sent
- The text can be "Thank you for your message. We will contact you shortly."


61451279-4e28f000-a959-11e9-8994-f64312f0f347

ladomyriak:
> @refi93 somethilng like that (I was not designing the modal, so this is just a very rough sketch).
> 
> * Just put a green text with the same size as the text below "Contact Us" headline.
> * Make the paddings larger than they are at the draft I've sent
> * The text can be "Thank you for your message. We will contact you shortly."

@martinmaricak We also need some message for submitted staking form.

martinmaricak:
@ladomyriak https://share.goabstract.com/37f84e54-d8c5-4bb5-8543-3ec85099ef96

martinmaricak:
centrovanie_hviezdy

To things regarding the staking banner:
1. star is not vertically centred (2px lower than it should be) on both desktop and tablet versions.
2. the banner is very tall compared to the designed version. Height should be 56px.

PR #498, opened 2019-07-19
I 412 add back button to export keyfile screen

xdzurman:
closes #498 
image (1)

### DIFF
--- a/app/frontend/components/pages/exportWallet/exportWalletPage.js
+++ b/app/frontend/components/pages/exportWallet/exportWalletPage.js
@@ -155,7 +155,15 @@ class ExportWalletDialog extends Component {
             }),
             h(
               'div',
-              {class: 'validation-row'},
+              {class: 'export-content-bottom'},
+              h(
+                'button',
+                {
+                  class: 'button secondary',
+                  onClick: () => window.history.back(),
+                },
+                'Back'
+              ),
               h(
                 'button',
                 {
--- a/app/public/css/styles.css
+++ b/app/public/css/styles.css
@@ -2419,6 +2419,15 @@ button[data-balloon] {
   margin-top: 24px;
 }
 
+.export-content-bottom {
+  display: flex;
+  align-items: center;
+}
+
+.export-content .button + .button {
+  margin-left: 24px;
+}
+
 /* VALIDATION */
 
 .validation-row {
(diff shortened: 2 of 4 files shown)


refi93:
when you enter the `https://<host:port>/exportWallet` page (directly, not via the login page), it first redirects you to the login page, and when you log in, it redirects you to the export wallet page, but the back button does not work as expected (you need to click it several times until you get to the wallet page).

I would suggest fixing this as well - my idea is that if you login via adalite it should bring you always to the txHistory page to avoid the conflict of what the "back" button is supposed to do if you did not really come from the `txHistory` page

app/public/css/styles.css
```diff
@@ -3025,7 +3034,18 @@ button[data-balloon] {
...
 
   .export-content .button {
-    width: 100%;
+    flex: 1;
+    white-space: nowrap;
+    padding: 10px 15px;
+  }
+
+  .export-content-bottom {
+    margin-bottom: 20px;
```
refi93: I think the general convention is to use multiples of 8, can you confirm @d-novak ?

app/frontend/components/pages/exportWallet/exportWalletPage.js
```diff
@@ -155,7 +155,15 @@ class ExportWalletDialog extends Component {
             }),
             h(
               'div',
-              {class: 'validation-row'},
+              {class: 'export-content-bottom'},
+              h(
+                'button',
+                {
+                  class: 'button secondary',
+                  onClick: () => window.history.back(),
```
refi93: when you enter the https://<host:port>/exportWallet page (directly, not via the login page), it first redirects you to the login page, and when you log in, it redirects you to the export wallet page, but the back button does not work as expected (you need to click it several times until you get to the wallet page).

I would suggest fixing this as well - my idea is that if you login via adalite it should bring you always to the txHistory page to avoid the conflict of what the "back" button is supposed to do if you did not really come from the txHistory page

app/public/css/styles.css
xdzurman: inspired by   
.contact-form-bottom {
    margin-bottom: 20px;
    align-items: stretch;
  },

d-novak: Yes, the spacing convention back when I implemented the redesign was that we use only values dividable by 8, in edge cases dividable by 4 (like when you need something between `0 - 8px`). I haven't seen the new designs for the contact form (modal) but the spacing there should also be `8 * X`, thanks for pointing that out. I think the best idea is to leave the `.contact-form-bottom` as it is for now, @ladomyriak is implementing this part in another PR so maybe he can take a look at the spacings there as well.

PR #502, opened 2019-07-29
Mailchimp integration

xdzurman:
depends on #496

------NOTE------
when testing, Mailchimp API KEY is required. I cannot commit mine to github, since it grants account access. PM me for it.


app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,102 @@
...
+            type: 'email',
+            placeholder: 'Enter your email to get notified',
+            value: email,
+            required: true,
+            onInput: this.updateEmail,
+          }),
+          h(
+            'button',
+            {
+              onClick: this.handleSubmit.bind(this),
```
PeterBenc: All the other methods are bound in the constructor, maybe I would bind this one too, just to stay consistent :).

app/frontend/components/common/contactForm.js
```diff
@@ -55,6 +59,13 @@ class ContactForm extends Component {
             method: 'POST',
             target: '_blank',
             action: `//formspree.io/${ADALITE_CONFIG.ADALITE_SUPPORT_EMAIL}`,
+            onSubmit: () => {
+              this.setState({sumbitted: true})
```
PeterBenc: Maybe you could create a method which changes the state of the the contactForm. It's just a detail but we dont use setState method inside render method anywhere in the project, so again just to stay consistent :).

app/frontend/components/pages/staking/stakingPage.js
xdzurman: fixed, originally it was a workaround inspired by /reactconf

app/frontend/components/common/contactForm.js
xdzurman: This is the code from the issue branch that this branch depends on so it's not mine, can you put this comment there so the author decides? :)
-> https://github.com/vacuumlabs/adalite/pull/497

PR #503, opened 2019-07-30
I 499 validate password when exporting wallet mnemonic

xdzurman:
Should solve 499


xdzurman:
> Have you tested it? I'm getting an error `Uncaught TypeError: this.isPasswordValid(...) is not a function` when trying to fill the password for exporting the wallet.
> 
> Also, not really sure whether this is a question for me as a frontend engineer whether the confirmation touch is redundant or not.

Sorry, yes, I tested it and after it I quickly added the comment and pushed, but I confused the comment syntax with other languages :/. Fixed now

Issue #488, opened 2019-07-01
Forward errors to Sentry


refi93:
Currently the client side errors have no way to getting to us so we know that something went wrong only from user reports which isn't reliable. The idea is to show a modal each time an unexpected error happens and give the user the option to forward the error to Sentry or other error tracking service.

Issue #388, opened 2019-02-19
Refactor error handling and error messages


refi93:
Currently exceptions are being modified at different places in different ways - sometimes a NamedError is thrown, sometimes the error we've got from a module, sometimes the "sendResponse" is taken as a source for the error message. This is inconsistent and makes debugging harder and many error messages and exceptions are more generic than they deserve to be.

We should unify the exceptions, have a consistent format and way of handling for them.

It's probably worth starting solving after we merge the redesign branch

PR #505, opened 2019-08-12
I 488 forward errors to sentry

PeterBenc:
Should solve issue #488 , Sentry should in a way, capture all the errors except those which are explicitly set to not be captured. Creating a list of errors which should not be captured by Sentry, in order to display the unexpectedErrorModal with the option to send the exception to Sentry, the modal can be triggered by throwing an exception which is not part of this list and then later capturing it by Sentry.


app/frontend/helpers/expectedErrors.js
```diff
@@ -0,0 +1,20 @@
+const expectedErrors = {
+  ConversationRatesError: () => '',
```
xdzurman: a simple array would be sufficient to store these values if you don't intend to do something else

app/frontend/walletApp.js
```diff
@@ -51,6 +53,25 @@ window.onhashchange = () =>
     },
   })
 
+Sentry.init({
+  dsn: 'https://d77d3bf9d9364597badab9c00fa59a31@sentry.io/1501383',
```
xdzurman: don't forget to put this in some kind of config

app/public/css/styles.css
```diff
@@ -553,6 +558,8 @@ button[data-balloon] {
 
 .alert-content {
   text-align: left;
+  max-height: 300px;
+  overflow: hidden;
```
xdzurman: overflow: scroll/auto; would work better with long exceptions

app/frontend/actions.js
```diff
@@ -496,12 +511,17 @@ module.exports = ({setState, getState}) => {
         error: e.name,
         message: e.message,
       }
+      if (captureBySentry(e)) {
+        return
+      }
+      setState({
+        showTransactionErrorModal: !sendResponse.success,
```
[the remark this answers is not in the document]
PeterBenc: Yeah, no problem, I just moved the line from finally to catch and didn't change it because it served the same purpose.

app/frontend/helpers/expectedErrors.js
PeterBenc: Good point, I just reused the code from translations since I had plans to merge the two files, but later decided to solve that in later in #388 so I left it this way since it server the short-term purpose.

app/public/css/styles.css
PeterBenc: I was inspired by myEtherwallet, anyway I tried the "scroll" option, but the modal looks visually much worse with two scrollbars so I changed it back.

app/frontend/components/common/unexpectedErrorModal.js
```diff
@@ -0,0 +1,70 @@
...
+        JSON.stringify(sendSentry.event)
+      ),
+      h(
+        'div',
+        {class: 'modal-footer send-error'},
+        h(
+          'button',
+          {
+            class: 'button outline',
+            onClick: () => {
```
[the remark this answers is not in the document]
PeterBenc: Good point, I thought about that too but had run into some problems so I changed it back. Anyway, Fixed.

app/frontend/walletApp.js
PeterBenc: Again I was inspired by myEtherwallet, but it definitely could be in .env and let it be added by Heroku but it didn't seem necessary.

PeterBenc:
Closing due to another open PR containing proposed changes.

PR #507, opened 2019-08-26
Donations

xdzurman:
Implements 3 donation options and adds a summary extension to 'Send ADA' screen.


martinmaricak:
@xdzurman 

1. There are unbalanced stroke widths between fonts and icons (check screenshot). It's not like that in design files. Please try to replace font with this one: https://drive.google.com/open?id=1q5rHTXB983eely41oZi0vRIbOSiayYjf

Screenshot 2019-10-02 at 15 16 18

Screenshot 2019-10-02 at 15 30 42


2. This issue is related to the one above, but I think it deserves a separate point. :) Please vertically align the ADA symbol properly. Also, please replace it with this one (the current one has poor quality): https://drive.google.com/open?id=15MI1Vx-00rDsXPLo2rLfipYWZmC5IMgW

Screenshot 2019-10-02 at 15 29 06


3. When user selects the amount too high so he won't be able to use donation preset, please remove the button (e. g. if I will be able to donate 0,1%, but won't be able to donate 40 ADA, show just 0,1%).

Screenshot 2019-10-02 at 15 25 01


4. If user selects a custom donation which combined with the transaction amount is larger than balance, we should differentiate the error message. In deisgns `Insufficient balance for a donation.` message is proposed for such occurence.

Screenshot 2019-10-02 at 15 26 34


cc @MichalPetro @jmuravsky @adamrybak1

xdzurman:
I agree with points 1,2 and 4, but we discussed point 3 with Miso, that hiding the buttons may not be such a good idea and we do not want buttons to jump around for the user. However I need @MichalPetro 's confirmation on this.

martinmaricak:
@xdzurman 

- I can live with point 3 not being implemented even though I don't think that buttons will "jump around". :) It might happen once, but I don't think users would play around with transaction amount, moving it up and down frenetically causing the buttons dance.

In my opinion, it's better to hide the button instead of showing an error for those, who click on it.

- Also, I don't understand the decision to introduce the `Back` link to the custom donation input. Like when I'm using the app, I just click `Custom` and select the amount. There's no reason for heading back and selecting 40ADA. It's even slower than typing `40` I guess.

xdzurman:
@martinmaricak

martinmaricak:
@xdzurman 

Great. Please implement ponit 3 as well then. :)

I'm not sure if I understand what reason it would possibly be, but not saying, there isn't one. I'm not pushing on removing the `Back`. Just expressed my opinion.

martinmaricak:
@xdzurman Maybe `Close` would better describe the action of hiding the custom input. 🤔

xdzurman:
I need to discuss point 3 with Miso, I'll let you know.

xdzurman:
@martinmaricak 

Hmm, wouldn't just disabling and grey-ening borders of unaffordable donation options be less destructive, but it would still solve the problem?

martinmaricak:
@xdzurman OK, let's leave calling the shots on @MichalPetro. But I would like to take part in a discussion about that.

martinmaricak:
@xdzurman @MichalPetro I also thought about disabling the button as a first thing but I'm not sure about having the fourth state of the button.

First, it's grey outline, then it turns pink outline to alert a user about its presence, and then, when you select some option, it goes solid pink.

From the three non-ideal options (leaving the button, hiding it, disabling it), hiding seems to be the least confusing one.

xdzurman:
@martinmaricak 

It would be grey outline, same as % donation when threshold amount 500 Ada has not been reached. Or am I missing something?

martinmaricak:
@xdzurman Visually same buttons doing different things in different contexts is definitely the worst option.

martinmaricak:
@xdzurman If you insist on disabling the button, please do it as suggested below so we will not use one state for two different things.

Screenshot 2019-10-03 at 16 31 06

martinmaricak:
Actually, I like it with a disabled button, thanks for convincing me. 👍Just please don't use this grey outline variation.

martinmaricak:
@xdzurman Also can we have [a tooltip](https://getbootstrap.com/docs/4.0/components/tooltips/#disabled-elements) saying `Insufficient funds`?

app/frontend/wallet/cardano-wallet.js
```diff
@@ -139,13 +151,56 @@ const CardanoWallet = async (options) => {
...
+    return getMaxSendableAmountWithDonation(address, txInputs, coins, donationAmount, donationType)
+  }
+
+  async function getMaxDonationAmount(address, sendAmount) {
+    const {txInputs, coins} = await getTxInputsAndCoins()
+    const txFee = computeTxFee(txInputs, address, sendAmount, true, coins - sendAmount)
+    return Math.max(coins - txFee - sendAmount, 0)
+  }
+
+  async function getTxFee(address, coins, hasDonation, donationAmount) {
```
PeterBenc: It seems you are using the `hasDonation` argument in couple of functions in this file, but it occurs to me, that the following argument, `donationAmount` only makes sense in case the `hasDonation` is `true`. Wouldn't it be more simple to remove the `hasDonation` argument altogether and check whether `donationAmount` isn't undefined? What do you think?

xdzurman:
Already merged.

PR #518, opened 2019-09-11
I-388 refactor error handling and error messages

PeterBenc:
Adds error capturing by sentry, including sending exceptions and user feedback to sentry. Adds unexpected error modal for displaying errors that could be send to sentry. Adds help section to error modal in some error cases. Refactors and unifies the way errors are handled and fixes some minor issues with displaying error messages.


app/frontend/helpers/submitFeedbackToSentry.js
```diff
@@ -0,0 +1,30 @@
...
+async function sendFeedback(comments, email, name, eventId) {
+  await sleep(5000) // to ensure the feedback gets send after the error
+  if (!comments) return
+  const params = {
+    event_id: eventId,
+    name,
+    email,
+    comments,
+  }
+  const url = 'https://sentry.io/api/0/projects/vacuumlabs-sro/adalite-frontend/user-feedback/'
```
xdzurman: check constants.js file, I'd place the URL there. It'd be easier to maintain.

app/frontend/actions.js
```diff
@@ -60,13 +63,50 @@ module.exports = ({setState, getState}) => {
...
+            message: e.message,
+            ...options,
+          },
+        },
+        error: e,
+      })
+    } else {
+      setState({
+        [errorName]: e,
+      })
```
xdzurman: why does this function do anything with state? From what I see, you extracted some state-setting from calling functions and moved them here.

app/frontend/actions.js
```diff
@@ -60,13 +63,50 @@ module.exports = ({setState, getState}) => {
...
+        conversionRates: await conversionRates,
+      })
+    } catch (e) {
+      debugLog('Could not fetch conversion rates.')
+      setState({
+        conversionRates: null,
+      })
+      throw NamedError('ConversionRatesError', '`Could not fetch conversion rates.')
+    }
+    return true
```
xdzurman: I don't see any caller function use this 'true' return value

app/frontend/actions.js
```diff
@@ -266,29 +281,28 @@ module.exports = ({setState, getState}) => {
...
+
+      // timeout setting loading state, so that loading shows even if everything was cached
+      setTimeout(() => setState({loading: false}), 500)
       setState({
-        conversionRates: await conversionRates,
+        balance,
+        ownAddressesWithMeta,
+        transactionHistory,
       })
+      await fetchConversionRates(state, conversionRates)
```
xdzurman: fetchConversionRates takes one parameter

app/frontend/actions.js
PeterBenc: I dont really see a problem with this function changing state, after all all the functions in `actions.js` change state. Yes I extracted some state changing functionality from calling function and moved it here, I see several advantages to this solution:

1. This function ensures that setting the error state is handled the same way every time an error occurs and its handling is defined in one place.

2. This way, setting an error state is dedicated from setting other states, in lot of cases states that has nothing to do with the fact that an error has occurred.

3. The extracted code was in lot of cases redundant since I tried to make the error handling as consistent as possible. This function spared us tens of redundant lines.

4. The function supports the way I implemented the sentry error capturing, since every error is handled by this function, every error checked against translations whether the error is expected or not, instead of calling `captureBySentry` all over the code.

What do you think?

app/frontend/helpers/submitFeedbackToSentry.js
PeterBenc: thanks for suggestion, makes sense..

app/frontend/actions.js
PeterBenc: Ah thanks, I removed the `state` argument from two occurrences, forgot to remove it in the third. I'll fix it.

app/frontend/actions.js
PeterBenc: Thanks, you are right, there is no use for returning true.

app/frontend/components/common/unexpectedErrorModal.js
```diff
@@ -0,0 +1,116 @@
...
+        h('textarea', {
+          placeholder: 'Your message',
+          autocomplete: 'off',
+          name: 'message',
+          class: 'input fullwidth textarea',
+          required: true,
+          onBlur: (e) => {
+            this.props.updateMessage(e)
+          },
+        })
```
xdzurman: why do all of these have 'required: true'?

app/frontend/actions.js
```diff
@@ -358,9 +379,36 @@ module.exports = ({setState, getState}) => {
...
+      state.userEmail || 'user@email.com',
+      state.userName || 'user',
+      state.sendSentry.event.event_id
+    )
+  }
+
   const debouncedCalculateFee = debounceEvent(calculateFee, 2000)
 
-  const validateSendFormAndCalculateFee = () => {
+  const validateSendFormAndCalculateFee = (state) => {
```
xdzurman: why add state?

app/frontend/actions.js
```diff
@@ -385,29 +433,30 @@ module.exports = ({setState, getState}) => {
...
     validateSendFormAndCalculateFee()
   }
 
   const sendMaxFunds = async (state) => {
     setState({calculatingFee: true})
-
-    const maxAmount = await wallet.getMaxSendableAmount(state.sendAddress.fieldValue)
-
-    if (maxAmount > 0) {
+    let maxAmount
```
xdzurman: maxAmount is not used beyond the scope of try{} so you can safely declare it as 'const' inside the block instead of 'let' outside

app/frontend/actions.js
```diff
@@ -385,29 +433,30 @@ module.exports = ({setState, getState}) => {
...
+    try {
+      maxAmount = await wallet.getMaxSendableAmount(state.sendAddress.fieldValue)
       setState({
         sendResponse: '',
-        sendAmount: sendAmountValidator(printAda(maxAmount)),
+        sendAmount: {
+          fieldValue: printAda(maxAmount),
+          coins: maxAmount || null,
```
xdzurman: before the changes, maxAmount was being checked if it is grater than zero. Are you sure removing it is safe?

PeterBenc: I am quite sure, first of all, the max amount cant be less then 0, check `getMaxSendableAmount` function in `cardano-wallet.js`, the least amount it can return is 0 in which case the maxAmount is set to `null`. Secondly, the amount gets validated either way so there is no way, negative value would pass.

app/public/css/styles.css
```diff
@@ -764,6 +775,13 @@ button[data-balloon] {
   white-space: pre-wrap;
 }
 
+.modal-instructions {
+  margin-top: 20px;
+  margin-bottom: 20px;
+  margin-right: 24px;
+  margin-left: 24px;
+}
```
xdzurman: this can be shortened to 
.modal-instructions {
  margin: 20px 24px;
}

read more about shorthand syntax here: https://www.w3schools.com/css/css_margin.asp

app/public/css/styles.css
```diff
@@ -553,6 +562,8 @@ button[data-balloon] {
 
 .alert-content {
   text-align: left;
+  max-height: 180px;
+  overflow: hidden;
 }
```
xdzurman: did you consult the overflow property with someone from design? I don't think its a good idea to hide scrollbars and just cut the content, but I might be wrong

app/frontend/helpers/captureBySentry.js
```diff
@@ -0,0 +1,10 @@
+const {getTranslation} = require('../translations')
+
+function captureBySentry(e) {
+  if (!getTranslation(e.name)) {
+    throw e
+  }
+  return
```
xdzurman: is this 'return' needed for anything?

app/frontend/translations.js
```diff
@@ -1,43 +1,51 @@
...
+
+  CryptoProviderError: ({message}) => `CryptoProviderError:${message}`,
+  NetworkError: () =>
+    'NetworkError:Network connection failed. Please check your network connection.',
 }
 
 function getTranslation(code, params) {
   if (!translations[code]) {
     debugLog(`Translation for ${code} not found!`)
+    return null
```
xdzurman: the function previously returned the 'code' it received as a parameter, now 'null' is returned in such a scenario. Now the calling functions assign null to errors instead of at least code. Is this correct behavior?

app/frontend/actions.js
PeterBenc: resolved during merging

app/frontend/actions.js
PeterBenc: resolved during merging

app/frontend/components/common/unexpectedErrorModal.js
PeterBenc: Thanks, I copy-pasted the code from `contactForm.js` and forgot to remove these lines, they serve no purpose here.

app/frontend/components/common/unexpectedErrorModal.js
```diff
@@ -0,0 +1,116 @@
...
+            type: 'text',
+            autocomplete: 'off',
+            placeholder: 'Your name',
+            name: 'name',
+            class: 'input fullwidth',
+            required: true,
+            onBlur: (e) => {
+              this.props.updateName(e)
+            },
+          }),
```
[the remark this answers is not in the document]
PeterBenc: I dont think we need their names, thanks.

app/frontend/helpers/captureBySentry.js
PeterBenc: fixed

app/frontend/translations.js
PeterBenc: I think this is a correct behavior. Previously, `code` was returned in case it didnt have a translation, but currently, this can never happen since every expected error has a translation, and if not, unexpectedErrorModal should be displayed which relies on this function returning `null` in that case.

app/public/css/styles.css
PeterBenc: fixed

app/public/css/styles.css
PeterBenc: I did not consult this with anyone from design, but we can discuss this on team meeting, I added the `max-height` and `overflow: hidden` just because of the unexpectedErrorModal in which I am displaying part of the json that is about to be send to sentry. But I dont really see a problem with the hidden overflow. First of all, there is no other case in which this gets applied other than unexpectedErrorModal since we simply dont insert that much content into the alert, secondly IMO the modal looked a lot worse with two scrollbars but I am open to other opinions.

app/frontend/translations.js
```diff
@@ -1,43 +1,51 @@
...
+      ''} not found in blockchain after being submitted, check it later please.`,
+  TrezorSignTxError: ({message}) => `TrezorSignTxError:${message}`,
+  LedgerOperationError: ({message}) => `LedgerOperationError:${message}`,
+  TxSerializationError: ({message}) => `TxSerializationError:${message}`,
+  CoinAmountError: () => 'CoinAmounError:Unsupported amount of coins.',
+  TrezorError: ({message}) => `TrezorError:${message}`,
+
+  CryptoProviderError: ({message}) => `CryptoProviderError:${message}`,
+  NetworkError: () =>
+    'NetworkError:Network connection failed. Please check your network connection.',
```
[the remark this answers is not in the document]
PeterBenc: Thanks, I fixed the issue with whitespaces between the error name and message, I had previously divided the errors into groups on purpose, but It wasnt perfect, I fixed it.

Issue #513, opened 2019-09-04
Copy address on click on mobile devices


PeterBenc:
Currently, users on mobile devices have to copy the wallet address manually, more efficient would be to copy it on click and show a message saying the address was copied.

PR #520, opened 2019-09-18
I 513 copy address on click on mobile devices

PeterBenc:
Solves #513

### DIFF
--- a/app/frontend/components/pages/receiveAda/addressItem.js
+++ b/app/frontend/components/pages/receiveAda/addressItem.js
@@ -11,8 +11,8 @@ class AddressItem extends Component {
     this.openAddressDetail = this.openAddressDetail.bind(this)
   }
 
-  openAddressDetail({address, bip32path}) {
-    this.props.openAddressDetail({address, bip32path})
+  openAddressDetail({address, bip32path}, enableTooltip) {
+    this.props.openAddressDetail({address, bip32path}, enableTooltip)
   }
 
   updateDimensions() {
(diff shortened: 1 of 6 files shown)


app/public/css/styles.css
```diff
@@ -3207,14 +3207,24 @@ button[data-balloon] {
...
-  [data-balloon]:hover::before,
-  [data-balloon]::after,
   [data-balloon]::before {
     display: none;
   }
 
+  [data-balloon]::hover::after,
+  [data-ballon]::hover::before{
+    display: none;
+  }
```
xdzurman: these two can be joined together if they do same stuff

app/frontend/components/pages/receiveAda/addressItem.js
```diff
@@ -1,68 +1,109 @@
...
+  }
+
+  openAddressDetail({address, bip32path}, copyOnClick) {
+    this.props.openAddressDetail({address, bip32path}, copyOnClick)
+  }
+
+  updateDimensions() {
+    if (window.innerWidth < 768) {
+      this.setState({onSmallDevice: true})
+    } else if (window.innerWidth > 767) {
```
xdzurman: this is a little bit strange, why not check for <= 767 in the first 'if' and have just 'else' instead of 'else if'?

PeterBenc: You are right, dont know how that happened, :smile:  it is strange, fixed.

PeterBenc:
> image
> All addresses are now links on desktop devices as well (not in design), I think adding the button as it is on bigger devices would eliminate the inconsistency and would not change addresses to links. Is there a possibility of keeping the copy functionality on mobiles and not underlining addresses? If not, I'd opt for button solution.

Thanks, I havent noticed it, fixed..

app/public/css/styles.css
PeterBenc: fixed

PR #522, opened 2019-10-05
Staking page and mailchimp integration

xdzurman:


app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,103 @@
...
+            class: 'staking-form',
+            id: 'stakingForm',
+          },
+          h('input', {
+            class: 'input',
+            type: 'email',
+            placeholder: 'Enter your email to get notified',
+            value: email,
+            required: true,
+            onInput: this.updateEmail,
```
PeterBenc: Currently the error message shows as soon as user starts typing the email, saying it has invalid format even though the user is clearly not done typing and the error message stays displayed even though the user clears the input. IMO update email `onBlur` and remove the `disabled` property from submitButton altogether. This way, no error message is shown until a user tries to submit. What do you think?

app/frontend/components/pages/staking/stakingPage.js
```diff
@@ -0,0 +1,103 @@
...
+    const re = /^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$/gim
+    return re.test(email)
+  }
+
+  updateEmail(e) {
+    const isEmailValid = this.isValidEmail(e.target.value)
+    this.setState({
+      email: e.target.value,
+      emailValid: isEmailValid,
+      errorMessage: !isEmailValid && 'Invalid email format',
```
PeterBenc:
```js
errorMessage: !isEmailValid && e.target.value !== '' && 'Invalid email format',
```
This way the email stays invalid in case its empty, but the error message does not get displayed.

xdzurman:
Already merged.

Issue #529, opened 2019-11-14
Improve loading of the login page


markopuzav:
**Current behaviour**

With slow connection, the icons/images on the login page take some amount of time to load. This is most noticeable on the hardware wallet login page, where content jumping occurs and the "Unlock with" buttons are missing the corresponding "Trezor"/"Ledger" for a couple of seconds.

This is visible (very briefly) even with fast connection.

**Proposed changes**

- Eliminate content jumping.
- Move the relevant icons from `::before` and `::after` pseudoelements into Preact generated `svg` elements, for faster loading.

![slow_load](https://user-images.githubusercontent.com/57289782/68856646-544d9400-06e1-11ea-8009-91fd129da230.gif)

PR #531, opened 2019-11-15
I 529 improve loading of the login page

markopuzav:
Targetting #529 

slow_load_fixed


app/frontend/components/pages/login/hardwareAuth.js
```diff
@@ -1,6 +1,9 @@
 const {h} = require('preact')
 const {CRYPTO_PROVIDER_TYPES} = require('../../../wallet/constants')
 
+const {TrezorLogoWhite} = require('../../common/svg')
+const {LedgerLogoWhite} = require('../../common/svg')
```
xdzurman: This can be shortened: const {TrezorLogoWhite, LedgerLogoWhite} = require('../../common/svg')

app/public/css/styles.css
```diff
@@ -1270,6 +1260,13 @@ button[data-balloon] {
   max-height: 178px;
 }
 
+.authentication-image-container {
+  position: relative;
+  padding-bottom: calc(178px + 24px);
```
xdzurman: What is the purpose of this? From what I know, calc is used mainly for percentages combined with pixel values.

app/frontend/components/common/svg.js
```diff
@@ -1,5 +1,84 @@
...
+      height: '20',
+      viewBox: '0 0 82 20',
+    },
+    h(
+      'g',
+      {
+        fill: '#FFF',
+        stroke: 'none',
+        // Preact currently doesn't properly parse some CamelCase svg attributes
+        //     (strokeWidth, filLRule, ...). Fixed in Preact X.
```
markopuzav: https://github.com/preactjs/preact/issues/813

app/frontend/components/pages/login/hardwareAuth.js
markopuzav: Fair enough.

app/public/css/styles.css
markopuzav: This was just to emphasize that the constant comes from the addition of the image height and its margin. Moving this information to a comment.

PR #532, opened 2019-11-20
Integrate js_chain_libs

PeterBenc:
Integrates js-chain-libs. The .wasm file is imported asynchronously before the page renders (in order to import the .wasm file , CSP had to be set to 'unsafe-eval' and MIME type 'application/wasm' had to be added).


app/frontend/walletApp.js
```diff
@@ -76,6 +76,6 @@ Sentry.init({
 
 const Wrapper = h(Provider, {store}, h(App))
 
-loadWasm().then(() => {
+loadWasm().finally(() => {
```
PeterBenc: replaced .then with .finally since in case of rejected promise, whole page would not render. It makes more sense to me to render it anyway and catch the possible error somewhere else, with Sentry lets say.

SebastienGllmt:
Just an FYI that if you use IOHK's version of js-chain-libs you will be missing a lot of functionality.

You can find the EMURGO fork here: https://github.com/Emurgo/js-chain-libs (look at the PR history to see which features we've added on top of the IOHK version)

If you just want to built code I host them here: https://github.com/SebastienGllmt/js-chain-libs-pkghttps://github.com/SebastienGllmt/js-chain-libs-pkg
My work also comes with Flow type definitions inside it also.

We still make updates to our fork occasionally so be sure to watch for any updates.

server/middlewares/csp.js
```diff
@@ -4,7 +4,7 @@ const directives = {
   'connect-src': ['*'],
   'img-src': ["'self'", 'data:'],
   'script-src': [
-    "'self'",
+    "'self' 'unsafe-eval'",
```
ppershing: is unsafe-eval still required?

server/index.js
```diff
@@ -4,9 +4,14 @@ const compression = require('compression')
 const fs = require('fs')
 const https = require('https')
 const {frontendConfig, backendConfig} = require('./helpers/loadConfig')
+const removeChainLibsWasmImport = require('./helpers/removeChainLibsWasmImport')
 
 let app = express()
 
+express.static.mime.types.wasm = 'application/wasm'
+
+removeChainLibsWasmImport()
```
ppershing: this is not good. it will probably not work in non-devel mode. Instead, we should just run `node helpers/remove...` as a part of `package.json` scripts for running the server

server/middlewares/csp.js
PeterBenc: I'm afraid yes, without the unsafe-eval the wasm streaming initiation fails and I havent been able to find any other solution to this.

server/index.js
PeterBenc: agreed, I included it in the `build` script.

PR #533, opened 2019-11-21
Staking UI

markopuzav:
UI for delegating to staking pools.


app/frontend/components/pages/txHistory/txHistoryPage.js
```diff
@@ -64,11 +67,14 @@ const TxHistoryPage = connect(
...
         {class: 'dashboard-column'},
         h(Balance, {balance, reloadWalletInfo, conversionRates}),
+        h(DelegationHistory),
         h(TransactionHistory, {transactionHistory, conversionRates})
       ),
       h(
         'div',
         {class: 'dashboard-column'},
+        h(DelegatePage),
+        h(CurrentDelegationPage),
```
markopuzav: Temporary - the cards will move to a staking tab.

app/frontend/actions.js
```diff
@@ -26,6 +26,8 @@ const submitEmailRaw = require('./helpers/submitEmailRaw')
 const captureBySentry = require('./helpers/captureBySentry')
 const submitFeedbackToSentry = require('./helpers/submitFeedbackToSentry')
 
+const adalitePoolAddress = 'adalite-pool-address'
```
xdzurman: you can prepare this for later by putting it in your .env file, check how it's loaded from it in loadConfig.js

app/frontend/components/pages/delegations/currentDelegationPage.js
```diff
@@ -0,0 +1,102 @@
...
+          'div',
+          {class: 'stakepool-name-wrapper'},
+          h('p', {class: 'stakepool-name'}, `${currentStakepool.name}`),
+          h(
+            'button',
+            {
+              class: 'button undelegate',
+              id: 'add-pool',
+              onClick: undelegateStakepool,
+              disabled: false,
```
xdzurman: is the 'id' and 'disabled' needed?

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,58 @@
...
+    h(
+      'button',
+      {
+        class: 'button delegate',
+        id: 'add-pool',
+        onClick: setAdalitePool,
+        disabled: false,
+      },
+      'AdaLite stake pool'
+    ),
```
xdzurman: Personally I'd rather have it as an underlined pink text inside the text input, as it is implemented with sending Max funds 
image but that can be discussed.

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,58 @@
...
+    h(
+      'div',
+      {class: 'validation-row'},
+      h(
+        'button',
+        {
+          class: 'button primary',
+          disabled: currentStakepool.address !== '',
```
xdzurman: image
I propose showing a tooltip saying 'Already delegated' or something like that over the disabled button, its not important, just a nice thing to have. Find inspiration here:
image

app/frontend/components/pages/delegations/delegatePage.js.rej
```diff
@@ -0,0 +1,9 @@
+diff a/app/frontend/components/pages/delegations/delegatePage.js b/app/frontend/components/pages/delegations/delegatePage.js	(rejected hunks)
+@@ -5,6 +5 @@ const actions = require('../../../actions')
+-const DelegatePage = ({
+-  poolAddress,
+-  updatePoolAddress,
+-  setAdalitePool,
+-  submitDelegation,
+-}) => {
++const DelegatePage = ({poolAddress, updatePoolAddress, setAdalitePool, submitDelegation}) => {
```
xdzurman: whats this

app/public/css/styles.css
```diff
@@ -478,6 +482,75 @@ p {
...
 
+.button.undelegate {
+  border: 1px solid var(--color-grey);
+  font-size: 14px;
+  font-weight: 400;
+  background-color: transparent;
+  padding: 0px 4px;
+  color: var(--color-grey);
+  white-space: nowrap;
+}
```
xdzurman: regarding all css pixel numbers, we try to follow multiples-of-4 convention, unless it's absolutely needed otherwise. That applies to all other css changes. Also, this button is way too small. Please give it static width and height so that it looks more like other buttons concerning size.

app/public/css/styles.css
```diff
@@ -1830,6 +1903,45 @@ button[data-balloon] {
...
 
+/* REWARDS */
+
+.rewards-wrapper {
+  border-top: 1px solid var(--color-border);
+  margin: 18px 0 0 0;
+}
+
+.rewards-title {
+  margin: 6px 0 0 0;
```
xdzurman: perhaps this is personal taste, but changing this to margin-top: ... would be nicer

app/public/css/styles.css
```diff
@@ -1830,6 +1903,45 @@ button[data-balloon] {
...
+  position: absolute;
+  width: 11px;
+  height: 15px;
+  right: 0;
+  top: 16px;
+  margin: auto;
+  background-image: url('../assets/ada_symbol_balance.svg');
+  background-repeat: no-repeat;
+  background-size: contain;
+  background-position: center;
```
xdzurman: I know it's implemented like this in other parts of code, but let's avoid putting svgs as background images in the future. Try to change it the same way you changed ledger icons - if it won't be too much trouble. Same goes for other delegations css.

app/public/css/styles.css
```diff
@@ -2308,6 +2435,66 @@ button[data-balloon] {
...
+  right: 40px;
+  bottom: 40px;
+  height: 40px;
+  background-image: linear-gradient(
+    to top,
+    rgba(255, 255, 255, 1),
+    rgba(255, 255, 255, 0)
+  );
+  z-index: 2;
+}
```
xdzurman: image
whats this for?

app/frontend/components/pages/delegations/currentDelegationPage.js
```diff
@@ -0,0 +1,102 @@
...
+        },
+        'Your Ada is undelegated. Delegate now!'
+      )
+      : h(
+        'div',
+        {class: 'stakepool-info'},
+        h(
+          'div',
+          {class: 'stakepool-name-wrapper'},
+          h('p', {class: 'stakepool-name'}, `${currentStakepool.name}`),
```
xdzurman: Since we're doing camelCase, please change "currentStakepool" to "currentStakePool" since it's not a single word. Same goes for every occurence of "stakepool" -> "stakePool" in the code.

app/frontend/components/pages/delegations/currentDelegationPage.js
```diff
@@ -0,0 +1,102 @@
...
+              },
+              'Seiza'
+            ),
+            h('span', {}, ' | '),
+            h(
+              'a',
+              {
+                class: 'address-link',
+                href: `https://adascan.net/address/${currentStakepool.address}`,
+                style: 'margin-right:24px',
```
xdzurman: whoa didn't know we had that in adalite code, it's generally a bad idea to have styles out of styles file. Please try to put this in respective css class, perhaps do it as well for the code you were inspired by.

app/frontend/components/pages/delegations/currentDelegationPage.js
markopuzav: Fixed.

app/frontend/components/pages/delegations/delegatePage.js.rej
markopuzav: I think this arose when I changed the base branch. Deleted.

app/public/css/styles.css
markopuzav: - Does the multiples-of-4 convention apply to font-size? There are multiple instances of it not being the case.

- I changed the padding of the delegate/undelegate buttons to `6px 10px` to match the donations buttons. Open to discussion as they're not multiples of 4.

xdzurman: font size - not necessarily, regarding that, we try to not have too many different font sizes, so that its similar across the page

app/public/css/styles.css
markopuzav: I used this to set the other margins to 0 (it wouldn't be the case otherwise).

app/frontend/components/pages/delegations/currentDelegationPage.js
markopuzav: I had my doubts indeed. Fixed.

app/frontend/components/pages/delegations/currentDelegationPage.js
markopuzav: Came from `addressItem.js`. Fixed both.

app/public/css/styles.css
markopuzav: Comes from the Transaction History card. Seems to be a design choice to have some whitespace at the bottom of the list.

xdzurman: image
you're right. It just seemed weird to me that it covers a little bit of scrollbar which threw me off a little bit.

markopuzav: Yes, it seems to be overflowing by a few pixels in all of the lists. Especially visible when you open the page in high resolution.

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: Tooltip added.

app/frontend/actions.js
markopuzav: Done.

app/public/css/styles.css
markopuzav: Moved both Balance and Rewards ADA icons from pseudo-element backgrounds into svgs.

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: This is no longer relevant. We will be reverting back to multiple stake pools.

app/frontend/actions.js
```diff
@@ -122,10 +155,14 @@ module.exports = ({setState, getState}) => {
       const walletIsLoaded = true
       const ownAddressesWithMeta = await wallet.getFilteredVisibleAddressesWithMeta()
       const transactionHistory = await wallet.getHistory()
+      const delegationHistory = await getDelegationHistory()
```
ppershing: I guess these methods should be under `wallet`

app/frontend/actions.js
```diff
@@ -109,6 +109,39 @@ module.exports = ({setState, getState}) => {
...
+  // DUMMY
+  const getDelegationHistory = () => {
+    return [
+      {
+        ctbTimeIssued: '06/12/2019, 11:01:22',
+        stakePoolName: 'Stake Pool Uno',
+        stakePoolAddress: 'stake-pool-numero-uno-address',
+      },
+      {
+        ctbTimeIssued: '07/12/2019, 11:01:23',
```
ppershing: let's don't use `ctb` prefix. After all, we will control backend anyway

app/frontend/actions.js
```diff
@@ -251,9 +292,9 @@ module.exports = ({setState, getState}) => {
 
   const openAddressDetail = (state, {address, bip32path}, copyOnClick) => {
     /*
-    * because we don't want to trigger trezor address
-    * verification for the  donation address
-    */
+     * because we don't want to trigger trezor address
```
ppershing: let's extract this into separate commit which we should add on top of both master and develop

app/frontend/components/pages/receiveAda/addressItem.js
```diff
@@ -68,9 +68,8 @@ class AddressItem extends Component {
           h(
             'a',
             {
-              class: 'address-link',
+              class: 'address-link seiza',
```
ppershing: similarly, let's do this (I believe) unrelated change in a separate PR

app/frontend/actions.js
markopuzav: Moved.

app/frontend/actions.js
```diff
@@ -122,10 +169,26 @@ module.exports = ({setState, getState}) => {
...
+          percent: 100,
+          ...fetchPoolInfo(ADALITE_CONFIG.ADALITE_STAKE_POOL_ADDRESS),
+          ID: 0,
+        },
+      ]
+      const delegatedPercent = 100
+      const calculatingDelegationFee = false
+      const delegationFee = 0
+      const isDelegationValid = true
+      const currentDelegation = await wallet.getCurrentDelegation()
```
markopuzav: Not used anywhere so far.

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,159 @@
...
+                min: '0',
+                max: pool.percent + undelegatedPercent,
+                accuracy: '1',
+                class: 'input stake-pool-percent',
+                id: `${pool.ID}`,
+                value: pool.percent,
+                placeholder: pool.percent,
+                onInput: updateStakePoolPercent,
+                autocomplete: 'off',
+              })
```
markopuzav: A non-editable `%` is missing in this input content.

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,159 @@
...
+const formatStakePoolInfo = (className, message) => {
+  return h('div', {class: `stake-pool-info ${className}`}, message)
+}
+
+const CalculatingFee = () => h('div', {class: 'validation-message send'}, 'Calculating fee...')
+
+const getStakePoolValidationMessage = (stakePool, stakePools) => {
+  let className, message
+
+  const addressUsages = stakePools.filter((pool) => pool.address === stakePool.address)
```
ppershing: What is `addressUsages` ?

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,159 @@
...
+  const isDuplicate = addressUsages.length > 1 && stakePool.ID !== addressUsages[0].ID
```
ppershing: I am confused. What are we trying to validate here? Importantly, why are we validating against only [0] ?

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,159 @@
...
+      message = `${stakePool.name} | Delegate a non-zero amount`
+    } else {
+      className = 'valid'
+      message = `${stakePool.name}`
+    }
+  } else {
+    className = 'invalid'
+    message = stakePool.address === '' ? '' : 'Invalid address'
+  }
+  return [className, message]
```
ppershing: Why don't we return object instead of an array? That way it would simplify code a lot

app/frontend/components/pages/delegations/delegatePage.js
```diff
@@ -0,0 +1,159 @@
...
+
+const CalculatingFee = () => h('div', {class: 'validation-message send'}, 'Calculating fee...')
+
+const getStakePoolValidationMessage = (stakePool, stakePools) => {
+  let className, message
+
+  const addressUsages = stakePools.filter((pool) => pool.address === stakePool.address)
+  const isDuplicate = addressUsages.length > 1 && stakePool.ID !== addressUsages[0].ID
+
+  if (stakePool.valid) {
```
ppershing: hmm, from where do we get `.valid` ?

app/frontend/actions.js
```diff
@@ -864,6 +937,70 @@ module.exports = ({setState, getState}) => {
     validateSendFormAndCalculateFee()
   }
 
+  const updateStakePoolAddress = (state, e) => {
+    setState({
+      stakePools: state.stakePools.map((pool) => {
+        return Object.assign(
+          {},
+          pool,
+          pool.ID === parseInt(e.target.id, 10)
```
ppershing: parseInt for pool.ID? they are hex values

app/frontend/actions.js
ppershing: Then we should use it somehow

app/frontend/actions.js
```diff
@@ -122,10 +169,26 @@ module.exports = ({setState, getState}) => {
...
       const conversionRates = getConversionRates(state)
       const sendAmount = {fieldValue: '', coins: 0}
       const sendAddress = {fieldValue: ''}
+      const stakePoolID = 0
+      const stakePools = [
+        {
+          address: ADALITE_CONFIG.ADALITE_STAKE_POOL_ADDRESS,
+          percent: 100,
+          ...fetchPoolInfo(ADALITE_CONFIG.ADALITE_STAKE_POOL_ADDRESS),
+          ID: 0,
```
ppershing: stakepool IDs are (hex-encoded)hashes!

app/frontend/actions.js
```diff
@@ -122,10 +169,26 @@ module.exports = ({setState, getState}) => {
...
+      const delegationHistory = await wallet.getDelegationHistory()
       const balance = await wallet.getBalance()
+      const rewards = await wallet.getRewards()
       const conversionRates = getConversionRates(state)
       const sendAmount = {fieldValue: '', coins: 0}
       const sendAddress = {fieldValue: ''}
+      const stakePoolID = 0
+      const stakePools = [
+        {
+          address: ADALITE_CONFIG.ADALITE_STAKE_POOL_ADDRESS,
```
ppershing: what is STAKE_POOL_ADDRESS? Stake pools *do not* have addresses, only IDs

app/frontend/actions.js
```diff
@@ -864,6 +937,70 @@ module.exports = ({setState, getState}) => {
...
+    validateDelegationAndCalculateDelegationFee()
+  }
+
+  const removeStakePool = (state, e) => {
+    const ID = parseInt(e.target.id, 10)
+    setState({
+      delegatedPercent: state.stakePools
+        .map((pool) => (pool.ID === ID ? 0 : pool.percent))
+        .reduce((x, y) => x + y, 0),
+      stakePools: state.stakePools.filter((pool) => {
```
ppershing: no need for return, could be simple lamda

app/frontend/actions.js
```diff
@@ -864,6 +937,70 @@ module.exports = ({setState, getState}) => {
...
+      stakePoolID: state.stakePoolID + 1,
+    })
+    validateDelegationAndCalculateDelegationFee()
+  }
+
+  const removeStakePool = (state, e) => {
+    const ID = parseInt(e.target.id, 10)
+    setState({
+      delegatedPercent: state.stakePools
+        .map((pool) => (pool.ID === ID ? 0 : pool.percent))
```
ppershing: why would we do it in such a weird way. Actually, we do we even store `delegatedPercent` in state at all if we can calculate it?

app/frontend/actions.js
```diff
@@ -864,6 +937,70 @@ module.exports = ({setState, getState}) => {
...
+  }
+
+  const addStakePool = (state, e) => {
+    setState({
+      stakePools: state.stakePools.concat({
+        address: '',
+        percent: 0,
+        name: '',
+        valid: false,
+        ID: state.stakePoolID + 1,
```
ppershing: This is definitely not what we want. You don't need to store index. Just get it when iterating: `stakePools.map((data, i) => ....)`

app/frontend/actions.js
```diff
@@ -864,6 +937,70 @@ module.exports = ({setState, getState}) => {
...
+              ...fetchPoolInfo(e.target.value),
+            }
+            : {}
+        )
+      }),
+    })
+    validateDelegationAndCalculateDelegationFee()
+  }
+
+  const updateStakePoolPercent = (state, e) => {
```
ppershing: This is very obscure way of updating this value when it can be easily recomputed from the ground data. As I mentioned somewhere else, we actually don't need the `delegatedPercent` value at all

app/frontend/wallet/cardano-wallet.js
```diff
@@ -211,6 +211,87 @@ const CardanoWallet = async (options) => {
...
+
+  // DUMMY
+  function getDelegationHistory() {
+    return [
+      {
+        timeIssued: '06/12/2019, 11:01:22',
+        stakePools: [
+          {
+            name: 'Stake Pool Uno',
+            address: 'stake-pool-numero-uno-address',
```
ppershing: as mentioned elsewhere, stakePools do not have "address", just id

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: Changed.

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: It contains the stake pools with the same ID (what I mistakenly called "address") as `stakePool` in the argument. This is here to check whether the user has typed a stake pool multiple times.

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: If the user has typed a stake pool multiple times, I would like to display a warning message to the duplicates (that is, except the first occurrence, with index `[0]`)

app/frontend/components/pages/delegations/delegatePage.js
markopuzav: Only a dummy so far.

app/frontend/actions.js
markopuzav: fixed

app/frontend/actions.js
markopuzav: Removed the `delegatedPercent` value.

app/frontend/actions.js
markopuzav: This was rubbish indeed, I wasn't aware there exists indexed iteration. Fixed and thanks.

app/frontend/actions.js
markopuzav: Removed.

app/frontend/actions.js
markopuzav: Renamed in all occurrences.

ppershing:
No longer needed as @PeterBenc is working on extending the code in another branch

PR #536, opened 2019-11-27
Converting CommonJS to ES6 Imports

ppershing:
Note: mostly done by clicking on vscode's "fix" feature + some small adjustments when it did something wrong

Note: Needs thorough testing as there is potential for some missed import problems

### DIFF
--- a/app/frontend/components/router.js
+++ b/app/frontend/components/router.js
@@ -1,12 +1,12 @@
-const {h} = require('preact')
-const connect = require('unistore/preact').connect
+import {h} from 'preact'
+import {connect} from 'unistore/preact'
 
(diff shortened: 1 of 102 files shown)


app/frontend/components/router.js
```diff
@@ -44,6 +44,4 @@ const TopLevelRouter = connect((state) => ({
   return content
 })
 
-module.exports = {
-  TopLevelRouter,
-}
```
xdzurman: Let's keep it consistent and export this as default, so that places it's imported in don't have to specify extra {}

ppershing: I'll do this one as it clearly exports a component but I am reluctant to do others as there it is not so clear what is being exported.

PR #541, opened 2019-12-02
Convert project to typescript

ppershing:
Converts the project to typescript & preact calls to jsx syntax


app/frontend/wallet/helpers/request.ts
```diff
@@ -14,8 +14,8 @@ const request = async function request(url, method = 'GET', body = null, headers
   }
 
   try {
-    const response = await fetch(url, requestParams)
-    if (!response) throw NamedError('NetworkError')
+    const response = await fetch(url, (requestParams as any)) // TS does not like credentials
```
ppershing: Figure out if we can convince TS that requestParams credentials is ok. (I suspect type is lost on object.assign above). Also, when we are at it, we can just assing to `requestParams` there directly,

app/frontend/wallet/helpers/request.ts
```diff
@@ -14,8 +14,8 @@ const request = async function request(url, method = 'GET', body = null, headers
...
+    if (!response) throw NamedError('NetworkError','')
```
ppershing: TODO: Figure out if the second parameters goes to UI and if yes, if we don't want some more details there.

app/frontend/wallet/cardano-wallet-secret-crypto-provider.ts
```diff
@@ -52,6 +52,7 @@ const CardanoWalletSecretCryptoProvider = (
     const result = derivePrivate(hdNode.toBuffer(), childIndex, state.derivationScheme.number)
 
     return HdNode({
+      secret: null,
```
ppershing: Instead of this hack, we should have `HdNodeFromSecret` or something like that. Also, `secret` is probably wrong parameter name, if I recall the code, this should be more like `compact` (maybe `HdNodeFromCompactedForm` or something like that?)

app/frontend/libs/unistore/preact.d.ts
```diff
@@ -0,0 +1 @@
+export function connect(mapStateToProps: any, actions?: any): any
```
ppershing: get better types, maybe even update unistore? https://github.com/developit/unistore/blob/master/preact.d.ts

app/frontend/helpers/printConversionRates.ts
```diff
@@ -4,7 +4,7 @@ const addThousandsCommas = (amount, fractionDigits) =>
     maximumFractionDigits: fractionDigits,
   })
 
-const formatFiat = (amount, maxDigits) => {
+const formatFiat = (amount, maxDigits=10) => {
```
ppershing: Find out where (and if all) we used maxDigits and what should be correct default!

app/frontend/helpers/NamedError.ts
```diff
@@ -1,8 +1,8 @@
-function NamedError(name, message, showHelp) {
+function NamedError(name, message, showHelp='') {
```
ppershing: all in all, this should probably be rewritten into class form

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -36,8 +36,40 @@ const SendValidation = ({sendFormValidationError, sendResponse}) =>
       sendResponse.success &&
       h('div', {class: 'validation-message transaction-success'}, 'Transaction successful!')
 
-class SendAdaPage extends Component {
-  // Needs to be a component because we keep refs
+
+interface Props {
```
ppershing: Looking at this component, I suspect it has too much responsibility! And even if not, maybe we should probably structure some props as a single object

app/frontend/components/pages/sendAda/donationButtons.tsx
```diff
@@ -63,7 +77,7 @@ class DonationButtons extends Component {
           ...tooltip('Insufficient funds', isFixedInsufficient),
         },
         `${ADALITE_FIXED_DONATION_VALUE} `,
-        h(AdaIcon)
+        h(AdaIcon, {className: ''})
```
ppershing: Instead, fix the default on AdaIcon

app/frontend/components/pages/login/logoutNotification.tsx
```diff
@@ -26,7 +28,7 @@ class LogoutNotification {
             class: 'button primary',
             onClick: () => setLogoutNotificationOpen(false),
             onKeyDown: (e) => {
-              ;['Enter', 'Escape'].includes(e.key) && e.target.click()
+              ;['Enter', 'Escape'].includes(e.key) && (e.target as HTMLButtonElement).click()
```
ppershing: Research what type is `e` instead.

app/frontend/components/pages/login/logoutNotification.tsx
```diff
@@ -4,6 +4,8 @@ import actions from '../../../actions'
 import Modal from '../../common/modal'
 
 class LogoutNotification {
+  understandDemoBtn: any
```
ppershing: HTMLButtonElement.

app/frontend/helpers/NamedError.ts
PeterBenc: btw. the `showHelp` argument is completely useless, just a forgotten piece code from previous version, we can remove it.

app/frontend/components/autoLogout.ts
```diff
@@ -5,7 +5,15 @@ import {ADALITE_CONFIG} from '../config'
 
 const {ADALITE_LOGOUT_AFTER} = ADALITE_CONFIG
 
-class AutoLogout extends Component {
+interface Props {
+  logout: () => void
+  setLogoutNotificationOpen: (no_idea_what: boolean) => void
```
ppershing: @xdzurman any idea?

app/frontend/components/common/addressDetailDialog.tsx
```diff
@@ -1,12 +1,22 @@
...
 import {connect} from '../../libs/unistore/preact'
 import actions from '../../actions'
 
 import Modal from './modal'
 import CopyOnClick from './copyOnClick'
 
-class AddressDetailDialogClass extends Component {
+interface Props {
+  showDetail: any,
+  closeAddressDetail: (state: any) => void,
```
ppershing: why this takes stake?

app/frontend/components/common/conversions.tsx
```diff
@@ -1,7 +1,12 @@
 import {h} from 'preact'
 import {printConversionRate} from '../../helpers/printConversionRates'
 
-const Conversions = ({balance, conversionRates}) =>
+interface Props {
+  balance: number,
+  conversionRates: any,
```
ppershing: don't we know better type here?

app/frontend/components/common/copyOnClick.tsx
```diff
@@ -1,9 +1,18 @@
-import {h, Component} from 'preact'
+import {h, Component, VNode} from 'preact'
 
 import debugLog from '../../helpers/debugLog'
 import tooltip from './tooltip'
 
-class CopyOnClick extends Component {
+interface Props {
+  value: any
```
ppershing: value shoud be string, no?

app/frontend/components/common/modal.tsx
```diff
@@ -2,7 +2,15 @@
 import {h, Component} from 'preact'
 import Tag from './tag'
 
-class Modal extends Component {
+interface Props{
+  closeHandler: () => void
+  children?: any
```
ppershing: VNode[]

app/frontend/components/common/conversions.tsx
PeterBenc:
```
interface ConversionRates {
  USD: number,
  EUR: number,
}

PR #547, opened 2019-12-12
Convert actions and type state

ppershing:


jmuravsky:
maybe remake the `any`s to a custom TODO type? so that its easier in the future to catch them.

PR #549, opened 2019-12-13
Refactor some global state to hooks

ppershing:


app/frontend/components/common/unexpectedErrorModal.tsx
```diff
@@ -1,84 +1,81 @@
...
+  const [userComments, setComments] = useState('')
 
-  closeAndResolve(resolveValue) {
-    this.props.sendSentry.resolve(resolveValue)
-    this.props.closeUnexpectedErrorModal()
-  }
+  const closeAndResolve = useCallback(
+    (shouldSend: boolean) => {
+      // first tell sentry to send the event
+      sendSentry.resolve(shouldSend) // TODO(ppershing): is this a promise or what?
```
xdzurman: @PeterBenc , can you elaborate?

PeterBenc: yes, it resolves a promise, the promise is pending in `beforeSend` function in `Sentry.init`. Whether the event is send to setry or not is based on the value of the resolved promise.

PeterBenc: in walletApp.js
```
beforeSend(event) {
    return new Promise((resolve) => {
      store.setState({
        sendSentry: {
          event,
          resolve,
        },
        showUnexpectedErrorModal: true,
      })
    }).then((res) => {
      return res === true ? event : null
    })
  },

PR #555, opened 2019-12-14
Refactor - put tx summary into single substate

ppershing:


app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -225,7 +221,7 @@ class SendAdaPage extends Component<Props> {
           />
         )}
         {rawTransactionOpen && <RawTransactionModal />}
-        {showConfirmTransactionDialog && <ConfirmTransactionDialog total={total} />}
+        {showConfirmTransactionDialog && <ConfirmTransactionDialog />}
```
xdzurman: why recalculate total in ConfirmTransactionDialog?

ppershing: Because it should be it's responsibility to correctly calculate the value, especially given the fact that we do not supply any other prop to it.

PR #560, opened 2019-12-15
Refactor - slightly cleanup mnemonic auth

ppershing:


app/frontend/actions.ts
```diff
@@ -220,21 +220,17 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
     const mnemonicInputValue = e.target.value
     setState({
       mnemonicInputValue,
-      showMnemonicValidationError: false,
+      mnemonicValidationError: null,
     })
-    handleError(
-      'mnemonicValidationError',
-      await mnemonicValidator(sanitizeMnemonic(mnemonicInputValue))
-    )
```
xdzurman: @PeterBenc implemented handleError function for unifying error handling, which sets the state for specified error text. Nearly all errors are handled this way. Did you find it this part of code messy or why change it?

app/frontend/components/pages/login/mnemonicAuth.tsx
```diff
@@ -139,10 +140,10 @@ class LoadByMnemonicSectionClass extends Component<Props> {
 export default connect(
   (state) => ({
     mnemonicInputValue: state.mnemonicInputValue,
+    mnemonicValidationError: state.mnemonicValidationError,
+    //
```
xdzurman: //

app/frontend/components/pages/login/mnemonicAuth.tsx
```diff
@@ -8,15 +8,17 @@ import tooltip from '../../common/tooltip'
...
+
 const {ADALITE_DEMO_WALLET_MNEMONIC} = ADALITE_CONFIG
 
 interface Props {
   mnemonicInputValue: string
   mnemonicValidationError: any
   updateMnemonic: (e: any) => void
-  checkForMnemonicValidationError: () => void
+  updateMnemonicValidationError: () => void
+  //
```
xdzurman: did you forget this?

app/frontend/components/pages/login/mnemonicAuth.tsx
```diff
@@ -81,7 +83,7 @@ class LoadByMnemonicSectionClass extends Component<Props> {
           placeholder="Enter your wallet mnemonic"
           value={mnemonicInputValue}
           onInput={updateMnemonic}
-          onBlur={checkForMnemonicValidationError}
+          onBlur={updateMnemonicValidationError}
```
xdzurman: is sanitizedMnemonic validated in two different places now thanks to 'canSubmit'?

app/frontend/components/pages/login/mnemonicAuth.tsx
ppershing: nope, it is there to avoid prettier collapsing things :-)

app/frontend/components/pages/login/mnemonicAuth.tsx
ppershing: same as above, make sure prettier does not collapse things

app/frontend/components/pages/login/mnemonicAuth.tsx
ppershing: Yes. I am open for discussion here but basically, rationale is
- updateValidation() is called whenever we need to display validation things to user
- real validation state can be different and should always be actual.

When thinking about it, maybe I can keep explicit canSubmit flag in state as well. But this would need re-writing the state shape into substate...

app/frontend/actions.ts
ppershing: I am strongly against mis-using "handleError" as means of validation, given its name it should handle a real errors (e.g., failed to submit tx, internal error, etc.). [I guess we can change the name though. However, I am not happy it puts everything into top-level state.]
Speaking of which, `handleError` is terrible name for something which does not handle anything, it should be `setError` in the first place

PeterBenc: I defentlly agree with the `setError` name, and I agree putting everything into top level state isnt good idea either, the purpose of the `handleError` was to save some duplicate lines. The situation was that we were putting almost all errors into top level state, the workflow was always the same, set error state and then have it checked whether it's unexpected thus sending it to sentry. So instead of doing this in all cases, you could just call handleError with name and value that should be set.

As I mentioned, `setError` is a lot more appropriate and when it comes to "puts everything into to level state", we can change that and then the `handleError` function wont be nessesary and we can find a different way to always check for unexpected errors.

ppershing: Let's gradually figure out the final design. For now let's move relevant validation "errors" into corresponding substates and handle them in their respective actions. But I would do this only when touching given part of codebase ...

PR #564, opened 2019-12-16
Fix useActions hook

ppershing:
- I did not properly test the lib version and it does not bind actions correctly


ppershing:
Plese try to test it as well

PR #565, opened 2019-12-16
Refactor -- separate address generation from address discovery

ppershing:


app/frontend/wallet/address-manager.ts
```diff
@@ -3,79 +3,91 @@ import {toBip32StringPath} from './helpers/bip32'
...
-        childIndexBegin + gapLimit
-      )
+      const currentAddressBlock = await deriveAddressesBlock(from, from + gapLimit)
+
       isGapBlock = !(await blockchainExplorer.isSomeAddressUsed(currentAddressBlock))
+
       addresses =
         isGapBlock && addresses.length > 0 ? addresses : addresses.concat(currentAddressBlock)
-      childIndexBegin += gapLimit
+      from += gapLimit
```
PeterBenc: Previously, we distinguished between `v1` and `v2` address types and we were checking whether "some addresses are used" only for `v2` addresses. What is the reason for not distinguishing anymore and checking for used address for `v1` addresses as well? just to clarify..

ppershing: Assuming `defaultAddressCount < gapLimit` the new code will anyway discover all old addresses. It might discover more of them but that is not a problem for the UI or the rest of functionality ...

PR #569, opened 2019-12-19
Submit tx and delegate stake

PeterBenc:


server/helpers/loadConfig.js
```diff
@@ -28,8 +28,10 @@ const checkMap = check.map(process.env, {
   ADALITE_SUPPORT_EMAIL: check.nonEmptyString,
   ADALITE_FIXED_DONATION_VALUE: isPositiveIntString,
   ADALITE_MIN_DONATION_VALUE: isPositiveIntString,
+  ADALITE_STAKE_POOL_ID: check.nonEmptyString,
   ADALITE_ENV: check.nonEmptyString,
   SENTRY_DSN: check.nonEmptyString,
+  ADALITE_JORMUNGANDR_NODE_URL: check.nonEmptyString,
```
ppershing: Now that this is going to backend, we shouldn't need this

server/helpers/loadConfig.js
```diff
@@ -28,8 +28,10 @@ const checkMap = check.map(process.env, {
   ADALITE_SUPPORT_EMAIL: check.nonEmptyString,
   ADALITE_FIXED_DONATION_VALUE: isPositiveIntString,
   ADALITE_MIN_DONATION_VALUE: isPositiveIntString,
+  ADALITE_STAKE_POOL_ID: check.nonEmptyString,
```
ppershing: FYI, it's going to be two pools

ppershing: Maybe we should change this to "defaultStakingPool".

package.json
```diff
@@ -23,14 +23,16 @@
   "scripts": {
     "start-server": "node server/index.js",
     "build": "cd app && yarn install && yarn build",
+    "install": "cd app && yarn install && cd .. && yarn remove-wasm-import",
```
ppershing: let's call it `fix-chain-libs`

app/frontend/wallet/delegation/addresses.js
```diff
@@ -0,0 +1,69 @@
+import {
```
ppershing: this file can be removed, it is just an example usage of chain-libs

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const changeDelegation = async (state) => {
+    console.log('changing delegation')
+  }
+
+  const reloadWalletInfo = () => {
+    return null
+  }
+
+  const calculateDelegatedPercent = (pools) => {
```
ppershing: hmm, why is this in actions?

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+    })
+    setState({
+      calculatingDelegationFee: false,
+    })
+  }
+
+  const debouncedCalculateDelegationFee = debounceEvent(calculateDelegationFee, 2000)
+
+  // DUMMY
+  const validateDelegationAndCalculateDelegationFee = () => {
```
ppershing: state should be first argument, no?

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  const updateStakePoolId = (state, e) => {
+    const poolId = e.target.value
+    setState({
+      stakePools: state.stakePools.map((pool, i) => {
+        return Object.assign(
+          {},
+          pool,
+          i === parseInt(e.target.name, 10)
```
ppershing: extract `parseInt` to some constant.

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      setState({ delegationFee: 0 })
+    }
+  }
+
+  const updateStakePoolId = (state, e) => {
+    const poolId = e.target.value
+    setState({
+      stakePools: state.stakePools.map((pool, i) => {
+        return Object.assign(
```
ppershing: Why do we need object assign at all if `simple .map(i == x ? orig : new)` would suffice?

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+    validateDelegationAndCalculateDelegationFee()
+  }
+
+  const updateStakePoolPercent = (state, e) => {
+    const index = parseInt(e.target.name, 10)
+    const delegatedPercent = calculateDelegatedPercent(state.stakePools)
+    const oldPercent = state.stakePools
+      .map((pool, i) => (i === index ? pool.percent : 0))
+      .reduce((x, y) => x + y, 0)
+    const nextPercent = Math.min(
```
ppershing: What is going on here? Why do we do `math.min` at all and just not set the value from the event?

app/frontend/testnet/testnet-actions.ts
```diff
@@ -0,0 +1,167 @@
...
+      stakePools: state.stakePools.map((pool, i) => {
+        return Object.assign({}, pool, i === index ? { percent: nextPercent } : {})
+      }),
+    })
+    validateDelegationAndCalculateDelegationFee()
+  }
+
+  const addStakePool = (state, e) => {
+    setState({
+      stakePools: state.stakePools.concat({
```
ppershing: we should prefer `[...stakePools, {}]` as this makes it more explicit the function does not mutate anything

ppershing: Also, this should probably be extracted into `defaultPoolInfo` or something like that

app/frontend/testnet/libs/chainlib-wrapper.js
```diff
@@ -0,0 +1,319 @@
+// import * as W from "js-chain-libs/pkg/js_chain_libs_bg.wasm";
```
ppershing: So is this the real wrapper and the two new files later in this PR should be deleted?

app/frontend/testnet/libs/bech32.js
```diff
@@ -0,0 +1,32 @@
+import bech32 from "bech32";
+
+// Note(ppershing): yeah, really, bech32 lib has this stupid api
+export const bech32_encode = (prefix, data) => {
+  const words = bech32.toWords(data);
+  return bech32.encode(prefix, words, 1000);
```
ppershing: add a note that we need longer than default length for privkeys and 1000 should suffice

app/frontend/testnet/libs/bech32.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+};
+
+export const buf2hex = arrayBuffer => {
+  return Array.from(new Uint8Array(arrayBuffer))
+    .map(n => n.toString(16).padStart(2, "0"))
+    .join("");
+};
+
+export const parse_bech32 = str => {
```
ppershing: this should not be needed after we debug all problems, only `bech32_decode` should be used

app/frontend/state.ts
```diff
@@ -90,6 +90,17 @@ export interface State {
   donationAmountValidationError?: any
   sendAddressValidationError?: any
   transactionSubmissionError?: any
+  stakePools: any
```
ppershing: we need to separate this into relevant substates

app/frontend/testnet/libs/chainlib-wrapper.js
```diff
@@ -0,0 +1,225 @@
...
+  TransactionSignDataHash,
+} = lib
+
+const hex2buf = (str) => Buffer(str, 'hex')
+
+export const calculate_fee = ({input_count, output_count, cert_count, chain_config}) => {
+  // Fee (#inputs + #outputs) * coefficient + #certificates*certificate + constant
+  const fees = chain_config.fees
+
+  return (
```
SebastienGllmt: FYI this is wrong. Jormungandr has a special fee depending on the kind of certificate in the transaction. In Yoroi we have a hack so that it works for StakeDelegation certificates but I will make a proper fix for it this weekend. This is also why in Yoroi -- instead of hard-coding the fee like this -- we calculate the fee by building the transaction in Rust and calling the WASM function to get the fee

ppershing: This is interesting. Especially given the fact that chain settings jormungandr returns mention only the three coefficients.

PR #577, opened 2019-12-21
Refactor - separate mnemonic/crypto mode from address path derivation

ppershing:


app/tests/src/address-manager.js
```diff
@@ -67,7 +66,7 @@ const initAddressManager = async (settings, i) => {
   )
 
   addressManagers[i] = AddressManager({
-    accountIndex: HARDENED_THRESHOLD,
+    accountIndex: 0,
```
xdzurman: whats the reason behind this?

ppershing:
Because you anyway refer to HD+x as x', accounts are always hardened and
numbering accounts from zero therefore makes sense.

Issue #608, opened 2020-05-24
Misleading error message when adalite backend is down


refi93:
While our nodes are down, the users see the following message which is misleading: `Error loading wallet `. and bellow this message: `NetworkError: nection failed. Please check your network connection.`

Proposed solution - expand the loading of the wallet with a call to explorer3.adalite.io/api/v2/healthcheck and if the service is unhealthy, write something more specific like "we are having issues with our service, please try again later" or something like that which would not be misleading the users into thinking there are any network issues on their side

We may also think of serving the data from the backend even if it is not syncing and just displaying an overlay that we are having issues and the wallet is currently available only in read-only mode (tx submission disabled since we would not be able to show updated balance to the user) with potentially outdated information.

PR #610, opened 2020-05-25
I 608 fix misleading error message when backend is down

PeterBenc:
Partly addresses #608, just fixing the error message so its more general and includes a case when Adalite backend is down.

### DIFF
--- a/app/frontend/actions.js
+++ b/app/frontend/actions.js
@@ -583,10 +583,11 @@ export default ({setState, getState}) => {
 
   const calculateMaxDonationAmount = async () => {
     const state = getState()
-    const maxDonationAmount = await wallet.getMaxDonationAmount(
-      state.sendAddress.fieldValue,
-      state.sendAmount.coins
-    )
+    const maxDonationAmount = await wallet
+      .getMaxDonationAmount(state.sendAddress.fieldValue, state.sendAmount.coins)
+      .catch((e) => {
+        handleError('sendAmountValidationError', {code: e.name})
+      })
     let newMaxDonationAmount
     if (maxDonationAmount >= toCoins(ADALITE_CONFIG.ADALITE_MIN_DONATION_VALUE)) {
       newMaxDonationAmount = maxDonationAmount
@@ -645,13 +646,16 @@ export default ({setState, getState}) => {
 
   const sendMaxFunds = async (state) => {
     setState({calculatingFee: true})
-    const maxAmounts = await wallet
+    await wallet
       .getMaxSendableAmount(
         state.sendAddress.fieldValue,
         state.donationAmount.fieldValue !== '',
         state.donationAmount.coins,
         state.checkedDonationType
       )
+      .then((maxAmounts) => {
+        validateAndSetMaxFunds(state, maxAmounts)
+      })
       .catch((e) => {
         setState({
           calculatingFee: false,
@@ -659,7 +663,6 @@ export default ({setState, getState}) => {
         handleError('sendAmountValidationError', {code: e.name})
         return
       })
-    validateAndSetMaxFunds(state, maxAmounts)
   }
 
   const waitForTxToAppearOnBlockchain = async (state, txHash, pollingInterval, maxRetries) => {
--- a/app/frontend/translations.js
+++ b/app/frontend/translations.js
@@ -39,9 +39,12 @@ const translations = {
   TrezorError: ({message}) => `TrezorError: ${message}`,
   LedgerOperationError: ({message}) => `LedgerOperationError: ${message}`,
 
-  CoinAmountError: () => 'CoinAmounError: Unsupported amount of coins.',
+  CoinAmountError: () => 'CoinAmountError: Unsupported amount of coins.',
   CryptoProviderError: ({message}) => `CryptoProviderError: ${message}`,
-  NetworkError: () => 'NetworkError: nection failed. Please check your network connection.',
+  NetworkError: () =>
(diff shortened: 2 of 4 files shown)


app/frontend/translations.js
```diff
@@ -41,7 +41,8 @@ const translations = {
 
   CoinAmountError: () => 'CoinAmounError: Unsupported amount of coins.',
   CryptoProviderError: ({message}) => `CryptoProviderError: ${message}`,
-  NetworkError: () => 'NetworkError: nection failed. Please check your network connection.',
+  NetworkError: () =>
```
refi93, line 44: we can do better by going back to the "request" helper: https://github.com/vacuumlabs/adalite/blob/master/app/frontend/wallet/helpers/request.js#L22 and making the returned error more granular. As you can see, we are currently returning "Network Error" regardless of the response code, but we could refactor it for example like this:

```
const request = async function request(url, method = 'GET', body = null, headers = {}) {
  let requestParams = {
    method,
    headers,
    credentials: 'include',
  }
  if (method.toUpperCase() !== 'GET') {
    requestParams = Object.assign({}, requestParams, {body})
  }

  let response = null
  try {
    response = await fetch(url, requestParams)
  } catch (e) {
    throw NamedError('NetworkError', `${method} ${url} has failed with the following error: ${e}`)
  }

  if (!response) throw NamedError('NetworkError', `No response from ${method} ${url}`)
  
  if (response.status === 429) {
    await sleep(DELAY_AFTER_TOO_MANY_REQUESTS)
    return await request(url, method, body, headers)
  }
  
  if (response.status >= 500) {
    throw NamedError(
      'ServerError',
      `${method} ${url} returns error: ${response.status} on payload: ${JSON.stringify(requestParams)}`
    )
  } else if (response.status >= 400) {
    throw NamedError(
      'NetworkError',
      `${method} ${url} returns error: ${response.status} on payload: ${JSON.stringify(requestParams)}`
    )
  }

  return response.json()
}
```

So we could return a "ServerError" if the server returns an error 500 and then add a more specific message for that. 

The messages could then look like this:
```
NetworkError: () => 'NetworkError: Request to our servers has failed. Please check your network connection and if the problem persists, contact us.',
  ServerError: () => 'ServerError: Our servers are probably down. Please try again later and if the problem persists, contact us.',
```
Which makes clear when our servers are down vs an unknown network issue which I believe makes for a better UX and hopefully less confused users contacting us.

I also noticed we are throwing some NetworkErrors in `cardano-wallet.js` - but those are essentially wrapping network requests that would fail with it anyway, so I would suggest to remove those try/catch blocks completely as they just strip away context from the errors thrown and possibly obfuscate other non-network related errors

app/frontend/translations.js
```diff
@@ -41,7 +41,8 @@ const translations = {
 
   CoinAmountError: () => 'CoinAmounError: Unsupported amount of coins.',
```
refi93, line 42: not realted but easy to fix this typo - should be "CoinAmountError"

app/frontend/wallet/helpers/request.js
```diff
@@ -19,6 +19,13 @@ const request = async function request(url, method = 'GET', body = null, headers
     if (response.status === 429) {
       await sleep(DELAY_AFTER_TOO_MANY_REQUESTS)
       return await request(url, method, body, headers)
+    } else if (response.status >= 500) {
```
refi93, line 22: this would get out of the request function as a generic "NetworkError" because it's re-caught on line 36. I covered that case in the refactor I suggested in the comment, of course, feel free to improve it

refi93: Also, I would remove the try/catch blocks throwing the generic "NetworkError" in cardano-wallet.js, otherwise if those calls fail with a "ServerError" it would not propagate to the "message-rendering" layer and the user would see a misleading error message.

refi93: I see - wouldn't that debuglog be handled by the global error handler https://github.com/vacuumlabs/adalite/blob/5edd51631c1fa1e93d3bbe96b5ba4a82d4f17f1a/app/frontend/actions.ts#L74 though?

PeterBenc: hm, :D yeah it would, thanks, :+1: fixing...

app/frontend/actions.js
```diff
@@ -583,10 +583,11 @@ export default ({setState, getState}) => {
...
     const state = getState()
-    const maxDonationAmount = await wallet.getMaxDonationAmount(
-      state.sendAddress.fieldValue,
-      state.sendAmount.coins
-    )
+    const maxDonationAmount = await wallet
+      .getMaxDonationAmount(state.sendAddress.fieldValue, state.sendAmount.coins)
+      .catch((e) => {
+        handleError('sendAmountValidationError', {code: e.name})
+      })
```
PeterBenc, line 590: When testing I noticed the network error is unhandled in case of calculating max amount and donation amount, so in case a user has network issues or our servers are down during fee calculation, unexpected error modal shows up.

app/frontend/actions.js
```diff
@@ -645,21 +646,23 @@ export default ({setState, getState}) => {
...
+    await wallet
       .getMaxSendableAmount(
         state.sendAddress.fieldValue,
         state.donationAmount.fieldValue !== '',
         state.donationAmount.coins,
         state.checkedDonationType
       )
+      .then((maxAmounts) => {
+        validateAndSetMaxFunds(state, maxAmounts)
+      })
```
PeterBenc, line 658: Same situation as mentioned before, adding a fix to this PR since its related to the network and server error.

PR #604, opened 2020-04-23
Replace validation error message

xdzurman:

PR #615, opened 2020-06-02
Refactor testnet functionality

PeterBenc:
Marked WIP due to some known issues that will be added to this description shortly.

**Commits:** 

- **rename handle error to setErrorState**
renames the handleError function to setErrorState since it makes more sense, the function isn't handling the error rather than setting just setting some states
- **reorder actions.ts file**
reorders the functions in this file to some logical sections, the file is too long anyway so its hard to navigate either way, but helps at least a bit
- **refactor tx plan building**
adds some types to build-transaction file, adds type attribute to tx plan, refactors account plan computing
in shelley-wallet file, refactors plan building functions so separates staking and non-staking utxos, shuffling them separately,
adds types to txPlan arguments
- **refactor fee calculation**
renames `showTxSuccess` to `txSuccessTab` (this variable determines, on which tab is the tx success shown
previously, if a user typed a number of coins he doesn't have, the fee calculation was triggered anyway, this commit fixes this behavior, adds a validation error to validators along with balance argument ( this replaces the hotfix in this PR #604 so it all works as expected)
refactors the fee calculating functions (for normal sending, delegations, delegation revoking and non-staking conversion), these have been unnecessarily long and inconsistent, so it removes unnecessary lines and make them logically follow the same pattern
creates separate functions for setting a transaction plan and preparing the plan, 
also, some values in `sendAdaPage` haven't been resetting properly ( f.e if user cleaned the form, or after tx) so this behavior is also fixed in this commit.
- **remove useless error caching in wallets and refactor**
as pointed by Rafo, previously the networks errors have been generalized in wallets, this commit removes the behavior 
- **add proper server error and fix translations**
refactors request helper the way it is on mainnet and fixes the error translations
- **show network errors in components and fix setErrorState function**
previously there was a bug in `setErrorState` function
all fee calculating function has been slightly redone, to handle the way a possible error in preparing tx plan is shown. Previously we were showing all these errors in `TransactionErrorModal` which doesn't make much sense, this way, if an error occurs during fee calculation, in case of normal tx, or change in donation amount and delegation fee calculation, the error is shown in the `sendAdaPage` or `delegationPage` components, and a modal is shown for converting funds or revoking delegation (since these doesn't have a proper place to show the error)
- **fix modal error overflow and hotfix tool tip width**
again an issue already solved in mainet branch and hot-fixes the min-width attribute for the data-balloon
- **reorder actions.ts file to fix eslint errors**
the file had to be reordered once again, and SEND ADA and DONATION sections merged since its impossible to have them separated and avoid 'used before defined' error

- next free commits should be squashed before merge,
just fixing all the other linting errors and audit errors, 

- **bump yargs-parser from 13.1.1 to 13.1.2 and bump acorn**
bumps

- add real Shelley validation
previously the addresses were validated very simply, this commit fixes this behavior validating them with js-chain-libs

**Issues**
- according to the specification of the Shelley era, there won't be functionality for delegation by ratio, thus no multiple pools, with different ratios. Currently, the codebase is prepared for this functionality making it unnecessary complex, in multiple places.
- throughout the codebase, we use `coins`, `amount` and `value` terms, they should be unified or their difference in meaning should be specified
- the tx success sign (the green one) seems to be not showing in some cases after the tx is submitted
- `updateStakepoolIndetiifer` function should probably be modified, quick typing and deleting seems to be causing bugs

### DIFF
--- a/app/frontend/components/pages/delegations/currentDelegationPage.tsx
+++ b/app/frontend/components/pages/delegations/currentDelegationPage.tsx
@@ -15,10 +15,6 @@ const CurrentDelegationPage = ({
       {currentDelegation.length ? (
         <div>
           <div className="current-delegation-wrapper">
-            {/* <div className="delegation-history-time">
-              {formatDelegationDate(currentDelegation[0].time)}
-            </div>]
-            <div /> */}
             {currentDelegation.map((pool, i) => [
               <div key={i} className="delegation-history-name">
                 {pool.name}
--- a/app/frontend/components/pages/delegations/currentDelegationPage.tsx
+++ b/app/frontend/components/pages/delegations/currentDelegationPage.tsx
@@ -1,4 +1,4 @@
-import {h, Fragment} from 'preact'
+import {h} from 'preact'
 import {connect} from '../../../libs/unistore/preact'
 import actions from '../../../actions'
 import printAda from '../../../helpers/printAda'
@@ -17,21 +17,25 @@ const CurrentDelegationPage = ({
           <div className="current-delegation-wrapper">
             {/* <div className="delegation-history-time">
               {formatDelegationDate(currentDelegation[0].time)}
-            </div>
+            </div>]
             <div /> */}
-            {currentDelegation.map((pool) => [
-              <div className="delegation-history-name">{pool.name}</div>,
-              <div className="delegation-history-percent">{`${pool.ratio} %`}</div>,
-              <div className="delegation-history-id">{pool.pool_id}</div>,
-              <div className="delegation-history-id">{`Ticker: ${pool.ticker}`}</div>,
-              <div className="delegation-history-id">
+            {currentDelegation.map((pool, i) => [
+              <div key={i} className="delegation-history-name">
+                {pool.name}
+              </div>,
+              <div key={i} className="delegation-history-percent">{`${pool.ratio} %`}</div>,
+              <div key={i} className="delegation-history-id">
+                {pool.pool_id}
+              </div>,
+              <div key={i} className="delegation-history-id">{`Ticker: ${pool.ticker}`}</div>,
+              <div key={i} className="delegation-history-id">
                 {`
                 Tax: ${(pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1] || ''}%
                 ${pool.rewards.fixed ? ` , ${`Fixed: ${printAda(pool.rewards.fixed)}`}` : ''}
                 ${pool.rewards.limit ? ` , ${`Limit: ${printAda(pool.rewards.limit)}`}` : ''}
               `}
               </div>,
-              <div className="delegation-history-id">
+              <div key={i} className="delegation-history-id">
                 {'Homepage: '}
                 <a href={pool.homepage}>{pool.homepage}</a>
               </div>,
--- a/app/frontend/components/pages/login/loginPage.tsx
+++ b/app/frontend/components/pages/login/loginPage.tsx
@@ -5,7 +5,7 @@ import isLeftClick from '../../../helpers/isLeftClick'
 
 import KeyFileAuth from './keyFileAuth'
 import MnemonicAuth from './mnemonicAuth'
-import HardwareAuth from './hardwareAuth'
+// import HardwareAuth from './hardwareAuth'
 import DemoWalletWarningDialog from './demoWalletWarningDialog'
 import GenerateMnemonicDialog from './generateMnemonicDialog'
 import LogoutNotification from './logoutNotification'
--- a/app/frontend/components/pages/login/loginPage.tsx
+++ b/app/frontend/components/pages/login/loginPage.tsx
@@ -5,7 +5,6 @@ import isLeftClick from '../../../helpers/isLeftClick'
 
 import KeyFileAuth from './keyFileAuth'
 import MnemonicAuth from './mnemonicAuth'
-// import HardwareAuth from './hardwareAuth'
 import DemoWalletWarningDialog from './demoWalletWarningDialog'
 import GenerateMnemonicDialog from './generateMnemonicDialog'
 import LogoutNotification from './logoutNotification'
@@ -129,11 +128,6 @@ class LoginPage extends Component<Props, {isDropdownOpen: boolean}> {
         <h2 className="authentication-title">How do you want to access your Cardano Wallet?</h2>
         <div className="auth-options">
           {AuthOption('mnemonic', ['12, 15 or 27 word passphrase'], 'fastest')}
-          {/* {AuthOption(
-            'hw-wallet',
-            ['Trezor T', 'Ledger Nano S/X', 'Android device & Ledger'],
-            'recommended'
-          )} */}
           {AuthOption('file', ['Encrypted .JSON file'], '')}
         </div>
       </div>
@@ -149,12 +143,10 @@ class LoginPage extends Component<Props, {isDropdownOpen: boolean}> {
           {CurrentDropdownItem(authMethod)}
           <ul className="dropdown-items">
             {DropdownItem('mnemonic')}
-            {/* {DropdownItem('hw-wallet', true)} */}
             {DropdownItem('file')}
           </ul>
         </div>
         {authMethod === 'mnemonic' && <MnemonicAuth />}
-        {/* {authMethod === 'hw-wallet' && <HardwareAuth loadWallet={loadWallet} />} */}
         {authMethod === 'file' && <KeyFileAuth />}
       </div>
     )
--- a/app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
+++ b/app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
@@ -4,6 +4,7 @@ import actions from '../../../actions'
 import printAda from '../../../helpers/printAda'
 import Modal from '../../common/modal'
 import RawTransactionModal from './rawTransactionModal'
+import roundNumber from '../../../helpers/roundNumber'
 
 interface Props {
   sendAddress: any
@@ -76,7 +77,7 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
                 <div className="review-amount">{pool.ticker}</div>
                 <div className="review-label">Tax</div>
                 <div className="review-amount">
-                  {pool.rewards && (pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
+                  {pool.rewards && roundNumber(pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
                 </div>
                 <div className="review-label">Homepage</div>
                 <div className="review-amount">{pool.homepage}</div>
--- a/app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
+++ b/app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
@@ -31,14 +31,14 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
     setRawTransactionOpen,
     rawTransactionOpen,
     stakePools,
-    txConfirmType
+    txConfirmType,
   }) {
     const total = summary.amount + summary.donation + summary.fee
     const titleMap = {
       delegate: 'Delegation review',
       revoke: 'Delegation revocation review',
       send: 'Transaction review',
-      convert: 'Stakable balance conversion review'
+      convert: 'Stakable balance conversion review',
     }
 
     return (
@@ -67,7 +67,7 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
 
           {txConfirmType === 'delegate' &&
             stakePools.map((pool, i) => (
-              <Fragment>
+              <Fragment key={i}>
                 <div className="review-label">Pool ID</div>
                 <div className="review-amount">{pool.poolIdentifier}</div>
                 <div className="review-label">Pool Name</div>
@@ -76,8 +76,7 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
                 <div className="review-amount">{pool.ticker}</div>
                 <div className="review-label">Tax</div>
                 <div className="review-amount">
-                  {pool.rewards &&
-                    (pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
+                  {pool.rewards && (pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
                 </div>
                 <div className="review-label">Homepage</div>
                 <div className="review-amount">{pool.homepage}</div>
@@ -97,10 +96,10 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
           <a
             className="review-cancel"
             onClick={cancelTransaction}
-            ref={element => {
+            ref={(element) => {
               this.cancelTx = element
             }}
-            onKeyDown={e => {
+            onKeyDown={(e) => {
               e.key === 'Enter' && (e.target as HTMLAnchorElement).click()
             }}
           >
@@ -117,12 +116,12 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
 }
 
 export default connect(
-  state => ({
+  (state) => ({
     sendAddress: state.sendAddress.fieldValue,
     summary: state.sendTransactionSummary,
     rawTransactionOpen: state.rawTransactionOpen,
     stakePools: state.shelleyDelegation.selectedPools,
-    txConfirmType: state.txConfirmType
+    txConfirmType: state.txConfirmType,
   }),
   actions
 )(ConfirmTransactionDialogClass)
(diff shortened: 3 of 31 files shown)


app/frontend/components/pages/delegations/currentDelegationPage.tsx
```diff
@@ -17,21 +17,25 @@ const CurrentDelegationPage = ({
           <div className="current-delegation-wrapper">
             {/* <div className="delegation-history-time">
               {formatDelegationDate(currentDelegation[0].time)}
-            </div>
+            </div>]
```
refi93, line 20: why wrapping the commented section in "{}" and also this "]" does not seem to have a matching "["

app/frontend/components/pages/login/loginPage.tsx
```diff
@@ -5,7 +5,7 @@ import isLeftClick from '../../../helpers/isLeftClick'
 
 import KeyFileAuth from './keyFileAuth'
 import MnemonicAuth from './mnemonicAuth'
-import HardwareAuth from './hardwareAuth'
+// import HardwareAuth from './hardwareAuth'
```
refi93, line 8: I would just remove the line completely, we have it git history anyway and whoever will need to reenable the functionality will need to do probably a lot more than just uncommenting this line

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
```diff
@@ -76,8 +76,7 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
                 <div className="review-amount">{pool.ticker}</div>
                 <div className="review-label">Tax</div>
                 <div className="review-amount">
-                  {pool.rewards &&
-                    (pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
+                  {pool.rewards && (pool.rewards.ratio[0] * 100) / pool.rewards.ratio[1]}%
```
refi93, line 79: if I understand right, this division may produce nasty values, like "66.66666667", shouldn't we wrap some rounding around it?

app/frontend/wallet/byron/cardano-wallet-secret-crypto-provider.ts
```diff
@@ -101,7 +102,7 @@ const CardanoWalletSecretCryptoProvider = ({
   async function signTxGetStructured(txAux, addressToAbsPathMapper) {
     const txHash = txAux.getId()
     const witnesses = await Promise.all(
-      txAux.inputs.map(async (input) => {
+      txAux.inputs.map((input) => {
```
refi93, line 105: I believe the "async" keyword was there to highlight that the function is returning a promise. Since this was refactored to typescript, I would add explicit type annotation to the function, probably `Promise<TxWitness>`

app/frontend/wallet/cardano-wallet.ts
```diff
@@ -10,10 +8,9 @@ import {MAX_INT32} from './constants'
 import shuffleArray from './helpers/shuffleArray'
 import NamedError from '../helpers/NamedError'
 import {Lovelace} from '../state'
-
 import {computeRequiredTxFee, selectMinimalTxPlan, isUtxoProfitable} from './byron-tx-planner'
-
 import {MaxAmountCalculator} from './max-amount-calculator'
+// eslint-disable-next-line no-unused-vars
 import {TxPlan} from './shelley/build-transaction'
```
refi93, line 14: you can import type directly and save this workaround if you update typescript to the latest version as this feature was added in typescript 3.8: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

app/frontend/wallet/cardano-wallet.ts
```diff
@@ -158,12 +133,7 @@ const CardanoWallet = (options) => {
 
   async function submitTx(signedTx) {
     const {txBody, txHash} = signedTx
-    const response = await blockchainExplorer.submitTxRaw(txHash, txBody).catch((e) => {
-      debugLog(e)
-      throw e
-    })
-
-    return response
+    return await blockchainExplorer.submitTxRaw(txHash, txBody)
```
refi93, line 136: I would use type annotation here to make clear the returned value is a promise instead of awaiting on return

app/frontend/wallet/cardano-wallet.ts
```diff
@@ -232,12 +202,12 @@ const CardanoWallet = (options) => {
     return blockchainExplorer.getTxHistory(addresses)
   }
 
-  async function getAccountInfo() {
+  function getAccountInfo() {
```
refi93, line 205: this seems weird - I would put a comment here explaining why we return undefined (possibly a "TODO" comment so it's not forgotten) or rewrite it

app/frontend/wallet/cardano-wallet.ts
```diff
@@ -262,15 +232,11 @@ const CardanoWallet = (options) => {
...
 
   async function getUTxOs(): Promise<Array<UTxO>> {
-    try {
-      const addresses = await myAddresses.discoverAllAddresses()
-      return await blockchainExplorer.fetchUnspentTxOutputs(addresses)
-    } catch (e) {
-      throw NamedError('NetworkError')
-    }
+    const addresses = await myAddresses.discoverAllAddresses()
+    return await blockchainExplorer.fetchUnspentTxOutputs(addresses)
```
refi93, line 236: No need to await on return. It may make sense for readability but I see that there is a return value type annotation already, so it's imho redundant: https://eslint.org/docs/rules/no-return-await

app/frontend/wallet/helpers/request.ts
```diff
@@ -17,54 +12,32 @@ const request = async function request(url, method = 'GET', body = null, headers
...
-  // } catch (e) {
-  //   debugLog(e)
-  //   throw NamedError(
-  //     e.name === 'NodeOutOfSync' ? e.name : 'NetworkError',
-  //     `${method} ${url} returns error: ${e}`
-  //   )
-  // }
+  if (response.status === 429) {
+    await sleep(DELAY_AFTER_TOO_MANY_REQUESTS)
+    return await request(url, method, body, headers)
```
refi93, line 22: no need to await on return

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -200,7 +193,9 @@ const ShelleyBlockchainExplorer = (config) => {
       throw NamedError('NetworkError', e.message)
     }
     const poolArray = JSON.parse(await response.text())
+    // eslint-disable-next-line no-sequences
     const validStakepools = poolArray.reduce((dict, el) => ((dict[el.pool_id] = {...el}), dict), {})
+    // eslint-disable-next-line no-sequences
     const ticker2Id = poolArray.reduce((dict, el) => ((dict[el.ticker] = el.pool_id), dict), {})
     return {validStakepools, ticker2Id}
```
refi93, line 200: personally, I would prefer to return a single array of stake pools and store it like that in the state as well, avoiding having duplicate related variables in it which may accidentally fall out of sync. And then implement helpers to retrieve that information from that array.

Btw, accessing/setting object properties by `[propName]` may lead to weird glitches, if some stake pool happens to be named like a JS object prototype property, imagine this code:

```
let x = {"a": 3}
console.log(x["constructor") // outputs a function()
```

app/frontend/helpers/validators.ts
```diff
@@ -94,7 +108,7 @@ const poolIdValidator = (poolId, selectedPools, validStakepools) => {
     }
   }
   const selectedPoolsIds = selectedPools.map((pool) => pool.pool_id)
-  if (selectedPoolsIds.includes(poolId)) {
+  if (selectedPoolsIds.every((id, index) => poolId === id && index !== poolIndex)) {
     return {
       code: 'RudundantStakePool',
```
refi93, line 113: typo -> "Rudundant"

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -295,61 +288,75 @@ const ShelleyWallet = ({config, randomInputSeed, randomChangeSeed, cryptoProvide
...
-    const {address, coins, donationAmount, pools} = args
-    const utxoFilters = {
-      delegation: isShelleyUtxo,
-      nonStakingConversion: isUtxoNonStaking,
-      utxo: () => true,
-    }
-    const utxoFilter = utxoFilters[txType]
-    const accountAddress = await myAddresses.accountAddrManager._deriveAddress(0)
-    const availableUtxos = (await getUTxOs()).filter(utxoFilter)
+  type utXoArgs = {
```
refi93, line 295: I find those alternating lower and upper case letters in `utxo` in the variable name confusing and it differs a few lines below in `uTxOTxPlanner` function name. I would rather coin the `utxo` as a single name and not try to match the words that it stands for, because it complicates camel casing as can be seen here

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -295,61 +288,75 @@ const ShelleyWallet = ({config, randomInputSeed, randomChangeSeed, cryptoProvide
...
-    // we do it pseudorandomly to guarantee fee computation stability
+    const availableUtxos = await getUTxOs()
+    const nonStakingUtxos = availableUtxos.filter(({address}) => !isGroup(address))
+    const groupAddressUtxos = availableUtxos.filter(({address}) => isGroup(address))
     const randomGenerator = PseudoRandom(seeds.randomInputSeed)
-    const shuffledUtxos = shuffleArray(availableUtxos, randomGenerator)
+    const shuffledUtxos =
+      txType === 'convert'
+        ? shuffleArray(nonStakingUtxos, randomGenerator)
+        : [
```
refi93, line 313: is it really intended to shuffle non staking and group address utxos separately, doesn't it hamper the randomness? I would find more straightforward (and randomized) to shuffle the whole array

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -383,7 +390,7 @@ const ShelleyWallet = ({config, randomInputSeed, randomChangeSeed, cryptoProvide
   }
 
   async function getHistory() {
```
refi93, line 392: I would add a return type annotation making clear it returns a promise

app/frontend/wallet/shelley/build-transaction.ts
```diff
@@ -1,10 +1,7 @@
 import {computeRequiredTxFee} from './helpers/chainlib-wrapper'
-import _ from 'lodash'
-// import {Input, Output, TxPlan} from '../cardano-wallet'
+// import _ from 'lodash'
```
refi93, line 2: I would remove this if it isn't used, no need to keep it. Btw, we don't have lodash in package.json, so we were relying on some dependency providing it

app/frontend/wallet/shelley/build-transaction.ts
```diff
@@ -42,7 +49,7 @@ export function computeTxPlan(
   inputs: Array<Input>,
   outputs: Array<Output>,
   possibleChange?: Output,
-  cert?: any
+  cert?: any // TODO: cert
```
refi93, line 52: can't we use "Cert" type here as well? It seems as if we were just passing that parameter to the TxPlan return value anyway

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -1,5 +1,5 @@
 import * as lib from '@emurgo/js-chain-libs'
-import {isValidAddress as isValidByronAddress} from 'cardano-crypto.js'
+// import {isValidAddress as isValidByronAddress} from 'cardano-crypto.js'
```
refi93, line 2: I would just remove it if it's not used

app/frontend/wallet/shelley/shelley-address-provider.ts
```diff
@@ -26,7 +26,8 @@ const shelleyStakeAccountPath = (account: number) => {
 }
 
 const _validateScheme = (cryptoProvider) => {
```
refi93, line 28: I would rather remove it completely - it's an "internal" function after all and I really hope that Shelley will learn from Byron "mistakes" and won't need to have multiple derivation schemes

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -160,12 +152,13 @@ const MyAddresses = ({accountIndex, cryptoProvider, gapLimit, blockchainExplorer
...
 
 const ShelleyBlockchainExplorer = (config) => {
+  // TODO: move to separate file
   const be = BlockchainExplorer(config)
 
   const fixAddress = (address) => (isShelleyAddress(address) ? bechAddressToHex(address) : address)
   const fix = (addresses: Array<string>): Array<string> => {
     return addresses.map(fixAddress)
-  }
+  } // TODO: rename to better name than "fix"
```
refi93, line 161: it would probably be more idiomatic to use "compose" helper:

`const compose = (...fns) => x => fns.reduceRight((v, f) => f(v), x);`

and then you can write the exported functions using it as compose(fn, transformShelleyAddresses)

source: https://medium.com/javascript-scene/master-the-javascript-interview-what-is-function-composition-20dfb109a1a0

refi93, line 161: I would name it `shelleyAddressesToHex` and also I would rewrite the exported functions below that are unnecessarily verbose, I mean reriting e.g:


```
getTxHistory: (addresses) => {
      return be.getTxHistory(fix(addresses))
},
...

to
```
getTxHistory: (addresses) => be.getTxHistory(shelleyAddressesToHex(addresses)),
```

File: app/frontend/components/pages/login/loginPage.tsx (hunk 2 above)
Comment (PeterBenc) [reply app/frontend/components/pages/login/loginPage.tsx]: Yes, you are right, someone reenabling this functionality would have to to a lot more than just commenting these lines, but not in this file. This file most likely wont change at all, just the implementation of the crypto providers, so that's the reason I only commented these lines. But anyway, we can always get them from git history, so I removed the lines related to hardware wallets.

File: app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx (hunk 3 above)
Comment (PeterBenc) [reply app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx]: Yeah, I think we should as well, I created a helper for this. It is used only in this one place but I thought we should separate its logic from `confirmTransactionDialog`. Open to suggestions tho, creating a helper for this might not seem necessary.

File: app/frontend/wallet/cardano-wallet.ts (hunk 7 above)
Comment (PeterBenc) [reply app/frontend/wallet/cardano-wallet.ts]: We definitely dont need the `getAccountInfo` function, that was just a forgotten piece of code, but the others I left there are there because there are called from the `action.ts` file. Of course only in case of `shelley` implementation of the wallet, but typescript shows an error if the `byron` wallet doesn't have them implemented.

File: app/frontend/wallet/shelley-wallet.ts (hunk 10 above)
Comment (PeterBenc) [reply app/frontend/wallet/shelley-wallet.ts:201]: The point of two dictionaries was to be able to choose a stake pool by ID and TICKER, and do both of these in O(1). But I agree we should create a separate issue for this. For example, it seems a ticker is not unique identifier for a stake pool, so this whole searching by ticker might not be a good idea.

File: app/frontend/wallet/shelley-wallet.ts (hunk 12 above)
Comment (PeterBenc) [reply app/frontend/wallet/shelley-wallet.ts]: You are right, fixed.

File: app/frontend/wallet/shelley-wallet.ts (hunk 13 above)
Comment (PeterBenc) [reply app/frontend/wallet/shelley-wallet.ts:311]: Yes, it is. At least we 've talked about it at the meeting, the plan was to spend non-staking utxos first. One reason is that non-staking utxos are not something you want to possess (in most cases) and the second one is that spending some coins from non-staking balance, and some from staking-balance might be confusing for users. But we can definitely change this, it doesn't seem that crucial to me.

File: app/frontend/wallet/cardano-wallet.ts (hunk 7 above)
Comment (refi93) [reply app/frontend/wallet/cardano-wallet.ts]: then I would make them throw an Error with a message stating that Byron wallet does not support this operation to make sure anybody calling it by mistake realizes it soon enough

File: app/frontend/wallet/shelley-wallet.ts (hunk 13 above)
Comment (refi93) [reply app/frontend/wallet/shelley-wallet.ts:311]: I see - that's worth a comment in the code then for future contributors

File: app/frontend/wallet/cardano-wallet.ts (hunk 7 above)
Comment (PeterBenc) [reply app/frontend/wallet/cardano-wallet.ts]: Yes you are right, I thought about this too, but the typescript had a problem with the function returning void instead of some object. Anyway, I fixed it so it throws Error on this unsupported operations.

**Comment by refi93** [comment]
CI is still failing with the following eslint warnings:

```
/home/circleci/repo/app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
 
  80:1  warning  This line has a length of 101. Maximum allowed is 100  max-len

/home/circleci/repo/app/frontend/wallet/byron/cardano-wallet-secret-crypto-provider.ts
 
   4:20  warning  'PkWitness' is defined but never used         no-unused-vars
  11:9   warning  'DerivationScheme' is defined but never used  no-unused-var

/home/circleci/repo/app/frontend/wallet/cardano-wallet.ts

  14:9  warning  'TxPlan' is defined but never used  no-unused-vars
 
✖ 4 problems (0 errors, 4 warnings)

```

could you please fix them @PeterBenc ?

PeterBenc:
The eslint errors remerged because I commented out the ignoring and upgrading typescript didn't help, anyway, I ignored them again for now and rebased the branch.

refi93:
changes deployed to testnet staging, could you please @PeterBenc test it with @xdzurman and see if everything is working as expected?

PR #620, opened 2020-06-09
Temporarily disable transactions parsing

xdzurman:
Parsing is not currently implemented and always fails, even if it is in a try-catch block. In the history, this very functionality caused memory leaks in GA-tracking.

### DIFF
--- a/server/middlewares/statsRedis.js
+++ b/server/middlewares/statsRedis.js
@@ -3,7 +3,6 @@ const device = require('device')
 const client = redis.createClient(process.env.REDIS_URL)
 const mung = require('express-mung')
 const normalizeUrl = require('normalize-url')
-const {parseTxBodyOutAmount, parseTxBodyTotalAmount} = require('../helpers/parseTxBody')
 
 const knownIps = new Set()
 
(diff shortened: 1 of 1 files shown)


server/middlewares/statsRedis.js
```diff
@@ -56,20 +55,8 @@ const trackTxSubmissions = mung.json((body, req, res) => {
...
-        txOutAmount = parseTxBodyOutAmount(txBody)
-        txTotalAmount = parseTxBodyTotalAmount(txBody)
-      } catch (e) {
-        // eslint-disable-next-line no-console
-        console.error(e)
-      }
-
-      incrCountersBy(`${txSubmissionType}:sentOut`, txOutAmount)
-      incrCountersBy(`${txSubmissionType}:sentTotal`, txTotalAmount)
+      incrCountersBy(`${txSubmissionType}:sentOut`, 0)
```
refi93, line 58: I would just omit the counters if it's not working anyway

refi93: same for statsGoogleAnalytics where I see that the value is commented out right now

refi93:
CI is still failing due to yarn audit errors, could you update the packages to fix that?

PR #624, opened 2020-06-12
Add simple table of contents to redis stats

PeterBenc:
Adds a table of contents to Redis stats page along with link to this table of contents from sections of the stats. I made this very simple as is the rest of the page, so open to suggestions.

### DIFF
--- a/server/statsPageRedis.js
+++ b/server/statsPageRedis.js
@@ -75,6 +75,25 @@ module.exports = function(app, env) {
     try {
       const stats = await getStats()
 
+      const tableOfContents = Object.keys(stats)
+        .map(
+          (subject) => `
+          <b>${subject}:</b>
+          ${Object.keys(stats[subject])
+    .map(
+      (period) => `
+            <ul>
+                <a href="#${(subject + period).replace(':', '-')}">${
+  isAdaAmountKey(period) ? `${period} (ADA)` : period
+}</a>
+          </ul>
+        `
+    )
+    .join('')}
+      `
+        )
+        .join(' ')
+
       const statsHtml = Object.keys(stats)
         .map(
           (subject) => `
@@ -83,7 +102,9 @@ module.exports = function(app, env) {
     .map(
       (period) => `
               <ul>
-                <li>${isAdaAmountKey(period) ? `${period} (ADA)` : period} </li>
+                <li id=${(subject + period).replace(':', '-')}>
+                  ${isAdaAmountKey(period) ? `${period} (ADA)` : period}
+                </li>
                 <ul>
                   ${stats[subject][period]
     .map((item) => {
@@ -108,12 +129,18 @@ module.exports = function(app, env) {
         <!doctype html>
         <html>
           <head>
-            <meta charset="utf-8"/>
+            <meta charset="utf-8" name="viewport"
+                      content="height=device-height, 
+                      width=device-width, initial-scale=1.0, 
+                      minimum-scale=1.0, maximum-scale=1.0, 
+                      user-scalable=no, target-densitydpi=device-dpi">
             <title>AdaLite Wallet Stats</title>
             <link rel="icon" type="image/ico" href="assets/favicon.ico">
           </head>
           <body>
             Stats of transaction submissions and estimates of unique IPs visits per day, month and in total.
+            <a href="#tableOfContents" style="position:fixed;top:93%;left:calc(10% + 250px)">Table</a>
+            <div id="tableOfContents">${tableOfContents}</div>
             <div>${statsHtml}</div>
           </body>
         </html>


server/statsPageRedis.js
```diff
@@ -75,15 +75,37 @@ module.exports = function(app, env) {
...
-          (subject) => `
+          (subject, i) => `
             <b>${subject}:</b>
             ${Object.keys(stats[subject])
     .map(
-      (period) => `
+      (period, j) => `
               <ul>
-                <li>${isAdaAmountKey(period) ? `${period} (ADA)` : period} </li>
+                ${j % 3 === 0 ? '<a href="#tableOfContents">Table</a>' : ''}
```
refi93, line 105: This seems weird - why don't we use as the anchor ids directly the `<subject>:<period>` key instead of numbers?

server/statsPageRedis.js
```diff
@@ -75,15 +75,37 @@ module.exports = function(app, env) {
...
+                <li id=${i * 10 + j}>
```
refi93, line 106: This seems weird - why don't we use as the anchor ids directly the `<subject>:<period>` key instead of numbers?

server/statsPageRedis.js
```diff
@@ -75,15 +75,37 @@ module.exports = function(app, env) {
     try {
       const stats = await getStats()
 
+      const tableOfContents = Object.keys(stats)
+        .map(
+          (subject, i) => `
+          <b>${subject}:</b>
+          ${Object.keys(stats[subject])
+    .map((period, j) =>
+      j % 3 === 0
```
refi93, line 84: I would not limit the table of contents only to the first period - it's imho fine to have them all there (especially given how many daily records are there before the monthly ones) and save this condition

server/statsPageRedis.js
PeterBenc: You are right, using the numbers is a bit weird, but the purpose of it was to make the ids unique, using subject + period makes a lot more sense.

server/statsPageRedis.js
PeterBenc: I decided to do this for two reasons, if the table contained every period it seemed too long. The second reason was that once you see the total number, which is just one line, you always see the whole monthly stats(just couple of lines) and at least the beginning of daily stats. I can't imagine someone going to the table of contents and looking for daily stats if the only thing he had to do is to scroll just a little bit or not at all. What do you think? I have no problem removing the condition if you think it makes more sense.

PeterBenc:
Did you mean `fixed` position? For some reason when a read the comment I saw fixed position and it seemed like a great idea since I was kind of struggling with where to put the link to the table. (Btw. I had to add some `meta` since the fixed position didn't work properly on mobile devices)
In addition to making the link to table `fixed position` I made it calc the `left` attribute. Giving it just left 95% and having the link in the right bottom corner of a desktop didn't seem handy at all. So this way it is right bottom, kind of floating in the middle of nowhere but at least closer to the content. I tested it on some mobile devices and It seems reasonably positioned in all of them. But again if this doesn't make sense to you I can change it so its just in the right bottom corner and stop over-engineering this anymore :smile: 
On full screen desktop:
image
On mobile devices: 
image

refi93:
Yeah, I actually meant fixed position - sorry, I'm a bit rusty on css :) Sure, I would not over-engineer it - as long as the link is visible and not overlapping with the text on reasonable resolutions, it's fine IMO

server/statsPageRedis.js
refi93: I see - still, I found the table quite straightforward to navigate even if there were all the intervals. There are already like 20 months in mainnet adalite, which takes a significant portion of the viewport and may be annoying to scroll over - and the daily stats are arguably more interesting than monthly or total as they provide the most direct feedback to any changes we did or potential issues we have

A possible tradeoff could be putting the daily, monthly and total link on a single line so they don't take up so much vertical space

server/statsPageRedis.js
```diff
@@ -75,6 +75,27 @@ module.exports = function(app, env) {
...
+        ? `
+            <ul>
+                <a href="#${subject + period}">${
```
refi93, line 87: this introduces colons (":") in the ids which is not good given that it clashes with css selector syntax. I would rather wrap it in a str.replace

server/statsPageRedis.js
```diff
@@ -108,12 +131,18 @@ module.exports = function(app, env) {
...
+                      content="height=device-height, 
+                      width=device-width, initial-scale=1.0, 
+                      minimum-scale=1.0, maximum-scale=1.0, 
+                      user-scalable=no, target-densitydpi=device-dpi">
             <title>AdaLite Wallet Stats</title>
             <link rel="icon" type="image/ico" href="assets/favicon.ico">
           </head>
           <body>
             Stats of transaction submissions and estimates of unique IPs visits per day, month and in total.
+            <a href="#tableOfContents" style="position:fixed;top:93%;left:calc(10% + 250px)">Table</a>
```
refi93, line 144: wouldn't it be more sensible to use em instead of 250px? What about narrow screens?

PeterBenc: maybe it would, I haven't experimented with for that long. This solution works for screens with width >= 320px which seemed reasonable but I agree this solution is far from ideal. I 'll look into it.

server/statsPageRedis.js
PeterBenc: Agreed, good idea, putting the total, monthly and daily links in on the same line seems like the best solution but it look terrible in mobile design, so I added all the values to the table.

server/statsPageRedis.js
PeterBenc: Using em instead of px didn't seem to make much of a difference, as mentioned the layout is still acceptable on 320px width devices which seems good enough to me.

PR #626, opened 2020-06-15
Fix Byron tests

### LINKED ISSUE #623: Failing unit tests in testnet branch
running `yarn test-chrome-headless` is currently failing on `testnet` branch. There are the folloring issues:

1. "Wallet Import" testes are failing with Unhandled promise rejection. It's probably caused by an outdated version of mocha and mocha-headless-chrome

2. Seven Byron-related unit tests are failing:
```
      successful transaction fee computation
        1) should compute the right transaction fee for given transaction
        2) should compute the right transaction fee for shorter outgoing address
      transaction serialization
        3) should properly serialize transaction before signing
        4) should properly discard utxos that cause an increase of fee higher than their value
        5) should properly compute transaction hash
        6) should properly serialize the whole transaction
      test transaction submission
        7) should properly submit transaction
```

elfinka:
1. for tests failing on getTxPlan:
getTxPlan expects just one argument, containing the values for address, coins and donationAmount. The original test code thus didn't provide any of these values and plan generation failed

2. Shelley testnet tests were failing on global leak detection - it seems the global Buffer causing the leaks is not needed (anymore)

3. Wallet import tests seem to be working fine

closes #623

### DIFF
--- a/app/frontend/helpers/wasmLoader.js
+++ b/app/frontend/helpers/wasmLoader.js
@@ -1,4 +1,3 @@
-import {Buffer} from 'buffer'
 // interface extendedWindow extends Window {
 //   wasm?: any;
 // }
@@ -6,5 +5,4 @@ import {Buffer} from 'buffer'
 export default async function loadWasmModule() {
   // (window as extendedWindow).wasm = await import('@emurgo/js-chain-libs/js_chain_libs_bg.wasm')
   window.wasm = await import('@emurgo/js-chain-libs/js_chain_libs_bg.wasm')
-  window.Buffer = Buffer
 }
--- a/app/tests/src/cardano-wallet.js
+++ b/app/tests/src/cardano-wallet.js
@@ -184,7 +184,10 @@ describe('successful transaction fee computation', () => {
     mockNet.mockBulkAddressSummaryEndpoint()
     mockNet.mockUtxoEndpoint()
 
-    assert.equal((await wallets.used.getTxPlan(myAddress, 47, 0)).fee, 179288)
+    assert.equal(
+      (await wallets.used.getTxPlan({address: myAddress, coins: 47, donationAmount: 0})).fee,
+      179288
+    )
     mockNet.clean()
   })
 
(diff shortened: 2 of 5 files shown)


app/frontend/helpers/wasmLoader.js
```diff
@@ -1,10 +1,8 @@
-import {Buffer} from 'buffer'
 // interface extendedWindow extends Window {
 //   wasm?: any;
 // }
 
 export default async function loadWasmModule() {
   // (window as extendedWindow).wasm = await import('@emurgo/js-chain-libs/js_chain_libs_bg.wasm')
   window.wasm = await import('@emurgo/js-chain-libs/js_chain_libs_bg.wasm')
-  window.Buffer = Buffer
```
refi93, line 9: @PeterBenc I see you commited this line in January 2020 - do you know why we needed it?

https://github.com/vacuumlabs/adalite/blob/4975453b030991f8fc6856f76d41c0ffcf004dbd/app/frontend/helpers/wasmLoader.js#L9

app/tests/src/cardano-wallet.js
```diff
@@ -184,15 +184,21 @@ describe('successful transaction fee computation', () => {
...
     mockNet.clean()
   })
 
   it('should compute the right transaction fee for shorter outgoing address', async () => {
     const mockNet = mockNetwork(mockConfig2)
     mockNet.mockUtxoEndpoint()
 
-    assert.equal((await wallets.used.getTxPlan(shortAddress, 47, 0)).fee, 177838)
+    assert.equal(
+      (await wallets.used.getTxPlan({address: shortAddress, coins: 47, donationAmount: 0})).fee,
```
refi93, line 199: @elfinka could you describe in the PR description what issue did you find in the tests and how do your changes address the problem(s)?

app/frontend/helpers/wasmLoader.js
PeterBenc: I remember trying something with it, I think it was a hotfix for something which was later solved properly. Its just a forgotten piece of code, to my knowledge the line is no longer useful for anything. I tested using the wasm library without this line and everything worked as expected.

app/tests/src/cardano-wallet.js
elfinka: Added description

PR #634, opened 2020-06-18
Fix tx submission & add hw wallet login

elfinka:
* Tx submission: the only problem was reloading stakepool info, which shouldn't be done for Byron
* HW wallet: only added the option to login screen, no additional changes were needed to log in successfully using trezor emulator
(trezor however complained that we're using outdated version, I'll look into that next)


refi93:
also, not sure if relevant, but in the develop branch that is the base for testnet banch there is an additional fix for a "trezor regression" https://github.com/vacuumlabs/adalite/commit/8f086c690c1e8583b1d17d521352a282295a581e , so I would suggest checking it out and see if it's needed to port it here as well

PR #636, opened 2020-06-19
Add missing master commits

elfinka:
This should contain all merges that were made to master but not to testnet branch since they both diverged from develop.

With one exception: https://github.com/vacuumlabs/adalite/commit/c2c1160495b063114298917075b56c70f7b81971
should I add this as well? I'm not sure if it's still relevant here and how to check it works after merge.


PeterBenc:
It s relevant to add that commit as well, to test it you need to run Redis locally and set REDIS_URL env variable to `redis://localhost:6379` (usually). But feel free to just add the commit and I can test it.

PR #639, opened 2020-06-29
Test for actions.ts

elfinka:
The purpose was to provide a way to test actions.ts (those that contain more logic than setting 1 state field).
In this PR this is only done for loadWallet, other actions will be covered in subsequent PRs.

I had some problems with initializing config, which is needed for both getting initial state and also used across the codebase directly - for that I set up some dummy config on tests startup (as other than actions tests try to load it, though other tests are OK with it remaining null).

### DIFF
--- a/app/frontend/store.ts
+++ b/app/frontend/store.ts
@@ -2,7 +2,7 @@ import devtools from './libs/unistore/devtools'
 import {ADALITE_CONFIG} from './config'
 import {initialState} from './state'
 
-const createDefaultStore = require('unistore').default
+const createDefaultStore = require('./libs/unistore').default
 
 const createStore = () =>
   ADALITE_CONFIG.ADALITE_ENABLE_DEBUGGING === 'true'
--- a/app/tests/src/actions/wallet-actions.js
+++ b/app/tests/src/actions/wallet-actions.js
@@ -1,0 +1,93 @@
+import mockNetwork from '../common/mock'
+import {ADALITE_CONFIG} from '../../../frontend/config'
+import {CRYPTO_PROVIDER_TYPES} from '../../../frontend/wallet/constants'
+import mnemonicToWalletSecretDef from '../../../frontend/wallet/helpers/mnemonicToWalletSecretDef'
+import assert from 'assert'
+import {assertPropertiesEqual, setupState} from './actions'
+
+let state, action
+
+beforeEach(() => {
+  ;[state, action] = setupState()
+})
+
+const expectedStateChanges = {
+  walletIsLoaded: true,
+  balance: 1500000,
+  loading: false,
+  mnemonicAuthForm: {
+    mnemonicInputValue: '',
+    mnemonicInputError: null,
+    formIsValid: false,
+  },
+  usingHwWallet: false,
+  hwWalletName: undefined,
+  isDemoWallet: false,
+  showDemoWalletWarningDialog: false,
+  showGenerateMnemonicDialog: false,
+  // send form
+  sendAmount: {fieldValue: '', coins: 0},
+  sendAddress: {fieldValue: ''},
+  donationAmount: {fieldValue: '', coins: 0},
+  sendResponse: '',
+}
+
+it('Load shelley wallet', async () => {
+  ADALITE_CONFIG.ADALITE_CARDANO_VERSION = 'shelley'
+
+  const mockNet = mockNetwork(ADALITE_CONFIG)
+  mockNet.mockBulkAddressSummaryEndpoint()
+  mockNet.mockGetAccountInfo()
+  mockNet.mockGetStakePools()
+  mockNet.mockGetConversionRates()
+
+  expectedStateChanges.validStakepools = {
+    d4b1243dfc0bec57f146a90d85b478cdd3e0e646c43801c2bebd6792580a7db2: {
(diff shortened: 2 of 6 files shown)


refi93:
could you please rebase it to the current develop branch?

refi93:
could you please rebase it to the current develop branch?

app/tests/src/actions.js
```diff
@@ -0,0 +1,127 @@
+import assert from 'assert'
```
refi93, line 1: I would split the actions tests into multiple files. Actions.ts is big, and if we are going to extend it, the respective tests file for it would be even bigger and harder to navigate. So I would suggest to have an "actions" folder and multiple test files within it

For now, I would split it into "byron" and "shelley" and possibly "common" tests

app/tests/src/common/setup-test-config.js
```diff
@@ -0,0 +1,21 @@
...
+  "ADALITE_ENABLE_DEBUGGING": false,
+  "ADALITE_APP_VERSION": "3.7.0",
+  "ADALITE_LOGOUT_AFTER": "900",
+  "ADALITE_TREZOR_CONNECT_URL": "",
+  "ADALITE_SUPPORT_EMAIL": "support@test.test",
+  "ADALITE_FIXED_DONATION_VALUE": "40",
+  "ADALITE_MIN_DONATION_VALUE": "1",
+  "ADALITE_STAKE_POOL_ID": "f61c42cbf7c8c53af3f520508212ad3e72f674f957fe23ff0acb49733c37b8f6",
+  "ADALITE_ENV": "local",
+  "SENTRY_DSN": "https://d77d3bf9d9364597badab9c00fa59a31@sentry.io/1501383",
```
refi93, line 16: I would leave this empty as we don't want to accidentally trigger sentry (an error tracking platform) from tests

app/frontend/store.ts
```diff
@@ -2,7 +2,7 @@ import devtools from './libs/unistore/devtools'
 import {ADALITE_CONFIG} from './config'
 import {initialState} from './state'
 
-const createDefaultStore = require('unistore').default
+const createDefaultStore = require('./libs/unistore').default
```
refi93, line 5: why is this change needed?

elfinka: without that I'm getting Module Not Found error, apparently when running tests I need the full path.

app/tests/src/actions.js
elfinka: As Byron & Shelley tests will often share testdata and other preparations for testing the same function, I split it rather by functional groupings (wallet - things around login, transaction - things around transactions & fees, ...)

app/tests/src/actions/wallet-actions.js
```diff
@@ -0,0 +1,93 @@
...
+  showDemoWalletWarningDialog: false,
+  showGenerateMnemonicDialog: false,
+  // send form
+  sendAmount: {fieldValue: '', coins: 0},
+  sendAddress: {fieldValue: ''},
+  donationAmount: {fieldValue: '', coins: 0},
+  sendResponse: '',
+}
+
+it('Load shelley wallet', async () => {
```
refi93, line 35: this test is timing on my machine, I would increase the timeout to a sfe value, let's say 5 seconds

elfinka: Increased just for this test. However it's a bit weird, when I run only this one test it runs 3x faster than when I uncomment the rest of the suite.

refi93: Interesting, also not sure why, but when I run the tests locally and I include the test set `Import/Export Wallet as JSON`, I get the following error: 

```
(node:132999) UnhandledPromiseRejectionWarning: Error: Page crashed!
    at Page._onTargetCrashed (/home/rafael/workplace/cardano/adalite/app/node_modules/puppeteer/lib/Page.js:170:24)
    at CDPSession.Page.client.on.event (/home/rafael/workplace/cardano/adalite/app/node_modules/puppeteer/lib/Page.js:125:56)
    at CDPSession.emit (events.js:197:13)
    at CDPSession._onMessage (/home/rafael/workplace/cardano/adalite/app/node_modules/puppeteer/lib/Connection.js:200:12)
    at Connection._onMessage (/home/rafael/workplace/cardano/adalite/app/node_modules/puppeteer/lib/Connection.js:112:17)
    at WebSocketTransport._ws.addEventListener.event (/home/rafael/workplace/cardano/adalite/app/node_modules/puppeteer/lib/WebSocketTransport.js:44:24)
    at WebSocket.onMessage (/home/rafael/workplace/cardano/adalite/app/node_modules/ws/lib/event-target.js:120:16)
    at WebSocket.emit (events.js:197:13)
    at Receiver.receiverOnMessage (/home/rafael/workplace/cardano/adalite/app/node_modules/ws/lib/websocket.js:789:20)
    at Receiver.emit (events.js:197:13)
(node:132999) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:132999) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Are you able to replicate it?

PR #646, opened 2020-07-06
Allow NamedError to keep parent stack

xdzurman:
Refactor, make arbitrary parameters independent on their order, remove showHelp parameter as it is not used anywhere

### DIFF
--- a/app/frontend/helpers/NamedError.ts
+++ b/app/frontend/helpers/NamedError.ts
@@ -1,8 +1,13 @@
-function NamedError(name, message = '', showHelp = '') {
-  const e = new Error(message || name || showHelp)
+type OptionalParams = {
+  message?: string
+  errorStack?: string
+}
+
+function NamedError(name: string, optionalParams: OptionalParams = {}) {
+  const e = new Error(optionalParams.message || name)
   e.name = name
-  e.message = message
-  ;(e as any).showHelp = showHelp
+  e.message = optionalParams.message || ''
+  e.stack = optionalParams.errorStack || e.stack
   return e
 }
 
--- a/app/frontend/helpers/NamedError.ts
+++ b/app/frontend/helpers/NamedError.ts
@@ -1,13 +1,15 @@
 type OptionalParams = {
   message?: string
-  errorStack?: string
+  causedBy?: Error
 }
 
 function NamedError(name: string, optionalParams: OptionalParams = {}) {
   const e = new Error(optionalParams.message || name)
   e.name = name
   e.message = optionalParams.message || ''
-  e.stack = optionalParams.errorStack || e.stack
+  e.stack = optionalParams.causedBy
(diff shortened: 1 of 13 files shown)


app/frontend/helpers/NamedError.ts
```diff
@@ -1,8 +1,13 @@
...
+  errorStack?: string
+}
+
+function NamedError(name: string, optionalParams: OptionalParams = {}) {
+  const e = new Error(optionalParams.message || name)
   e.name = name
-  e.message = message
-  ;(e as any).showHelp = showHelp
+  e.message = optionalParams.message || ''
+  e.stack = optionalParams.errorStack || e.stack
```
PeterBenc, line 10: Maybe if we wanted Optionalparams to contain (almost) anything like the txPlan, we could implement it like this. Im not that sure that just the stackTrace will help us to solve the bug, so imho adding the txPlan to the error makes sense. Additionally, it could probably be used also in other places.
``` 
function NamedError(name: string, optionalParams: OptionalParams = {}) {
let e = new Error(optionalParams.message || name)
  ;(e as any) = {
    name,
    ...optionalParams,
    stack: optionalParams.errorStack || e.stack,
  }

Comment (xdzurman) [reply app/frontend/helpers/NamedError.ts:10]: Since optional params happens to be an object, it would be easy to set the whole object as sentry error context you mentioned in #645 .

File: app/frontend/helpers/NamedError.ts [hunk 2]
```diff
@@ -1,8 +1,13 @@
-function NamedError(name, message = '', showHelp = '') {
-  const e = new Error(message || name || showHelp)
+type OptionalParams = {
+  message?: string
+  errorStack?: string
+}
+
+function NamedError(name: string, optionalParams: OptionalParams = {}) {
```
refi93, line 6: I think it would be a nicer API to have an optional `causedBy` parameter which would explicitly be required to be an instance of Error and then building our error stack like "Caused by " + causedBy.stack

related stack overflow question for reference: https://stackoverflow.com/questions/25763837/how-to-specify-a-caused-by-in-a-javascript-error

app/frontend/helpers/NamedError.ts
refi93: I'm not really sure how safe it is in - I would rather refrain from it if we don't need it for sure

app/frontend/helpers/NamedError.ts
xdzurman: Do you mean replacing `errorStack: string` in OptionalParams by `causedBy: Error`? Or replacing the whole OptionalParams type? Because some errors optionally provide an error message, which makes more optional parameters, thus creating chaos in their order, which is why I put them into one object.

xdzurman: Right, the edited version did not appear for me in time.

Issue #616, opened 2020-06-05
Add banner to report widespread issues if present


refi93:
Currently if our backend is down or if there are known issues with the stability of our tx submitting service or any other issue, the users have no information about that until they run into it. It would be better UX to have a banner on top of the page informing the users about the problem, ideally with a link to an article on what they can do about it.

The text would be configurable via an environment variable and generic issues like the servers being completely down could trigger that banner automatically, overriding the text configured by the environment variable, if present

refi93:
closed via https://github.com/vacuumlabs/adalite/pull/648

PR #648, opened 2020-07-09
add error banner and env variable for its content

PeterBenc:
Adds styles and component for error banner
Calls function on componentDidMount to load possible error message for the banner
Refactors all boolean states names
Partly addresses #616


app/frontend/actions.ts
```diff
@@ -1110,6 +1110,15 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     })
   }
 
+  const getErrorBannerContent = (state) => {
+    const errorBannerContent = ADALITE_CONFIG.ADALITE_ERROR_BANNER_CONTENT
+    const showErrorBanner = errorBannerContent !== '' && true
```
refi93, line 1115: I would rather use `!!errorBannerContent` here - maybe not an issue now, but `errorBannerContent !== ''` would evaluate to true even if the content is undefined which seems like a chance to shoot ourselves in the foot

app/frontend/actions.ts
```diff
@@ -1110,6 +1110,15 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+    setState({
+      errorBannerContent,
+      showStakingBanner: showErrorBanner ? false : state.showStakingBanner,
```
refi93, line 1118: I think it would be cleaner for the staking banner showing logic to check if error banner is being shown rather than the error banner overriding this. This way the logic to show/hide the staking banner would be kept in a single place and it would be less likely to lead to confusion in the future

app/frontend/components/pages/login/loginPage.tsx
```diff
@@ -160,9 +165,11 @@ class LoginPage extends Component<Props, {isDropdownOpen: boolean}> {
         {authMethod === 'file' && <KeyFileAuth />}
       </div>
     )
+    getErrorBannerContent()
```
refi93, line 168: I think this belongs rather to componentDidMount callback than the rendering logic and for this usecase and also since the function mutates the state, IMHO the verb "load" would fit more. "get" kind of leads the reader into thinking it's not a mutating operation which is not true

PeterBenc: true, I was considering calling the function in componentDidMount, but then I thought about a situation when the function would call some request to backend health check API, (this could take some time) while the banner would not be rendered, and then suddenly after the function returns something the whole view would jump down to render the banner. This way the type of banner is known before render. Either way, we are either postponing the render by the time the function returns or postponing the render of the banner while the function returns. It makes sense to me either way but lets call it in `componentDidMount` then.

app/frontend/actions.ts
```diff
@@ -1106,7 +1106,16 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
   const closeUnexpectedErrorModal = (state) => {
     setState({
-      showUnexpectedErrorModal: false,
+      shouldShowUnexpectedErrorModal: false,
+    })
+  }
+
+  const loadErrorBannerContent = (state) => {
+    const errorBannerContent = ADALITE_CONFIG.ADALITE_ERROR_BANNER_CONTENT
+    const shoulShowErrorBanner = !!errorBannerContent
```
refi93, line 1115: typo -> "shoul" -> "should"

refi93:
I tried setting the error banner with

```
ADALITE_ERROR_BANNER_CONTENT="<a href="www.google.sk">AHOJ</a> We are sorry but our support for ITN is temporarily disrupted. As we are focusing on the Shelley main-net fork we don't know when (and if) we will be able to restore this functionality."
```

and the adalite page failed to load at all since the JSON parser of the config failed - we should make sure to encode that string properly into the data attribute and ensure that even if we set some "weird" banner text the site does not crash and the malformed string gets caught during config validation

PeterBenc:
Thanks Rafa, I changed the implementation so you can insert an HTML to the .env variable and everything should work as expected. On top of this, I added validation for the inserted HTML. There seems to be no easy way to validate the HTML on server-side, (not without 3rd party library or long scripts) so I decided to add the validation to the frontend side. This way if a developer inserted an invalid HTML, the banner would not show at all. 
Open to suggestions on handling invalid values on server-side.

PR #651, opened 2020-07-16
Shelley mainnet address packing

PeterBenc:
Replaces cardano-crypto.js lib for a new version that is able to pack Shelley mainnet address types. Uses this library to derive Shelley mainnet addresses for a given secret and removes old testnet related logic.

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -139,7 +139,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
           await loadWasmModule()
           const cryptoProvider = ShelleyJsCryptoProvider({
             walletSecretDef,
-            network: NETWORKS.SHELLEY.MAINNET,
+            network: NETWORKS.SHELLEY[ADALITE_CONFIG.ADALITE_NETWORK],
           })
 
           wallet = await ShelleyWallet({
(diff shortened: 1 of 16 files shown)


refi93:
please fix eslint error for CI to pass, otherwise LGTM

app/frontend/actions.ts
```diff
@@ -139,7 +139,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
           await loadWasmModule()
           const cryptoProvider = ShelleyJsCryptoProvider({
             walletSecretDef,
-            network: NETWORKS.SHELLEY.INCENTIVIZED_TESTNET,
+            network: NETWORKS.SHELLEY.MAINNET,
```
refi93, line 142: I would make this configurable on .env level - we need a quick way to switch between networks once mainnet becomes live

app/frontend/wallet/constants.ts
```diff
@@ -22,6 +22,18 @@ export const NETWORKS = {
...
+    MAINNET: {
+      name: 'shelley-mainnet',
+      networkId: 0,
+      protocolMagic: 764824073,
+      ttl: undefined,
+    },
+    HASKELL_TESTNET: {
+      name: 'htn',
+      networkId: 0,
+      protocolMagic: 764824073,
```
refi93, line 34: this should be 42 indeed

refi93: can be seen e.g. here: https://iohk.zendesk.com/hc/en-us/articles/900001220783-Transaction-errors-Failed-reading-Your-input-is-either-malformed-or-not-hex-encoded

PR #654, opened 2020-07-20
Adjust staking related logic to new backend

PeterBenc:
First of all, tries to remove logic for multiple pools. Connects and adjust the staking tab to new backend response formats.
Gets additional pool into during fee calculation. Fixes account key hex so the correct account info is pulled from the backend.
Temporarily hot-fixes test for Shelley fee calculation. Just skipping the test causes some other test to be failing.


app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -27,6 +27,10 @@ export const accountAddressFromXpub = (stakeXpub: Xpub, networkId) => {
   return bech32.encode({prefix: 'addr', data: addrBuffer})
 }
 
+export const accountHexAddressFromXpub = (stakeXpub: Xpub, networkId) => {
```
refi93, line 30: I would annotate the return type as it is not clear at first glance - is it really a (hex)string or a buffer? I would do the same for the surrounding functions

PR #699, opened 2020-08-18
Set transaction ttl

DavidTranDucVL:
Implements logic for dynamic fee calculation. At first, we try to get the latest slot from out backend and add 3600 slots to it (meaning the ttl for tx is 1 hour) and if the request for some reason fails, we calculate the ttl from the date.

### DIFF
--- a/app/frontend/wallet/constants.ts
+++ b/app/frontend/wallet/constants.ts
@@ -27,7 +27,8 @@ export const NETWORKS = {
       name: 'mainnet',
       networkId: 1,
       protocolMagic: 764824073,
-      ttl: 10000000,
+      maxTtl: 10000000,
+      ttl: 3600,
     },
     HASKELL_TESTNET: {
       name: 'htn',
--- a/app/frontend/wallet/constants.ts
+++ b/app/frontend/wallet/constants.ts
@@ -27,7 +27,8 @@ export const NETWORKS = {
       name: 'mainnet',
       networkId: 1,
       protocolMagic: 764824073,
-      maxTtl: 10000000,
+      eraStartSlot: 4492800, // 21600 slot x 208 epochs
+      eraStartDateTime: Date.parse('29 Jul 2020 21:44:51 UTC'),
       ttl: 3600,
     },
     HASKELL_TESTNET: {
--- a/app/frontend/wallet/shelley-wallet.ts
+++ b/app/frontend/wallet/shelley-wallet.ts
@@ -169,6 +169,30 @@ const ShelleyBlockchainExplorer = (config) => {
     return {validStakepools}
   }
 
+  async function getBestBlock() {
+    let bestBlock: number = 0
+    try {
+      const response = await request(
+        `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/bestBlock`
+      )
+      bestBlock = response.Right.bestBlock
+    } catch (e) {}
+
+    return bestBlock
+  }
+
+  async function getBestSlot() {
+    let bestSlot: number = 0
+    try {
+      const response = await request(
+        `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/bestSlot`
+      )
+      bestSlot = response.Right.bestSlot
+    } catch (e) {}
+
+    return bestSlot
+  }
+
   return {
     getTxHistory: (addresses) => be.getTxHistory(addresses),
     fetchTxRaw: be.fetchTxRaw,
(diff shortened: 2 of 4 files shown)


app/frontend/wallet/shelley-wallet.ts
```diff
@@ -260,14 +286,24 @@ const ShelleyWallet = ({
...
-  function prepareTxAux(plan) {
+  async function calculateTtl() {
+    const bestSlot = await blockchainExplorer.getBestSlot()
+
+    if (bestSlot) {
+      return bestSlot + cryptoProvider.network.ttl
+    } else {
+      return cryptoProvider.network.maxTtl
+    }
+  }
```
PeterBenc, line 297: You can do this by 
```
return bestSlot ? bestSlot + cryptoProvider.network.ttl : cryptoProvider.network.maxTtl

File: app/frontend/wallet/shelley-wallet.ts [hunk 2]
```diff
@@ -260,14 +286,24 @@ const ShelleyWallet = ({
     }
   }
 
-  function prepareTxAux(plan) {
+  async function calculateTtl() {
```
PeterBenc, line 289: As mentioned, I would get the time of the beginning of the Shelley era and add the time passed since then.

app/frontend/wallet/constants.ts
```diff
@@ -27,7 +27,8 @@ export const NETWORKS = {
       name: 'mainnet',
       networkId: 1,
       protocolMagic: 764824073,
-      ttl: 10000000,
+      maxTtl: 10000000,
```
PeterBenc, line 30: What we should do at least, is make this constant bigger. But I think we should find a way to calculate this in case the request fails. We don't want tx to have TTL for years nor we want to update this constant every month. I would probably remove this constant altogether.

In addition to this, I would maybe add time representing the start of Shelley era (since it was the moment from which the epochs have a slot every second) and use this constant to calculate ttl by adding time passed since this moment.

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -191,6 +191,30 @@ const ShelleyBlockchainExplorer = (config) => {
...
+    let bestBlock: number = 0
+    try {
+      const response = await request(
+        `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/bestBlock`
+      )
+      bestBlock = response.Right.bestBlock
+    } catch (e) {}
+
+    return bestBlock
+  }
```
PeterBenc, line 204: Do we use this function anywhere?

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -191,6 +191,30 @@ const ShelleyBlockchainExplorer = (config) => {
...
+
+  async function getBestSlot() {
+    let bestSlot: number = 0
+    try {
+      const response = await request(
+        `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/bestSlot`
+      )
+      bestSlot = response.Right.bestSlot
+    } catch (e) {}
```
PeterBenc, line 213: Since you are not doing anything else in the try block, I would use `.catch(e)` syntax. See the implementation of the function above.

PeterBenc (changes requested):
Other than the comments, LGTM, I've tested with a couple of txs, please test it again once again after you implement the requested changes.

app/frontend/wallet/shelley-wallet.ts
DavidTranDucVL: Fixed

app/frontend/wallet/constants.ts
DavidTranDucVL: Fixed

app/frontend/wallet/shelley-wallet.ts
DavidTranDucVL: Removed

app/frontend/wallet/shelley-wallet.ts
DavidTranDucVL: Fixed

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -260,14 +265,28 @@ const ShelleyWallet = ({
...
+  async function calculateTtl() {
+    return (
+      cryptoProvider.network.ttl +
+      (await blockchainExplorer
+        .getBestSlot()
+        .then((res) => res.Right.bestSlot)
+        .catch(
+          () =>
+            cryptoProvider.network.eraStartSlot +
+            Math.floor((Date.now() - cryptoProvider.network.eraStartDateTime) / 1000)
```
PeterBenc, line 277: This may be the case where I would prefer to use normal try-catch syntax and try not to write all the logic into the return statement. I don't mind that much, but I can imagine this being a bit more readable.

DavidTranDucVL: Refactored

PeterBenc (approved):
LGTM, I've tested it with the ledger, working as expected.

PR #705, opened 2020-08-26
Change staking screen descriptions and make UI design more consistent

DavidTranDucVL:
Changes:
- updated balloon description on Staking Balance, Available Balance
- changed colors on staking page from blue to pink, so it's consistent with the rest of the app
- changed order of balances on staking page
  - available balance and rewards are above the line
  - staking balance and nonstaking balance are bellow the line (nonstaking balance is hidden if equals to 0)
  - this way it also makes sense mathematically
- similarly looking white buttons with grey text/icons are now using `button secondary` class
- withdraw button has an icon now
- refresh button and withdraw buttons are same width
- updated styles to fix black borders around buttons on chrome browser
- renamed some of existing elements to be consistent across pages (like Balance -> Available Balance)

New:
- added `Accordion` element
- changed "delegate stake card" to use `Accordion` element
  - when user has already delegated, this element is by default contracted
  - if he hasn't delegated yet, it's expanded by default
  - this is added only to pc user, not mobile/tablet users

### DIFF
--- a/app/frontend/components/common/accordion.tsx
+++ b/app/frontend/components/common/accordion.tsx
@@ -1,0 +1,33 @@
+import {Fragment, h} from 'preact'
+import {useState, useCallback} from 'preact/hooks'
+import {DropdownCaret} from './svg'
+
+interface Props {
+  initialVisibility: Boolean
+  header: h.JSX.Element
+  body: h.JSX.Element
+}
+
+const Accordion = ({initialVisibility, header, body}: Props) => {
+  const [visible, setVisible] = useState(initialVisibility)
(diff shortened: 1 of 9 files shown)


xdzurman (approved):
LGTM

PeterBenc (approved):
image
could you please add some padding to the SVG so it's centered and in line with the `Delegate Stake`.

And IMO, feel free to up the version to 5.1.0 in `package.json`

Other than then it seems great :+1:

DavidTranDucVL:
I fixed the positioning of the caret icon

PR #706, opened 2020-08-28
Fix eslint errors, remove local version of trezor-connect

refi93:
Motivation: yarn eslint was failing, polluting the pipeline with errors and causing it to fail, this PR fixes those errors

Changes:
* fix eslint errors
* remove local version of trezor-connect which is no longer needed and generated part of the eslint errors

Testing: 
* `yarn eslint` passes
* able to log in with trezor device, successfully sent tx to myself

unit tests (`yarn test-chrome-headless`) are still failing, I will investigate/fix in the next PR


PeterBenc (approved):

PeterBenc:
After the rebase, new eslint errors emerged, (except one or two) all related to importing typescript types. I ve looked into it a bit, tried a couple of things with adding some rules into `eslintrc` file and adding `@typescript-eslint` to dev dependencies but event tho this solved the issue with the typescript errors, new (illogical) `no-unsed-vars` errors emerged. Maybe we could fix this in the next PR so we don't have to `eslint-ignore` them all the time.

refi93:
ok, let's merge it and fix the remaining errors in another PR, I'm already working on the fix

PR #712, opened 2020-08-31
Fix failing unit tests

refi93:
Motivation: the pipeline on the development branch is currently failing on the unit tests

Changes:
* added mock data for new endpoints
* updated sample shelley wallet mnemonic to `abandon` x 14 + `address`
* update mock account data to reflect the account number of the shelley sample mnemonic
* fixed tx history test data (apparently, the backend started returning the fees instead of having them appended client-side to the returned data)

Testing: `yarn test-chrome-headless` is passing

Note: this is just the bare minimum to make the tests pass as groundwork for refactoring. The tests/codebase requires cleaning up from dead Byron code and we should also rethink the unit tests to be less cumbersome to maintain, as the mock data is hard to manage (need to explicitly paste and match requests) - either by sticking to 1-2 sample mnemonics and/or we should decouple fetching data from the app logic.


app/frontend/actions.ts
```diff
@@ -550,7 +550,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     let plan
     try {
       plan = await wallet.getTxPlan(args)
-      if (!plan.fee) {
+      if (!plan.fee && !plan.estimatedFee) {
```
PeterBenc, line 553: It seems to me, this would cause the `OutputTooSmall` error to never occur since if we cant build tx with big enough output, the function actually returns `estimatedFee`. 

I don't really see a reason to return the `estimatedFee` at all, since we don't use it anywhere. I am aware the currently implemented logic is not ideal since it was implemented in hurry few hours after the Shelley release and I am keeping it in mind for refactoring.

refi93: I see, sorry I misunderstood the intent of the original code - I left a TODO comment to refactor this and I agree the behavior right now is counter-intuitive.

I also fixed the testcase for fee computation (`test/src/transaction-actions.js`) by mocking an utxo for the wallet generated from the test mnemonic so it is actually able to build a valid tx and estimate the fees which was the reason why I extended this condition in the first place thinking it is just an omission

PeterBenc (approved):

Issue #627, opened 2020-06-15
Confusing UX in case of insufficient funds for a transaciton


refi93:
* attempting to send max funds on a wallet with too small of a balance results in "Invalid format: Amount has to be a positive number"

* possibly confusing UX when having too little funds to send a tx, we are currently distinguishing the case when you have too little funds to send any tx to the chosen address vs too little funds to send a specific tx with specific amount - that's unnecessary and can be merged into a single error

PeterBenc:
Resolved in #722

Issue #161, opened 2018-07-30
Refactor debugLog


jmuravsky:
Currently debugLog needs to be dirty patched to work with tests, because config is loaded from HTML body attribute, which is not present in test environment. Whether just debugLog or whole config handling should be refactored needs to be analysed and discussed.

refi93:
It also seems to be quite repetitive that before throwing errors, debugLog is being called each time. It would be nice to think about a way to remove this repetitiveness.

PeterBenc:
Closing this since the issue is resolved in PR #722. The PR removed all unnecessary debug logs, except those that are in byron files since they will be removed later.

PR #722, opened 2020-09-06
Refactor shelley functionality 1

PeterBenc:
This is one of many PRs refactoring the current Shelley codebase. This one refactors simplifies and fixes fee calculation and error handling when a delegating, sending ADA, converting, or withdrawing rewards. Then it fixes the bug with donations and TransactionCorrupted error and removes dead code and debug log. 

Addresses: 
#627 

- disables the max button if balance is 0
- the issue considered showing a specific error when there are not enough funds to send any tx useless, but I don't really see a reason to remove this, this error is could be shown for example if the user has 1.5 ADA and he tries to delegate for the first time, in that case, the fee is (fee + deposit) which is ~2.18 ADA, and this error message IMO makes sense in that case.

#161
- removes one useless debug log, 
- all the others debug logs are either in byron-related files that will be removed, or they are used in proper places and I don't really see a problem using debug log in 10 places. I am open to suggestions on how to decrease the number of used debug logs even more. 

Refactors:
- All 4 fee calculation functions, now they are a lot shorter, follow pretty much the same structure, and handle errors in better way and in one place.
- fee validation function, (renamed to `validateTxPlan`) and structures the other validations (for delegation etc. ) to extend this function, removing some duplicate code and ensuring similar logic when validating the tx plan. Please check the first commit for more details and review.

Fixes: 
- the `OutputAmountTooLow` was shown before any other error, so for example, if the user had 0 ADA and wanted to delegate, we showed him that `output amount is too low` which was not accurate. After the fix, this error is shown only if there is no other validation error.
- adds `stopLoadingAction` in `confirmTransaction` function, since when an error was thrown, user had to refresh the page to get rid of the loading.
- if user entered a donation which caused the `outputAmountTooLow` error, the donations were reset and fee set to 0 without showing any error.


app/frontend/helpers/validators.ts
```diff
@@ -73,33 +68,48 @@ const donationAmountValidator = (fieldValue, coins, balance) => {
...
+      code: 'DelegationBalanceError',
       params: {balance},
     }
   }
-  return null
+  const txPlanError = txPlanValidator(0, balance, txPlan)
+  return txPlanError || null
+}
+
+const withdrawalPlanValidator = (sendAmount, balance, txPlan) => {
```
xdzurman, line 106: I would rename sendAmount to rewardsAmount, since that's basically what being sent to the function and it fits more.

app/frontend/actions.ts
```diff
@@ -841,30 +808,15 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-      })
-    } catch (e) {
-      setErrorState('transactionSubmissionError', e, {
-        shouldShowTransactionErrorModal: true,
-      })
-      return
-    }
-    // const cryptoProviderVersionError = wallet.checkCryptoProviderVersion()
+    const plan = await prepareTxPlan({rewards, txType: 'redeem'})
     const redemptionValidationError =
```
xdzurman, line 812: Rename redemptionValidationError to withdrawalValidationError if you started renaming it in all other places to get rid of the term 'redeem' completely for consistency

xdzurman (changes requested):
Other than the requested changes - LGTM.

DavidTranDucVL (approved):
Beside comments from xdzurman LGTM

app/frontend/helpers/validators.ts
PeterBenc: Fixed

app/frontend/actions.ts
PeterBenc: Good point, fixed.

xdzurman (approved):

PR #723, opened 2020-09-07
Distinguish between failed invalid txs and server failures

xdzurman:


PeterBenc (approved):
LGTM, tested, there is an eslint error but that will be fixed after the rebase.

PR #727, opened 2020-09-09
Server sentry logging

PeterBenc:
New:
- Updates sentry to the newest version and moves sentry/browser to package.json in the app folder
- Sentry uncaught error capturing on the server side
- Explicit failed tx capturing in sentry
- Name, email and comment added to event tags on client side ( to be able to show these in slack)

FYI: audit errors are fixed in another PR which on this PR will be rebased.

### DIFF
--- a/server/emailSubmitter.js
+++ b/server/emailSubmitter.js
@@ -1,5 +1,6 @@
 require('isomorphic-fetch')
 const {backendConfig} = require('./helpers/loadConfig')
+const {captureException} = require('@sentry/node')
 
 module.exports = function(app, env) {
   app.post('/api/emails/submit', async (req, res) => {
@@ -49,8 +50,9 @@ module.exports = function(app, env) {
         Left: 'Email submission rejected by network',
       })
     } catch (err) {
+      captureException(err)
       return res.json({
-        Left: 'An unexpected error has happened',
+        Left: 'An unexpected error has occurred.',
       })
     }
   })
--- a/server/index.js
+++ b/server/index.js
@@ -8,6 +8,18 @@ const ipfilter = require('express-ipfilter').IpFilter
 const errorHandler = require('./middlewares/errorHandler')
 
 let app = express()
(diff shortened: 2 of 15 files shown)


server/emailSubmitter.js
```diff
@@ -49,8 +50,9 @@ module.exports = function(app, env) {
         Left: 'Email submission rejected by network',
       })
     } catch (err) {
+      captureException(err)
       return res.json({
-        Left: 'An unexpected error has happened',
+        Left: 'An unexpected error has occured.',
```
DavidTranDucVL, line 55: Typo: occurred

DavidTranDucVL (commented):
LGTM

server/emailSubmitter.js
PeterBenc: Thnx fixed.

DavidTranDucVL (approved):

server/index.js
```diff
@@ -8,6 +8,18 @@ const ipfilter = require('express-ipfilter').IpFilter
...
+
+Sentry.init({
+  dsn: 'https://43eac31915bb40caa03798a51048e756@o150853.ingest.sentry.io/5421403',
+  tracesSampleRate: 0,
+  beforeSend(event) {
+    return dropSensitiveEventData(event)
+  },
+})
+
+app.use(Sentry.Handlers.requestHandler())
```
xdzurman, line 22: does this have to be up here for any reason? If not, I'd move it all the way down to line 129 where "app.use(Sentry.Handlers.errorHandler())" is written so that Sentry stuff stays together

PeterBenc: Thnx, great question, there was a reason for this, checkout this [link](https://docs.sentry.io/platforms/node/guides/express/). Sentry should be initialized as soon as possible, thane the rest of the app and then the error handles, at least according to this.

xdzurman (approved):

PR #729, opened 2020-09-10
Integrate shelley ready cardano crypto.js library

PeterBenc:
Purpose of this PR is not to refactor current functionality but to integrate test pre-final shelley release of `cardano-crypto` 

New: 
- Imports the newest version of `cardano-crypto.js` (Shelley ready)
- adjusts currently used functionality to the new api
- removes local `bech32` helper and replaces it with cardano-crypto export
- gets address attributed for byron witnesses dynamically (not in ledger since ledger return only path and not address)

Fixes: 
 - an address derivation test
- audit errors
- removes lodash and fixes babel resolution

Ideas for refactoring:
- we could use `addresses.ts` as an extension layer over cardano-crypto.js` and import all address related functionality from there.


PeterBenc:
@roccomuso you are right :+1: thanks for the review :+1:

app/tests/src/shelley.js
```diff
@@ -24,9 +24,8 @@ describe('shelley address derivation', () => {
   const mnemonic15Words =
     'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon address'
   it('should derive base address from 15-words mnemonic', async () => {
-    const cp = await getCryptoProvider(mnemonic15Words, 'mainnet') //TODO: change discriminator to networkId for shelley
+    const cp = await getCryptoProvider(mnemonic15Words, 0) //TODO: change discriminator to networkId for shelley
```
refi93, line 27: Is that `0` the mainnet network id? If so, better reference a named constant here to minimize confusion

refi93:
I published the latest release candiate as version 6.0.0, you can update the import of cardano-crypto.js

PR #735, opened 2020-09-28
Fix passing of network params to displayAddressForPath for hw wallets

refi93:
Problem: Trezor was not passing protocolMagic to the cardanoGetAddress call, resulting in it being None. This was not an issue until now because it negated itself with a bug in trezor-connect resolved here:

Changes:
* add passing of protocolMagic to the cardanoGetAddress call for trezor crypto provider
* refactor the displayAddressForPath call to pass the proper network parameters for both Ledger and Trezor (the nework id was hardcoded to `0` until now even though the crypto provider received the parameter from outside)

Testing/validation: tested address verification in Adalite on localhost for both Ledger and Trezor and works as expected

### DIFF
--- a/app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
+++ b/app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
@@ -1,6 +1,6 @@
 import LedgerTransportU2F from '@ledgerhq/hw-transport-u2f'
 import LedgerTransportWebusb from '@ledgerhq/hw-transport-webusb'
-import Ledger from '@cardano-foundation/ledgerjs-hw-app-cardano'
+import Ledger, {AddressTypeNibbles} from '@cardano-foundation/ledgerjs-hw-app-cardano'
 import {encode} from 'borc'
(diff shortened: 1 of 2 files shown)


refi93:
depends on https://github.com/vacuumlabs/adalite/pull/736 (to make CI pass)

PeterBenc (approved):
Thanks for taking the time to fix this. As you mentioned the values were hardcoded since it was implemented in rush ~2 hours before shelley release.. :D I was thinking we could extend the function in `shelley-wallet.ts` and pass also the address buffer to cryptoProvider and get the `networkId` `protocolMagic` and `addressNibble` from the address buffer. Now I am bit confused tho, regarding the trezor implementation. 
`cardanoGetAddress` function receives `addressParameters` ( `addressType` and `paths`) and then `networkId`, `protocolMagic` separatelly. Arent `networkId` and  `protocolMagic` also considered address parameters thus should be retrieved from the address (if present)? Im not sure how specifically trezor/ledger uses these values but it seems we could get them from the the address we are verifying.

PR #738, opened 2020-11-03
Show rewards in Staking history and CSV

DavidTranDucVL:
# UI
New:
- Show rewards in staking history
- Staking rewards from each epoch that are added to the reward address are shown in green, indicating **_credit_**

Changes:
- Changed reward withdrawals color in staking history to blue
- Minor changes:
  - `Copy pool id icon` changed to icon that redirects user directly to the cardano scan
  - Changed capitalization of some labels

`Transaction history` remains unchanged and matching withdrawal transaction is green as **_credit_** and the `amount+fee` should be equal to the `amount` from the matching blue **_withdraw_** entry in `Staking history`
- Actually, this is only true when transferring funds from owners reward address to owners payment address, which is the only way to do it trough AdaLite. If somebody creates transaction with withdrawal that contains output to non-owner's address, this amount adds to the difference.

# CSV
New:
- Show staking rewards from each epoch in CSV as `Reward awarded` type

Changes:
- Change transactions in CSV that are associated with withdrawals - they are always classified as `Sent` type and the `Sent amount` equals to the amount of currency sent to the somebody else payment address (in most cases it will be 0, since withdrawals are usually made to the user's payment address)

This way accountants:
- can see clearly when rewards were added
- can see when the fee was applied when user decided to move funds from reward address to payment address
- can see more useful information by just summing the amount columns


xdzurman (commented):
Something throws me off about the idea of having reward reception in green and withdrawal in red, since both are incomes but to different addresses. I'd experiment with shades of green or even a different color, e.g. blue for reward reception and original green for withdrawal.

app/frontend/components/pages/txHistory/transactionHistory.tsx
```diff
@@ -74,52 +75,93 @@ interface Props {
...
+    rewardWithdrawal: 'Reward withdrawal',
+  }
+
+  const transactionsEntries = transactionHistory.map((transaction) => {
+    const common = {
+      txHash: transaction.ctbId,
+      dateTime: moment.utc(new Date(transaction.ctbTimeIssued * 1000)),
+    }
+
+    if (withdrawalHistoryTxHashes.includes(transaction.ctbId)) {
```
xdzurman, line 117: It's not a problem with a small number of withdrawals, but their count can get ramped up over time. I would therefore make withdrawalHistoryTxHashes into a hash table, so we have o(1) lookups here for each transaction

DavidTranDucVL: Fixed

DavidTranDucVL:
> Something throws me off about the idea of having reward reception in green and withdrawal in red, since both are incomes but to different addresses. I understand that you meant it on the basis of ins and outs from rewards account but I'm not sure if it's as transparent for an average user. I'd experiment with shades of green or even a different color, e.g. blue for reward reception and original green for withdrawal. Please consult with Michal about this if you didnt already

We went over details with Michal today and we made some further changes. I updated PR description accordingly.

xdzurman (approved):
LGTM

PeterBenc:
@xdzurman I was just checking in the exported CSV whether `sum(received) - sum(send) - sum(fee) = balance` which it should, but surprisingly it doesn't. It seems the testnet rewards have no record od being distributed thus causing the balance the CSV equaling `"real balance" - testnet rewards`. Is there a way to get a record of testnet rewards distribution?

xdzurman:
> @xdzurman I was just checking in the exported CSV whether `sum(received) - sum(send) - sum(fee) = balance` which it should, but surprisingly it doesn't. It seems the testnet rewards have no record od being distributed thus causing the balance in the CSV equaling `"real balance" - testnet rewards`. Is there a way to get a record of testnet rewards distribution?

Sure, it's a whole different db table since it was meant for a one time use, but I can add that to rewardHistory manually, the current format specifies for which epoch the reward has been distributed, but we can put a '-1' or 'ITN' in it to distinguish between normal rewards and this one time ITN reward

PeterBenc:
> > @xdzurman I was just checking in the exported CSV whether `sum(received) - sum(send) - sum(fee) = balance` which it should, but surprisingly it doesn't. It seems the testnet rewards have no record od being distributed thus causing the balance in the CSV equaling `"real balance" - testnet rewards`. Is there a way to get a record of testnet rewards distribution?
> 
> Sure, it's a whole different db table since it was meant for a one time use, but I can append that to rewardHistory manually, the current format specifies for which epoch the reward has been distributed, but we can put a '-1' or 'ITN' in it to distinguish between normal rewards and this one time ITN reward

Amazing, that would be great, I don't remember when, but there was a special epoch or time from which on, the testnet rewards were available, (maybe the start of shelley era I am not sure), if we want to precise, we could find out what was the date (or epoch).

xdzurman:
> > > @xdzurman I was just checking in the exported CSV whether `sum(received) - sum(send) - sum(fee) = balance` which it should, but surprisingly it doesn't. It seems the testnet rewards have no record od being distributed thus causing the balance in the CSV equaling `"real balance" - testnet rewards`. Is there a way to get a record of testnet rewards distribution?
> > 
> > 
> > Sure, it's a whole different db table since it was meant for a one time use, but I can append that to rewardHistory manually, the current format specifies for which epoch the reward has been distributed, but we can put a '-1' or 'ITN' in it to distinguish between normal rewards and this one time ITN reward
> 
> Amazing, that would be great, I don't remember when, but there was a special epoch or time from which on, the testnet rewards were available, (maybe the start of shelley era I am not sure), if we want to precise, we could find out what was the date (or epoch).

In the database, it says the rewards were "distributed" in the epoch 208, 30.7.2020, ranging from  17:28:51 UTC to 01:23:11, 31.7.. We can use this as the "distributing" time

PeterBenc (approved):

PR #743, opened 2020-11-12
Tx wallet type stats and premium banner

PeterBenc:
Adds stats for wallet type when submitting transactions.
Adds premium banner for users with stake higher than 2.5 ml ADA.


app/frontend/components/pages/dashboard/premiumBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {connect} from '../../../helpers/connect'
+import actions from '../../../actions'
+
+interface Props {
+  closePremiumBanner: () => void
+}
+
+const PremiumBanner = ({closePremiumBanner}: Props) => {
+  const premiumMessage =
+    'VIP STAKING SERVICES - Do you own more than 3 million ADA? Contact us on  adalite@vacuumlabs.com to join our VIP staking services. We will assign you a dedicated account manager and provide you with priority support.'
```
xdzurman, line 11: I think 3 million should not be hardcoded here, since its the current PREMIUM_MEMBER_BALANCE_TRESHHOLD says 2.5m

app/frontend/components/pages/dashboard/premiumBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div className="banner premium">
+      <div className="banner-text">{premiumMessage}</div>
+      <button
+        className="button close banner-close"
+        {
+        ...{ariaLabel: 'Close banner'} /* silence ts*/
+        }
```
xdzurman, line 19: what does this do?

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -204,13 +204,14 @@ const ShelleyWallet = ({
...
   }
 
-  function getHwWalletName() {
-    return isHwWallet ? (cryptoProvider as any).getHwWalletName() : undefined
+  function getWalletName() {
+    return cryptoProvider.getWalletName()
   }
 
   function submitTx(signedTx): Promise<any> {
+    const params = {walletType: getWalletName()}
```
xdzurman, line 212: Do you think we could also pass user's staking key? It would allow much easier support

app/frontend/components/pages/dashboard/premiumBanner.tsx
PeterBenc: Yeah I noticed the half million as well but I have to confirm this with Michael whether this is intentional or a mistake. Tho I dont see a reason to not-hardcode the valuer here, the banner might change in future, and I m not sure how would I convert 2500000000000 to something I could into the text.

app/frontend/wallet/shelley-wallet.ts
PeterBenc: I ll make this in another PR :+1:

app/frontend/components/pages/dashboard/premiumBanner.tsx
PeterBenc: Good question, I was just copy-paste-ing this so I have no idea, but it seems the previous version of typescript has some problem with this, and this was a hotfix. I ll remove it from where it's useless. :+1:

app/frontend/components/pages/dashboard/premiumBanner.tsx
PeterBenc: The difference was intentional indeed.

app/frontend/components/pages/dashboard/premiumBanner.tsx
PeterBenc: [This link](https://www.aditus.io/aria/aria-label/#:~:text=Over%2Dusing%20it-,Definition,content%20as%20the%20accessibility%20label.
) explains the usage of this attribute, so Im gonna let it be as it is.

xdzurman: I was referencing the style it's written, does 
```
<button
        className="button close banner-close"
        ariaLabel="Close banner"
        onClick={(e) => {
          closePremiumBanner()
        }}
/>
```

not work?

PeterBenc: Exactly, that doesn't work a that's the reason for that comment.

xdzurman (approved):

PR #753, opened 2020-12-02
Add description to deposit in delegation review

DavidTranDucVL:
Preview:
image


xdzurman (approved):
Perhaps I would note that this deposit is returned upon stake deregistration, but since we don't have that implemented yet, I'm not sure if that's the best idea.

PR #767, opened 2020-12-15
Multi-account support

PeterBenc:
**Motivation:** Users holding a big stake, or users wanting to delegate to multiple pools can do this now using multiple accounts. Furthermore, adds a dashboard displaying all balances, rewards, and delegation for every account. The dashboard can be also used to easily move funds from one account to another and delegate. Users using this feature might also want to export their public keys in bulk so they don't have to export them one by one.

**TODO:**
This PR first of all separates wallet logic from account logic. Regarding this, the account and wallet could be separated even more but would need a further refactor.
Add more types, and refactor state.ts

**Commits:**

- Fix alert margin in news modal
fixes margin in news modal on mobile devices

- Add advanced tab
refactor dashboard page so it works with accounts tab
add keys component to advanced tab

- Show public account keys
add support for getting account public key hex and show ti on keys component

- Refactor dashboard and fix tabs
further refactor dashboard page

- Init multiple account support
represent accounts as new instances of shelley-wallet, this was later refactored so it makes more sense,
adds initial version of account tab and moves some of the account functionality to account.ts

- Remove blochain explorer from wallet
removes blockchain explorer functonality from shelley wallet to blockchain-explorer

- Refactor - rename wallet to account
renames wallet to account, removes shelley-wallet file altogether and moves its functionality to account.ts
removes support for byron wallet in actions.ts

- Remove wallet functionality from account
tries to move the most of the functionality that should be responsibility of wallet and not account to wallet, 
TODO: the two should be separated event more but needs refactor ( future PRs)

- Make account part of wallet
moves all accounts to wallet, fixes logic for loading and selecting a new account

- Discover used accounts
implements logic for discovering account base on whether they we re used

- Bulk export public keys for accounts
most important part in `CachedDeriveXpubFactory.ts` 
  the behavior is changes so either one account public could be exported and saved to cache or of an account is not discovered yet, it tries to export range of public keys and cache them at once, byron public key is in that case appended to the first export
TODO: move MAX_BULK_EXPORT_AMOUNT to .env variable as well as ENABLE_BULK_EXPORT

- Fix rebase issues

- Update ledger lib and fix Ledger backwards functionality

- Add accounts tab design
couple of squashed commits regarding the accounts tab design
adds transaction modal and action buttons to accounts

- Send and delegate from correct account
account has to be briefly changed to build tx, sign it and then be changed back
introduces new state, that represent the account index from which from is the tx built, 
activeAccount => the one that is used to do operations
selectedAccount => the one that user selected last

- Change accounts correctly and refactor

- Fix tooltips and remove buttons on inactive accounts
remove tooltips from buttons and remove buttons from inactive accounts

- Update ledger lib

- Add address verification for funds transfer modal
adds a possibility to verify address on sendAda component before sending it from one account to another

- Add label for delegation input field
addss a prefix to delegation pool id field so it is obvious what the field is for

- Fix address verification
the right account has to be used to verify the address, in case of "Send to" the account that verifies address is the active account, in case of "SendFrom" the selectedAccount verifies the address, 
this intorduces another variable `targerAccount` which determines to which account the funds are going to

- Fix failing tests
fix tests regarding changed export public key functionality and remove one byron related test

- Fix rebase errors

- Refactor total balance calculation
move total balance calculation to wallet

- Refactor bulk pubkey export names
refactor confusing variable names regarding bulk public key export

- Refactor selectedMainTab
previously, there we only 2 tabs, so a boolean variable was suitable, now that we have 4, the type of variable was changed to string and also its name to make more sense

- Center items in accounts tab


app/frontend/actions.ts
```diff
@@ -175,11 +163,15 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
       const autoLogin = state.autoLogin
       const ticker2Id = null
       const shouldShowPremiumBanner =
-        state.shouldShowPremiumBanner && walletInfo.balance > PREMIUM_MEMBER_BALANCE_TRESHHOLD
-      const isBigDelegator = walletInfo.balance > BIG_DELEGATOR_THRESHOLD
+        state.shouldShowPremiumBanner && PREMIUM_MEMBER_BALANCE_TRESHHOLD < totalWalletBalance
```
refi93, line 166: typo - TRESHHOLD -> TRESHOLD

app/frontend/actions.ts
```diff
@@ -280,7 +281,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
   }
 
   const logout = () => {
-    wallet = null
+    account = null
```
refi93, line 284: I see that the `wallet` variable is still in use, shouldn't that be set to null as well?

app/frontend/actions.ts
```diff
@@ -985,6 +993,101 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+        ...state.accounts,
+        [accountIndex]: walletInfo,
+      },
+    })
+  }
+
+  const setAccount = async (
+    state: State,
+    accountIndex: number,
+    shoudlChangeSelectedAccount?: boolean
```
refi93, line 1012: typo shoudl -> should

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+const Account = ({
+  i,
+  account,
+  firstAddressPerAccount,
+  setAccount,
+  selectedAccount,
+  showDelegationModal,
+  showSendTransactionModal,
+}) => {
+  const isSelected = selectedAccount === i
```
refi93, line 22: I would rather name the variable `selectedAccountIndex` so it's clear it's not an Account object

app/frontend/actions.ts
```diff
@@ -175,11 +163,15 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+        state.shouldShowPremiumBanner && PREMIUM_MEMBER_BALANCE_TRESHHOLD < totalWalletBalance
+      const isBigDelegator = totalWalletBalance > BIG_DELEGATOR_THRESHOLD
       setState({
+        accounts: accountsInfo,
+        totalWalletBalance,
+        totalRewardsBalance,
+        shouldShowSaturatedBanner,
         walletIsLoaded: true,
-        ...walletInfo,
+        ...accountsInfo[0],
```
refi93, line 174: I'd rather pass the full accountsInfo and have the current account index set in one specific property of the state to prevent the scenario that some parts of the state are associated with one account and some with another

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+import range from '../../../../frontend/wallet/helpers/range'
+import printAda from '../../../helpers/printAda'
+import {Lovelace} from '../../../state'
+import {AdaIcon} from '../../common/svg'
+import Alert from '../../common/alert'
+import SendTransactionModal from './sendTransactionModal'
+import DelegationModal from './delegationModal'
+import tooltip from '../../common/tooltip'
+
+const Account = ({
```
refi93, line 13: I would name this AccountTile or something like that, to distinguish it from the account object

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+  i,
```
refi93, line 14: this seems weird - why do you need `i`, `selectedAccount` and `account` and what's the difference between them three? Isn't the account index contained within the `account` object?

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+  account,
+  firstAddressPerAccount,
```
refi93, line 16: I find this a bit cumbersome TBH, I would just pass the full list of accounts, the tile's account index, selected account index and do the address retrieval/transformations needed inside, removing the concern of constructing a mapping of accounts to addresses by the outside logic

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+          <DelegateButton />
+        </div>
+      ) : (
+        <div className="account-action-buttons desktop" style="width: 94px;" />
+      )}
+    </div>
+  )
+}
+
+const Accounts = ({
```
refi93, line 171: AccountsDashboard perhaps?

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+  reloadWalletInfo,
+  showSendTransactionModal,
+  showDelegationModal,
+  shouldShowSendTransactionModal,
+  shouldShowDelegationModal,
+  selectedAccount,
+  totalWalletBalance,
+  totalRewardsBalance,
+}) => {
+  const accountInfos = Object.values(accounts)
```
refi93, line 183: this seems fragile to me - I wouldn't take for granted that the values are always returned in ascending order, e.g. in this SO thread are contradicting answers on this: https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+        </div>
+        <div className="mobile">
+          <InfoAlert />
+        </div>
+        <div className="accounts-wrapper">
+          <div className="dashboard-column account list">
+            <div>
+              {range(0, Object.keys(accounts).length + 1).map(
+                (i) =>
+                  (i === 0 || accounts[i - 1].isUsed) && (
```
refi93, line 251: shouldn't we just break the loop as soon as we come across the first unused account? Otherwise we wouldn't be wouldn't be BIP44 compliant: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery in the sense that we shouldn't let the users use two non-consecutive accounts if no account in between is used as the discovery is supposed to stop as soon as the currently explored account is unused

app/frontend/components/pages/accounts/accounts.tsx
```diff
@@ -0,0 +1,286 @@
...
+            </button>
+          </div>
+        </div>
+        <div className="mobile">
+          <InfoAlert />
+        </div>
+        <div className="accounts-wrapper">
+          <div className="dashboard-column account list">
+            <div>
+              {range(0, Object.keys(accounts).length + 1).map(
```
refi93, line 249: why the +1? Is there the assumption that the last element of accounts is always unused?

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -98,38 +123,35 @@ class DashboardPage extends Component<Props> {
...
-        </div>
+        <div className="dashboard desktop">{displayedPages[selectedMainTab]}</div>
 
         <div className="dashboard mobile">
-          {displayStakingPage ? <ShelleyBalances /> : <Balance />}
-          <DashboardMobileContent displayStakingPage={displayStakingPage} />
-          {!displayStakingPage && shouldShowExportOption && <ExportCard />}
+          {displayedSubPages[selectedMainTab]}
+          <DashboardMobileContent selectedMainTab={selectedMainTab} />
+          {selectedMainTab === 'Sending' && shouldShowExportOption && <ExportCard />}
```
refi93, line 136: 'Sending' should be a constant, otherwise if we rename the tab, it may easily break, if I understand correctly

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -138,21 +160,35 @@ class DashboardMobileContent extends Component<Props> {
...
   }
-  render({displayStakingPage}, {selectedSubTab}) {
-    const stakingTabs = ['Delegate ADA', 'Current Delegation', 'Staking history']
-    const sendingTabs = ['Send ADA', 'Transactions', 'Recieve ADA']
-    if (displayStakingPage && sendingTabs.includes(selectedSubTab)) {
-      this.selectSubTab('Delegate ADA')
+  // TODO: refactor
+  accountsTabs = ['Accounts']
+  stakingTabs = ['Delegate ADA', 'Current Delegation', 'Staking history']
+  sendingTabs = ['Send ADA', 'Transactions', 'Recieve ADA']
```
refi93, line 169: typo - Recieve -> Receive

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -130,95 +134,120 @@ class Delegate extends Component<Props> {
...
+            className="button primary"
+            disabled={
+              !isShelleyCompatible ||
+              validationError ||
+              calculatingDelegationFee ||
+              stakePool.poolHash === ''
+            }
+            onClick={delegationHandler}
+            {...tooltip(
+              'You are using Shelley incompatible wallet. To delegate your ADA, follow the instructions to convert you wallet.',
```
refi93, line 194: typo "convert you wallet" -> "convert your wallet"

app/frontend/components/pages/login/hardwareAuth.tsx
```diff
@@ -1,14 +1,29 @@
...
+import actions from '../../../actions'
+import {useState, useCallback} from 'preact/hooks'
+import {localStorageVars} from '../../../localStorage'
 
 interface Props {
   loadWallet: any
 }
 
 const LoadByHardwareWalletSection = ({loadWallet}: Props) => {
+  const [dontBulkExportPubKeys, setBulkExport] = useState(
```
refi93, line 16: I'd prefer to name it something like `disableBulkPublicKeyExport`

app/frontend/components/pages/login/hardwareAuth.tsx
```diff
@@ -22,70 +37,100 @@ const LoadByHardwareWalletSection = ({loadWallet}: Props) => {
...
-          onClick={() =>
-            loadWallet({cryptoProviderType: CRYPTO_PROVIDER_TYPES.LEDGER, isWebUSB: true})
-          }
-        >
-          Connect with WebUSB
-        </button>
+          <button
+            {...tooltip(
+              'Support for Ledger is temporarily disabled',
+              !ADALITE_CONFIG.ADALITE_ENABLE_LEDGER
```
refi93, line 89: I'd get rid of this flag, it's no longer relevant

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,13 +37,43 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
     const lastIndex = derivationPath.slice(-1)[0]
     const parentXpub = await deriveXpub(derivationPath.slice(0, -1))
-
     return deriveChildXpub(parentXpub, lastIndex, derivationScheme.ed25519Mode)
   }
 
+  function createPathBulk(derivationPath: BIP32Path): BIP32Path[] {
+    const paths: BIP32Path[] = []
+    const accountIndex = derivationPath[2] - HARDENED_THRESHOLD
+    const baseAccountIndex = accountIndex - (accountIndex % MAX_BULK_EXPORT_AMOUNT)
```
refi93, line 49: I'd rather introduce the abstraction of "currentAccountPage" which would be the index of the current "batch" of addresses and it would be accountIndex / MAX_BULK_EXPORT_AMOUNT

app/frontend/actions.ts
```diff
@@ -985,6 +993,101 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     updateStakePoolIdentifier(getState(), null, state.poolRecommendation.recommendedPoolHash)
   }
 
+  /* MULTIPLE ACCOUNTS */
+
+  const loadNewAccount = async (state: State, accountIndex: number) => {
```
refi93, line 998: does it always have to be a "new" account? wouldn't `loadAccount` name be more appropriate?

app/tests/src/actions/wallet-actions.js
```diff
@@ -75,28 +75,3 @@ it('Should properly load shelley wallet', async () => {
 
   mockNet.clean()
 })
-
-it('Should properly load byron wallet', async () => {
```
refi93, line 79: this test didn't pass anymore I guess, but that means that corresponding tests for the multi-account feature need to be added to ensure the logic does not get broken over time - it's critical to derive the accounts right, otherwise the users may lose funds

refi93 (changes requested):
see comments

app/frontend/actions.ts
PeterBenc: Thnx, fixed...

app/frontend/actions.ts
PeterBenc: Fixed.

app/frontend/actions.ts
PeterBenc: Removed.

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: You are right :+1: , fixed.

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: Agreed. :+1:

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: Fixed.

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: I agree, I find it cumbersome as well, I forgot to remove this `address` argument and move it to actions.ts. Fixed.

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: Agreed :+1:

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: Removed.

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: We are BIPP 44 compliant, the current behavior is the exact as you explained. Let's say a wallet has 3 used accounts, in that case, we want to display 4 "AccountTiles" the last one to be explored if needed. If a user explores this account we would still display 4 account tiles until the user doesn't use the 4th account. But I agree this doesn't look very good. I refactored so we calculate the number of user accounts in advance.

```
{range(0, usedAccountsCount + 1).map(
                (i) =>
                  (

File: app/frontend/components/pages/accounts/accounts.tsx (hunk 12 above)
Comment (PeterBenc) [reply app/frontend/components/pages/accounts/accounts.tsx]: We want to display (numberOfUsedAccounts + 1) accountTiles. Refactored.

File: app/frontend/components/pages/dashboard/dashboardPage.tsx (hunk 13 above)
Comment (PeterBenc) [reply app/frontend/components/pages/dashboard/dashboardPage.tsx:136]: I agree I have some ideas about refactoring the dashboard page since it has too many constants and is cumbersome to work with, Would you mind if I made it into another PR tho?

File: app/frontend/components/pages/dashboard/dashboardPage.tsx (hunk 14 above)
Comment (PeterBenc) [reply app/frontend/components/pages/dashboard/dashboardPage.tsx]: Fixed.

File: app/frontend/components/pages/delegations/delegatePage.tsx (hunk 15 above)
Comment (PeterBenc) [reply app/frontend/components/pages/delegations/delegatePage.tsx]: Thnx fixed.

File: app/frontend/components/pages/login/hardwareAuth.tsx (hunk 16 above)
Comment (PeterBenc) [reply app/frontend/components/pages/login/hardwareAuth.tsx]: I used negation in localstorage variable name sucne by default `window.localStorage.getItem(localStorageVars.BULK_EXPORT) === 'true'` is `false`. Anyway, you are right, a negated the expression and now it makes more sense. Thanks. 

```
  const [enableBulkExport, setBulkExport] = useState(
    window.localStorage.getItem(localStorageVars.BULK_EXPORT) !== 'true'
  )
  const toggleBulkExport = useCallback(
    () => {
      window.localStorage.setItem(localStorageVars.BULK_EXPORT, `${enableBulkExport}`)
      setBulkExport(!enableBulkExport)
    },
    [enableBulkExport]
  )

app/frontend/actions.ts
PeterBenc: It would :+1:

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: Thanks, great idea. :+1:

app/frontend/actions.ts
PeterBenc: I agree, previously I implemented the multi-account support as quickly as possible and this is definitely one of the early designs which enabled me to implement this quickly without a lot of changes. I wanted to left some state-related refactoring to another PR and this would be probably a part of it but the fact that we changed the design and using the transfer modal a user can switch "source" account rather quickly, it makes a lot more sense to do it right now.

An alternative would be to make sure we don't use the "general" state for tx calculation etc and only use it for displaying (in which case we don't care that much about how quickly it changes). When using something from the state for txs (like balance) we should always refer to it like `state.accounts[account.accountIndex].balance)` . (`account` is a global variable that changes synchronously so its even better). 

Also implementing both of these approaches would be nice. We can discuss this tomorrow and I ll implement at least one of them :+1:

app/tests/src/actions/wallet-actions.js
PeterBenc: Yes, it didn't pass anymore since it was using the byron-wallet that has no notion of accounts so I removed it. And of course, tomorrow I 'll add tests for the multi-account support.

app/frontend/components/pages/dashboard/dashboardPage.tsx
refi93: no, as the usage of hardcoded strings seems widespread, it seems sensible to deal with this separately - worth creating an issue though so we don't forget about it even if priorities shift after merging this

app/frontend/actions.ts
refi93: I'm not sure if I get your second approach - if we use the "general" state when displaying, wouldn't a "general" state out of sync potentially mislead the user into thinking the balance of their account is different than it actually is? And wouldn't it be confusing if the funds got send from a different account than the one they see?

app/frontend/components/pages/accounts/accountsDashboard.tsx
```diff
@@ -0,0 +1,255 @@
...
+            This feature might not be supported on other wallets yet. If you decide to move your
+            funds to <strong>account</strong> other then <strong>account 0</strong>, you might not
+            see your funds in other wallets.
+          </p>
+        </Alert>
+      </div>
+    </Fragment>
+  )
+
+  const usedAccountsCount = Object.values(accounts).reduce(
```
refi93, line 183: `Object.values(accounts).filter((account) => account.isUsed).length` seems more straightforward

app/frontend/actions.ts
PeterBenc: I still think we should do the approach you suggested and even better would be to do both. I ll implement your approach tomorrow. But to answer your question, the more important thing is to use the correct state when creating txs thus never use the "general" state in those cases. But in the end, the general state is also almost up to date. 

TLDR: we should use `state.account[account.accountIndex]` which changes synchronously when using state for building tx a and we don't really need to worry about using "general" state everywhere else since it's only milliseconds late. This would be a middle way between (for example) refactoring everything from `state.txHistory` to `state. accounts[state.selectedAccount].txHistory`and doing nothing. :D

app/frontend/components/pages/accounts/accounts.tsx
refi93: hmm, `usedAccountsCount` seems to be used only here and if we are really BIP44 compliant, it should always be one less than the total number of accounts shown to the user. Wouldn't it make sense to just iterate over all the accounts loaded into the state?

Also, I'm considering whether it's actually a good idea to have as an `accounts` object and with its keys being their hardened indices. It seems more natural to me to just have an array of accounts and not rely on `Object.values` and the index can just be it's unhardened value (0, 1 etc) which in turn is more readable when debugging, etc

PeterBenc:
Just wanted to mention, we should refactor a lot of state-management-related logic, mainly resetting fields, etc because it's really hard to work with. 

Btw. the latest version also with the switch button (I ll look for a better svg because I don't like this one that much tomorrow) is [here](https://adalite-staging-testnet.herokuapp.com/)

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,118 @@
...
+  const accounts: Array<ReturnType<typeof Account>> = []
+
+  function loadAccount(accountIndex: number) {
+    const newAccount = Account({
+      config,
+      cryptoProvider,
+      blockchainExplorer,
+      accountIndex,
+    })
+    accounts.push(newAccount)
```
refi93, line 16: this can easily result in accidentally loading the same account repeatedly and it also heavily relies on the accounts being loaded in the right order - either `accounts` should be an object and deduplication/ordering should be ensured by writing to the keys, or there should be a safeguard that would remove old instances of the same account from the list and sort them accordingly - I find it a bit more fool-proof to just use an object

app/frontend/components/pages/accounts/accounts.tsx
PeterBenc: You removed the part regarding the accounts being an object and not an array, I was planning to refactor it but didnt get to it today. 

Regarding the usedAccountCount. `it should always be one less than the total number of accounts shown to the user.`, well that's exactly how it is now. There is a difference between the number of accounts in the state and the number of accounts used. We can call about this tomorrow.

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,118 @@
...
+    let accountIndex = 0
+    while (shouldExplore && !accounts[accountIndex]) {
+      const newAccount = Account({
+        config,
+        cryptoProvider,
+        blockchainExplorer,
+        accountIndex,
+      })
+      const isAccountUsed = await newAccount.isAccountUsed()
+      if (isAccountUsed || accountIndex === 0) accounts.push(newAccount)
```
refi93, line 30: why not just call loadAccount(index) repeatedly until the last account in the accounts list is unused/index is equal to 100? Wouldn't that yield the same result, or am I missing some nuance here?

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -108,18 +126,33 @@ class SendAdaPage extends Component<Props> {
...
+            onInput={updateAddress}
+            autoComplete="off"
+            onKeyDown={(e) => e.key === 'Enter' && this.amountField.focus()}
+            disabled={isModal}
+          />
+        )}
+        {isModal && (
+          <div className="send-values dropdowns">
+            <label className="account-label">From</label>
+            <AccountDropdown accountIndex={sourceAccountIndex} setAccountFunc={setSourceAccount} />
```
refi93, line 147: shouldn't this element be getting the list of accounts as well?

app/frontend/components/pages/accounts/accountDropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+    [shouldHideAccountDropdown]
+  )
+
+  return (
+    <div className="account-dropdown">
+      <button className="account-dropdown-button" onClick={() => toggleAccountDropdown()}>
+        Account {accountIndex}
+      </button>
+      <div className={`account-dropdown-content ${shouldHideAccountDropdown ? 'hide' : 'show'}`}>
+        {range(0, Object.values(accounts).length).map((i) => (
```
refi93, line 24: looking at wallet.ts it seems to me like accounts have been refactored to a list

app/frontend/actions.ts
refi93: I'm not really worried about a delay between the global state and the local account state, but rather a potential bug that may cause "artifacts" from a previously selected account to persist in the state and be shown to the user for a different account. This would be particularly nasty as it would manifest only when switching accounts so users may report very weird problems until we figure out what exactly is going on

app/frontend/actions.ts
```diff
@@ -982,7 +1000,129 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+    setState({
+      accounts: {
+        ...state.accounts,
+        [accountIndex]: accountInfo,
+      },
+    })
+  }
+
+  const setAccount = async (state: State, accountIndex: number) => {
+    if (!wallet.accounts[accountIndex]) {
```
refi93, line 1022: is this supposed to happen? Shouldn't the accounts be already loaded when setting them?

app/frontend/actions.ts
PeterBenc: I agree and I m not aware of any blocker so I'll implement your approach I just wanted to point out the second one and why it beneficial as well.

app/frontend/components/pages/sendAda/sendAdaPage.tsx
PeterBenc: AccountDropdown connects to the state and gets accounts from there, so I can make it so `sendAdaPage` gets accounts from state just to pass it here as props...but do you think it makes more sense?

app/frontend/components/pages/accounts/accountDropdown.tsx
PeterBenc: I think accounts is still an object.

app/frontend/components/pages/accounts/accountsDashboard.tsx
```diff
@@ -0,0 +1,296 @@
...
+            </button>
+          </div>
+        </div>
+        <div className="mobile">
+          <InfoAlert />
+        </div>
+        <div className="accounts-wrapper">
+          <div className="dashboard-column account list">
+            <div>
+              {range(0, usedAccountsCount + 1).map((accountIndex) => (
```
refi93, line 246: Looking at usedAccountsCount definition above:

```
const usedAccountsCount = accountsInfo.filter(({isUsed}) => isUsed).length
```

why not just do here:
```
{range(0, usedAccountsCount).map((accountIndex) => (
```
?

Because supposing this code does not crash, accountsInfo must contain even the unused account index as there is the line `account={accountsInfo[accountIndex]}` - doesn't it?

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  return {
+    isHwWallet,
+    getWalletName,
+    submitTx,
+    getWalletSecretDef,
+    fetchTxInfo,
+    checkCryptoProviderVersion,
+    accounts,
+    discoverNewAccount,
```
refi93, line 106: is this function really meant to be exported? If I understand correctly, whoever would call it from outside can mess up the wallet quite badly

app/frontend/components/pages/accounts/accountsDashboard.tsx
```diff
@@ -0,0 +1,296 @@
...
+  selectedAccountIndex,
+  showDelegationModal,
+  showSendTransactionModal,
+  setLastAccountExplored,
+  isLastAccountExplored,
+}) => {
+  const isSelected = selectedAccountIndex === accountIndex
+
+  const shouldShowAccountInfo =
+    account && (account.isUsed || isLastAccountExplored || accountIndex === 0)
```
refi93, line 29: under which conditions is the account undefined? Is it when bulk key export is not allowed?

app/frontend/components/pages/accounts/accountsDashboard.tsx
refi93: If the concern are non-explored accounts when bulk key export is disabled, isn't it easier to deal in this while cycle with all accounts - used or not, but explored (i.e. present in accountInfos, if I understand correctly)

and then, if and only if the last explored account is used (i.e. the discovery was not done fully), add one more account tile with the index increased by one and all account info hidden and showing the "explore" button? And when you click the "explore" button, the account gets loaded into accounts info and the AccountsDashboard should be re-rended and if the newly explored account is unused, the additional tile just won't be rendered

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  async function discoverAccounts() {
+    let shouldExplore = true
+    let accountIndex = 0
+    while (shouldExplore) {
+      const newAccount = accounts[accountIndex] || discoverNewAccount()
+      const isAccountUsed = await newAccount.isAccountUsed()
+
+      shouldExplore = isAccountUsed && config.shouldExportPubKeyBulk && accounts.length < 100
```
refi93, line 27: the number 100 should be extracted to a constant which may be used also in the UI e.g. to ensure that we don't show a dummy tile of an account to be explored, if the index happens to be beyond this `MAX_ALLOWED_ACCOUNT_INDEX`

PeterBenc:
Changes: 

- rebase the pool registration functionality
- refactor account discovery to recursive, 
- separate account selecting from exploring
- create account manager and export only necessary functions
- refactor account dashboard and account tile
- add error handling for the account discovery
- remove rejected promises of key exporting (if a user rejected the export he would be never able to export it without re-logging)

I ll add further comment tomorrow.

app/frontend/actions.ts
```diff
@@ -11,32 +11,33 @@ import {
...
 import ShelleyCryptoProviderFactory from './wallet/shelley/shelley-crypto-provider-factory'
-import {ShelleyWallet} from './wallet/shelley-wallet'
+import {Wallet} from './wallet/wallet'
+import {parseUnsignedTx} from './helpers/cliParser/parser'
+import {TxPlan, unsignedPoolTxToTxPlan} from './wallet/shelley/shelley-transaction-planner'
 import getDonationAddress from './helpers/getDonationAddress'
 import {localStorageVars} from './localStorage'
 
-let wallet: ReturnType<typeof CardanoWallet | typeof ShelleyWallet>
+let wallet: ReturnType<typeof Wallet>
```
refi93, line 40: I liked ShelleyWallet more - it's easier to find in the code and pollutes the namespace to a lesser extent than a generic "Wallet" class which also leaves to much space for doubt about what kind of wallet it is

app/frontend/actions.ts
```diff
@@ -121,52 +120,43 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-            isShelleyCompatible,
-          })
-          break
+      const cryptoProvider = await ShelleyCryptoProviderFactory.getCryptoProvider(
+        cryptoProviderType,
+        {
+          walletSecretDef,
+          network: NETWORKS.SHELLEY[ADALITE_CONFIG.ADALITE_NETWORK],
+          config,
+          isWebUSB, // TODO: into config
```
refi93, line 139: worth rebasing to latest develop which refactors this to "forceWebUSB"

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  async function getAccountsInfo(validStakepools) {
+    const accounts = await accountManager.discoverAccounts()
+    const accountsInfo = await Promise.all(
+      accounts.map((account) => account.getAccountInfo(validStakepools))
+    )
+    return {
+      accountsInfo,
```
refi93, line 72: wrapping the accountsInfo array into an object seems redundant to me, I would directly return the array of accounts. If we ever need to return something alongside the accounts array we would likely need to refactor this function and whatever depends on it anyway.

I would also add a type annotation to the function to make clear what is returning.

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -368,6 +368,26 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
+    }/api/account/info/${accountPubkeyHex}`
+    const response = await request(url)
+    return response
+  }
+
+  async function getValidStakepools() {
+    const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/stakePools`
+    const validStakepools = await request(url)
+
+    return {validStakepools}
```
refi93, line 384: I would expect this to return directly the array of valid stakePools, the functions that use it also appear to just pluck that "validStakePools" property out of the returned object so seems like an unnecessary overhead

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,95 @@
...
+    const totalWalletBalance = accountsInfo.reduce(
+      (a, {shelleyBalances}) =>
+        shelleyBalances.stakingBalance + shelleyBalances.nonStakingBalance + a,
+      0
+    )
+    const totalRewardsBalance = accountsInfo.reduce(
+      (a, {shelleyBalances}) => shelleyBalances.rewardsAccountBalance + a,
+      0
+    )
+    const shouldShowSaturatedBanner = accountsInfo.some(
```
refi93, line 56: I find it a bit odd to have a flag that represents an apparently higher-level concern (telling directly what should be shown in the UI) in this low-level wallet class. Given that it's been like this before this PR it's probably sensible to keep it, but probably deserves to be extracted together with the logic to determine what pool recommendation to show into actions

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -122,17 +129,53 @@ const CardanoTrezorCryptoProvider = ({network, config}) => {
...
+          stakingKeyHash: undefined,
+        }),
+      })),
+      relays: cert.relays.map((relay) => ({
+        type: relay.type,
+        ...(relay.type === 0 && {
+          ipv4Address: relay.params.ipv4,
+          ipv6Address: relay.params.ipv6,
+        }),
+        ...(relay.type < 2 && {port: relay.params.portNumber}),
```
refi93, line 158: comparing enums with `<` seems really hacky and fragile (what if new types of relays emerge which have a totally different structure?). What I would expect instead is having the enums represented by named constants and test whether the given relay type belongs to a specific collection (be it an array or set) of relay types

app/frontend/wallet/wallet.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  function getWalletName() {
+    return cryptoProvider.getWalletName()
+  }
+
+  function submitTx(signedTx): Promise<any> {
+    const params = {
+      walletType: getWalletName(),
+      // TODO: stakeKey
```
refi93, line 20: this comment doesn't seem very helpful to a casual reader/future ourselves - why the stakeKey would be needed here and what's the impact of not having it there right now?

app/public/js/init.js
```diff
@@ -37,7 +37,8 @@ function playsNice(cb) {
       return
     }
     this.done = 1
-    cb(!(event && (event.error || event.type === 'error' || event.lineno)))
+    cb(!(event && (false)))
+    // TODO: REVERT cb(!(event && (event.error || event.type === 'error' || event.lineno)))
```
refi93, line 41: why we needed to comment out this?

app/frontend/actions.ts
```diff
@@ -898,13 +926,13 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     const state = getState()
     setPoolInfo(state)
     const poolHash = state.shelleyDelegation.selectedPool.poolHash
-    const stakingKeyRegistered = state.shelleyAccountInfo.hasStakingKey
+    const stakingKeyRegistered = sourceAccountState(state).shelleyAccountInfo.hasStakingKey
```
refi93, line 929: naming a function with a noun i.e. `sourceAccountState` is quite misleading, something like `getSourceAccountState` would be more intuitive

app/frontend/actions.ts
PeterBenc: I decided to drop this since I am planning to remove all byron related code so we would need to add "Shelley" keyword to everything. But I agree about the namespace, I ll rename it to shelley-wallet and we can think about this in a future refactor.

app/frontend/actions.ts
PeterBenc: I will, sorry.

app/frontend/wallet/wallet.ts
PeterBenc: Previously I was returning also something else from the function so that's why it doesn't return accountsInfo right away. I ll fix it and also add a type notation.

app/frontend/wallet/blockchain-explorer.ts
PeterBenc: Similarly, this previously also returned something other than the valid pools and was removed some time ago, I think this might change again but until then I ll fix it so we return it right away.

app/frontend/wallet/wallet.ts
PeterBenc: You are right, I ll move the function to `actions.ts`.

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
PeterBenc: @xdzurman

app/frontend/wallet/wallet.ts
PeterBenc: @xdzurman

app/public/js/init.js
PeterBenc: @xdzurman

app/frontend/state.ts
```diff
@@ -212,42 +237,78 @@ const initialState: State = {
...
+    shouldShowPoolCertSignModal: false,
+    ttl: 0,
+    signature: null,
+    plan: null,
+  },
 }
+export type SetStateFn = (newState: Partial<State>) => void
+export type GetStateFn = () => State
+
+export const sourceAccountState = (state: State) => state.accountsInfo[state.sourceAccountIndex]
```
refi93, line 311: isn't that more like `sourceAccountInfo`? and probably `getSourceAccountInfo` would be a better fit as it's a function.

PeterBenc: You are right, I ll fix it :+1:

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+    xPubBulk.forEach((xpub: Buffer, i: number) => {
+      const memoKey = JSON.stringify(paths[i])
+      _derivedXpubs[memoKey] = Promise.resolve(xpub)
+    })
+    return _derivedXpubs
+  }
+
+  function cleanXpubCache() {
+    const _derivedXpubs = {}
+    Object.entries(derivedXpubs).map(([key, xpubPromise]: [string, Promise<Buffer>]) => {
```
refi93, line 78: for the sake of consistency `key` should probably be `memoKey`

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+      xpubPromise
+        .then((xpub) => {
+          _derivedXpubs[key] = Promise.resolve(xpub)
+        })
+        .catch((e) => null)
+    })
+    derivedXpubs = _derivedXpubs
+  }
+
+  return {deriveXpub, cleanXpubCache}
```
refi93, line 88: cleanXpubCache seems like a very low-level thing to expose. I'm actually trying to understand why it's needed (i.e. what could happen if it wouldn't exist) - could you please explain? And isn't there a way to keep the cache cleanup contained within CachedDeriveXpubFactory ?

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -1,10 +1,15 @@
...
 import {derivePublic as deriveChildXpub} from 'cardano-crypto.js'
 
-function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
-  const derivedXpubs = {}
+const BYRON_V2_PATH = [HARDENED_THRESHOLD + 44, HARDENED_THRESHOLD + 1815, HARDENED_THRESHOLD]
 
-  async function deriveXpub(absDerivationPath) {
+type BIP32Path = number[]
+
+function CachedDeriveXpubFactory(derivationScheme, shouldExportPubKeyBulk, deriveXpubFn) {
```
refi93, line 9: I would find it more intuitive to have the deriveXpubHardenedFn passed from above (i.e. as done originally), but the function's API would change to take potentially a list of (hardened) paths instead of a single path. Then it would be up to this function to assemble the right bulk of paths to derive and pass it down to that function. If some of the paths derived within that bulk would fail, the whole promise returned by deriveXpubHardenedFn should just fail, it's not worth dealing with partial results IMO

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -51,21 +53,39 @@ const ShelleyLedgerCryptoProvider = async ({network, config, isWebUSB}) => {
...
+  const {deriveXpub, cleanXpubCache} = CachedDeriveXpubFactory(
+    derivationScheme,
+    config.shouldExportPubKeyBulk,
+    async (derivationPaths) => {
+      const response = await exportPublicKeys(derivationPaths)
+      return response.map((res) => Buffer.from(res.publicKeyHex + res.chainCodeHex, 'hex'))
+    }
+  )
+
+  function checkVersion(requiredVersionType: string) {
```
refi93, line 81: `requiredVersionType` is kind of cryptic, my first hunch was whether this was about minor/major/patch and only afterwards I realized this is about the feature. Something like `ensureFeatureSupported(featureName)` would imho make for a more intuitive interface

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
```diff
@@ -84,21 +92,14 @@ class ConfirmTransactionDialogClass extends Component<Props, {}> {
...
           )}
 
-          {txConfirmType === 'send' && (
-            <Fragment>
-              <div className="ada-label">Amount</div>
-              <div className="review-amount">{printAda(summary.amount)}</div>
-            </Fragment>
-          )}
-
           {txConfirmType === 'convert' && (
```
refi93, line 95: seems like after removing the conditional component above the two fragments can be merged together as both are conditioned by `txConfirmType` being "convert"

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+  function createPathBulk(derivationPath: BIP32Path): BIP32Path[] {
+    const paths: BIP32Path[] = []
+    const accountIndex = derivationPath[2] - HARDENED_THRESHOLD
+    const currentAccountPage = Math.floor(accountIndex / MAX_BULK_EXPORT_AMOUNT)
+
+    /*
+    * in case of the account 0 we append also the byron path
+    * since during byron era only the first account was used
+    */
+    if (accountIndex === 0) paths.push(BYRON_V2_PATH)
```
refi93, line 55: in case of derivationPath being the byron account path, wouldn't this result in `BYRON_V2_PATH` being in paths twice?

app/frontend/wallet/account-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+    return Account({
+      config,
+      cryptoProvider,
+      blockchainExplorer,
+      accountIndex: accounts.length,
+    })
+  }
+
+  async function addNewAccount(account) {
+    await account.isAccountUsed() // To ensure user exported pubkey
```
refi93, line 22: why not ensuring that the key is exported right in `discoverNewAccount`? And if that's really a concern, then wither the `Account()` function itself should take care of that or if that would cause issues, then there should be an `init()` function that would do that and not just "abuse" a seemingly random function for that purpose

app/frontend/wallet/account-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+import {MAX_ACCOUNT_COUNT} from './constants'
+
+const AccountManager = ({config, cryptoProvider, blockchainExplorer}) => {
+  const accounts: Array<ReturnType<typeof Account>> = []
+
+  function getAccount(accountIndex: number) {
+    return accounts[accountIndex]
+  }
+
+  function discoverNewAccount() {
```
refi93, line 12: discoverNextAccount seems like a more suitable name. The "new" keyword does not really reflect that the discovery is performed in a sequential order

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+      const nextAccountIndex = currentAccountPage * MAX_BULK_EXPORT_AMOUNT + i + HARDENED_THRESHOLD
+      const nextAccountPath = [...derivationPath.slice(0, -1), nextAccountIndex]
+      paths.push(nextAccountPath)
+    }
+    return paths
+  }
+
+  async function deriveXpubHardenedFn(derivationPath: BIP32Path): Promise<any> {
+    const paths = shouldExportPubKeyBulk ? createPathBulk(derivationPath) : [derivationPath]
+    const xPubBulk = await deriveXpubFn(paths)
```
refi93, line 67: do we have to await here? Can't we just take the xPubBulk promise and do below:
```
      _derivedXpubs[memoKey] = xPubBulk.then((keys) => return keys[i]).reject(() => return null)
```

would we need the cleanXpubCache function then?

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -16,9 +21,13 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+
+      if (deriveHardened) {
+        const pubKeys = await deriveXpubHardenedFn(absDerivationPath)
+        Object.assign(derivedXpubs, pubKeys)
+      } else {
+        derivedXpubs[memoKey] = deriveXpubNonhardenedFn(absDerivationPath)
+      }
     }
 
     /*
```
refi93, line 33: if we are awaiting on the pubKeys bulk object before assigning it to derivedXpubs, then this comment is no longer true and should be removed. To be fair though - concurrent export of public keys should not be happening anywhere and even if it by any chance did, it would result in a duplicate prompt which would be annoying but not fatal

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
     return await derivedXpubs[memoKey]
```
refi93, line 37: if we rewrite this to
```
let result
    try {
      result = await derivedXpubs[memoKey]
    } catch (e) {
      delete derivedXpubs[memoKey]
      throw e
    }
```

wouldn't that remove the need for the `cleanXpubCache` function? If I understand correctly, the only purpose of that function is cleaning the failed promises.

And if we stored the actual xpubs instead of the promises, we should be able to get rid of the need to invalidate the cache for failed promises altogether. I'm not sure why I made this solution "resilient" against concurrent execution in the past but if there aren't any apparent issues with removing this "feature" I would not be afraid to do so - worst case it would result in a repeated prompt on the hw wallet, even in the hypothetic case that deriveXpub is called asynchronously multiple times

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -28,14 +37,55 @@ function CachedDeriveXpubFactory(derivationScheme, deriveXpubHardenedFn) {
...
+
+    for (let i = 0; i < MAX_BULK_EXPORT_AMOUNT; i += 1) {
+      const nextAccountIndex = currentAccountPage * MAX_BULK_EXPORT_AMOUNT + i + HARDENED_THRESHOLD
+      const nextAccountPath = [...derivationPath.slice(0, -1), nextAccountIndex]
+      paths.push(nextAccountPath)
+    }
+    return paths
+  }
+
+  async function deriveXpubHardenedFn(derivationPath: BIP32Path): Promise<any> {
```
refi93, line 65: deriveXpubFn should imho be renamed back to `deriveXpubHardenedFn` and this function could be then named `_deriveXpubHardenedFn` with a comment above explaining that on top of the original `deriveXpubHardenedFn` this is priming the cache of derived keys to minimize the number of prompts on hardware wallets

refi93 (changes requested):
see comments - mostly ideas to refactor/simplify CachedDeriveXpubFactory and

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
PeterBenc: Thanks :+1:, fixed

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
PeterBenc: I think there is a comment that this should be refactored so didn't want to make it an issue of this PR. Anyway, fixed and we ll refactor it in the future. Thnx.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: Good point, you are right, it would. Of course it cant happen right now, but in future somebody could use this in the wrong way and have a hard time debugging it. Interesting problem, if somebody asks for a byron account key before a Shelley account key this would completely break the logic of appending this key to the "first" bulk export. So the important step is to always export Shelley account key first. Currently this is ensured by the  `init()` and the `isAccountUsed()` functions which export only Shelley keys and they happen before any other pubkey export-related function.

I think we can't really enforce that Shelley pub keys will be always exported first (we should comment it somewhere in the code) but at least it makes sense to not create bulks if the path is from byron era.

app/frontend/wallet/account-manager.ts
PeterBenc: You are right, I tried to not "abuse" the function just for this purpose but we need to ensure that public key is exported before adding the account to the list. I didn't want to create an init function just for this purpose and thought that the comment would suffice but you are right, the init function is cleaner. Actually, the function is needed only in `exploreNewAccount` function because `discoverAccounts` function checks it with `isAccountUsed()`. 

I just wanted to ensure that `addNewAccount(account)` and pubKey export happen in one place.

app/frontend/wallet/account-manager.ts
PeterBenc: I agree `next` is slightly better :+1:

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: But we still store them as promises, all of them, so currently it is still true. And afaik, concurrent export of public keys is currently happening and was until now, (for example addresses have the same parent path) and these are awaiting the same promise.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: Why do you think the function that is passed from the cryptoProvider should have the `hardened` keyword in it. It can derive any key not only hardened compared to the `deriveXpubHardenedFn` in `CachedDeriveXpubFactory` which derives only hardened paths. So it makes more sense to to just call the one that is passed from crypto provider `deriveXpubs` and call this one `deriveXpubsHardenedFn`. What do you think? 

(sorry previously I forgot the `s` in pub`s`) so I'm sure it made it more confusing.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: I get it, I thought about this too but I wanted to keep the async behavior. I could have certainly made this easier for me if I dropped the whole logic of storing promises and used the solution you described. Currently the hardened key derivation happens synchronously (which it must because ledger throws error otherwise) and all the other key derivation asynchronously. 

Im not necessarily against dropping the async behavior but I wanted to find a way in which we don't have to. In that case, afaik there is no other way other than cleaning the xpubCache when re-discovering the accounts.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: My bad, the function should be called `deriveXpubsFn`. I like the idea that the logic for creating the bulk wouldn't be inside the `deriveXpubHardenedFn`.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
PeterBenc: Afaik there is no way to clean the xpubCache inside CachedDeriveXpubFactory if we are deriving the keys asynchronously. I agree it seems a bit low level. Only way afaik would be to drop the async behaviour. 

To answer your question, we need to clear the cache in case a user (lets say) declines the pubkey export prompt. If we didn't clean it he would need to reload and try again.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
refi93: technically, the function passed from outside is being used only for hardened derivation even if it supported non-hardened derivation. And the non-hardened derivation is performed with the cardano-crypto.js function for that, so even though the function passed from outside would be able to do it, it won't be used for that purpose - although it's rather a low-level concern, the point of this cached derivation is exactly that - that it minimizes the calls to "hardened" derivations which in turn minimizes calls to hw wallets, if used, so from that point of view it's IMHO fair to make it obvious to the reader and not make them misleadingly believe that whatever they pass would be used for non-hardened derivation.

app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
refi93: yes, but if you do

```
const pubKeys = await deriveXpubHardenedFn(absDerivationPath)
Object.assign(derivedXpubs, pubKeys)
```
That defeats the point of storing the derived keys as promises because if somebody executes asynchronously the function to derive a key, the promise won't be in the derivedXpubs dictionary yet

PeterBenc:
Changes: 

- Fixed all minor refactoring issuus in comments
  - renamed wallet to shelleyWallet
  - moved `getWalletInfo` func to actions
  - added types.ts file for now consisting only of type for accountInfo,
  - refactored getValidStakePools and getAccounts so they return the value right away without wrapping it in object
  - renamed sourceAccountState to getAccountState etc...
  - use a `ensurePubKeyIsExported` instead of abusing `isAccountUsed` function
  - renamed `new` to `next`
  
- Refactored version checking and implemented checks for bulk export and pool registration
   - this could be refactored further, the version checking and also some pool registration stuff but I would make that into another PR
- Implemented check so the discovery if a hw wallet doesn't support bulk export of pubkeys
- Refactored `CachedDerivedXpubFactory`
   - added condition so bulk is not created in case the path is byron
   - byron path is added to the bulk only if its not already present
   - changed the key derivation to synchronous 
- Fixed support for 12 word mnemonics
- Implemented check for hw wallet version when signing pool reg
- some minor UI issues and texts

app/frontend/actions.ts
```diff
@@ -982,7 +1030,132 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+    const targetAddress = await wallet.getAccount(targetAccountIndex).getChangeAddress()
+    updateAddress(state, null, targetAddress)
+  }
+
+  const showDelegationModal = (state: State, sourceAccountIndex: number) => {
+    setState({
+      sourceAccountIndex,
+      // TODO: move this title logic to the actual component
+      delegationTitle: state.shouldNumberAccountsFromOne
+        ? `Delegate Account #${sourceAccountIndex + 1} Stake`
```
refi93, line 1145: deciding whether to print `#${sourceAccountIndex + 1}` or just `sourceAccountIndex` imho deserves a helper function "formatAccountIndex" to centralize the logic of rendering it

refi93, line 1145: deciding whether to print `#${sourceAccountIndex + 1}` or just `sourceAccountIndex` imho deserves a helper function "formatAccountNumber" to centralize the logic of rendering it

PeterBenc: I agree, that's why I left there the comment also, we are computing it in too many places but let as it was for now. If you find it important enough I can definitely fix it.

app/frontend/wallet/account-manager.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  async function addNextAccount(account) {
+    await account.ensureXpubIsExported() // To ensure user exported pubkey
+    const isLastAccountUsed =
+      accounts.length > 0 ? await accounts[accounts.length - 1].isAccountUsed() : true
+    if (
+      account.accountIndex !== accounts.length ||
+      !isLastAccountUsed ||
+      account.accountIndex > MAX_ACCOUNT_COUNT
```
refi93, line 28: given this condition, MAX_ACCOUNT_COUNT is kind of a misleading name, MAX_ACCOUNT_INDEX would be more accurate

refi93: also, if this condition errors, this check should actually be in discoverAccounts, still it's probably not a bad idea to keep it here as a safeguard

refi93: and finally, I would add maxAccountCount parameter to AccountManager which by default would be  set to MAX_ACCOUNT_COUNT, to be able to set it to something lower from outside to simplify unit testing

app/frontend/actions.ts
refi93: ok let's leave it like this for now, sorry I didn't notice the comment previously

app/frontend/helpers/printAda.ts
```diff
@@ -1,3 +1,4 @@
 // eslint-disable-next-line no-unused-vars
 import {Lovelace} from '../state'
-export default (coins: Lovelace): string => (coins * 0.000001).toFixed(6)
+export default (coins: Lovelace, decimals = 6): string =>
```
refi93, line 3: why did you have to change this?

the new function has an off-by-one error, e.g. for `1645466` it prints `1.645465`

refi93: ah got it - for printing the number rounded to a different amount of decimals, still, it has an off by one error

refi93: but stll, why `(coins * 0.000001).toFixed(decimals)` would be an issue?

PeterBenc: you are exactly right, the reason was to not round the numbers, it was Michal's request to not round the numbers and also made sense to me not show for example `5.345` in account tile and `5.344622` in total balance just above, its pretty obvious if you have only one account. I didn't notice the 1 lovelace difference I am not sure if it happens always or just in some cases, anyway thanks a lot, even thou tx signing don't depend on this so submitting txs should be safe,  we cant display something different in UI and device. 

I dont mind reverting it and having it rounded or we can separate helper for this.

app/tests/src/index.js
```diff
@@ -1,28 +1,31 @@
...
-  })
-  describe('Import/Export Wallet as JSON', () => {
-    require('./keypass-json')
-  })
-  describe('Actions', () => {
-    require('./actions/actions')
-  })
-  describe('Shelley testnet', () => {
-    require('./shelley')
+  // describe('CBOR', () => {
```
refi93, line 4: I would keep the cbor tests, they seem to be testing live code (CborIndefiniteLength array serialization)

app/tests/src/index.js
```diff
@@ -1,28 +1,31 @@
...
+  // describe('Address Manager', () => {
+  //   require('./address-manager')
+  // })
+  // describe('Blockchain Explorer', () => {
+  //   require('./blockchain-explorer')
+  // })
+  // describe('Cardano Wallet', () => {
+  //   require('./cardano-wallet')
+  // })
+  // describe('Import/Export Wallet as JSON', () => {
```
refi93, line 19: the keypass file is also being used so I guess the test should still be passing

app/public/js/init.js
xdzurman: Can't find the original PR but I think this was a quick hack fix of trezor crashing the site during the pool owner functionality presentation.

app/tests/src/index.js
```diff
@@ -1,28 +1,31 @@
...
+  //   require('./keypass-json')
+  // })
+  // describe('Actions', () => {
+  //   require('./actions/actions')
+  // })
+  // describe('Shelley testnet', () => {
+  //   require('./shelley')
+  // })
```
refi93, line 27: basically, I would try to keep the tests that make sense and no ned to comment the rest, just delete them and we can always retrieve them from history rather than having dead code

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
xdzurman: This was initially done since the three types of relays shared the attributes and this saved a lot of lines and copypasting at the time, however I agree with refactoring it.

Basically, there's the:
1. "normal/basic" relay, which has type "0", port, ipv4 address, ipv6 address
2. "single host" relay, which has type 1, port, and dns hostname
and
3. "multi host" relay, which has type 2 and dns hostname

app/frontend/helpers/printAda.ts
refi93: I see there are long threads with many quite elaborate implementations tackling this for the general case, but given we are formatting ADA here we already know the maximum precision of the number and we are multiplying the amount in lovelaces by 0.000001 so there's no point in pretending this is a general purpose currency formatting function.

I would just do `toFixed(6).slice(0, -7 + decimals)` and assert on decimals being at most 6 as there is no point in formatting ADA to more than 6 decimals

refi93: ok I would not overthink it as there are apparently floating operation edgecases in pretty much any "smart" implementation I found. Something like this would do:

```
const coinsDivided = coins * 0.000001;

if (decimals === 0) {
  return Math.floor(coinsDivided).toString() 
} else if (decimals > 0 && decimals <= 6) {
  return coinsDivided.toFixed(7).slice(0, -7 + decimals)
} else {
  return coinsDivided.toFixed(decimals)
}
```

PeterBenc: you a right I totally agree, thanks for the solution :+1:

app/tests/src/index.js
PeterBenc: Of course, I am not planning on dropping any test right now, these were commented out just for development purposes, I ll make the removing the unused test into another PR..

PeterBenc:
Putting this on hold since the release waits for the publishing of articles about and accounts and stake pool owner support so we can link them.

refi93 (approved):

PR #790, opened 2021-01-15
Fix account discovery if max account is used, set max account to 20 for Trezor

### LINKED ISSUE #787: Having more than max account number of accounts results in AccountExplorationError
It's a known bug from the Multi-account feature to be released. Basically as soon as somebody has more than 31 used accounts, the account discovery fails with an AccountExplorationError, preventing the wallet from loading at all

refi93:
Motivation: Originally when account discovery hit the limit of accounts (30), it failed with AccountExplorationError. This PR fixes this and stops the exploration before hitting the limit.

Changes:
* modify discoverAccounts to stop exploring when maxAccountIndex is hit
* remove the "Explore" tile should the last discovered account be at the limit
* Finally, there is a bug in Trezor that prevents transactions on accounts over 20' from working properly (https://github.com/vacuumlabs/trezor-firmware/issues/43), so there is a hotfix that overrides that limit for Trezor and sets maxAccountIndex to 20

Testing:
* tested manually with Ledger by setting the maxAccountIndex to a lower value to reproduce the situation when the condition triggers, exploration stops as expected
* tested manually with Trezor, verifying that the maxAccountIndex is overriden

closes #787

### DIFF
--- a/app/frontend/wallet/shelley-wallet.ts
+++ b/app/frontend/wallet/shelley-wallet.ts
@@ -1,12 +1,23 @@
 import BlockchainExplorer from './blockchain-explorer'
 import {AccountManager} from './account-manager'
 import {AccountInfo} from '../types'
-import {CryptoProviderFeatures} from './constants'
+import {CryptoProviderFeatures, MAX_ACCOUNT_INDEX} from './constants'
 
 const ShelleyWallet = ({config, cryptoProvider}) => {
   const blockchainExplorer = BlockchainExplorer(config)
 
-  const accountManager = AccountManager({config, cryptoProvider, blockchainExplorer})
+  let maxAccountIndex = MAX_ACCOUNT_INDEX
+  if (cryptoProvider.getWalletName() === 'Trezor') {
+    // hotfix because of https://github.com/vacuumlabs/trezor-firmware/issues/43
+    maxAccountIndex = 20
+  }
+
+  const accountManager = AccountManager({
+    config,
+    cryptoProvider,
+    blockchainExplorer,
+    maxAccountIndex,
+  })
 
   function isHwWallet() {
     return cryptoProvider.isHwWallet()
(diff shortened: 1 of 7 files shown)


app/frontend/wallet/shelley-wallet.ts
```diff
@@ -1,12 +1,23 @@
...
 
 const ShelleyWallet = ({config, cryptoProvider}) => {
   const blockchainExplorer = BlockchainExplorer(config)
 
-  const accountManager = AccountManager({config, cryptoProvider, blockchainExplorer})
+  let maxAccountIndex = MAX_ACCOUNT_INDEX
+  if (cryptoProvider.getWalletName() === 'Trezor') {
+    // hotfix because of https://github.com/vacuumlabs/trezor-firmware/issues/43
+    maxAccountIndex = 20
+  }
```
PeterBenc, line 13: Just an idea, if maxAccount number is something dependant on hw wallet, (or its firmware version) we could maybe change this to 
```
const maxAccountIndex = Math.min(cryptoProvider.getMaxAccountIndex(), MAX_ACCOUNT_INDEX)
```
and implement bit more complex logic in trezor crypto provider, for example
```
if (version >= <x.x.x>) return 20; else return 100;
```
and implement the function in all the crypto providers, am not sure but Ledger has also a limit for account index..right?
This might be probably overkilling it but as I said, just an idea.

PeterBenc (approved):
LGTM, just one idea to consider.. :slightly_smiling_face:

app/frontend/wallet/shelley-wallet.ts
refi93: I think that would be an overkill - the max account limit provided by hw wallets should be well above what Adalite supports so the limit may as well be global and this is really a temporary hotfix until Trezor gets out. Of course, in the future we may opt to add custom limits but I wouldn't introduce abstractions we don't need as that would rather complicate testing, introducing crypto-provider specific behavior and ultimately result in a less predictable behavior for users

PR #793, opened 2021-01-20
Refactor fix and type

PeterBenc:
First in a series of refactors and typings

Changes:

- created types.ts where we could declare all types that ale used beyond the scope of one component
- started a refactor of accountInfo getting, I ll try to progress function by function and refactor it to a more sensible way, adding always a test and  in the end ideally optimizing some concurrent behavior, also typing everything in the process, 
- fixed some of the namings in the account.ts, in lot of places, I renamed `account` to `stakingAccount`, and others
- created an interface for cryptoProvider and typed most of its functions


app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -18,14 +18,15 @@ import {
...
-type HexString = string & {__typeHexString: any}
-
-const ShelleyJsCryptoProvider = ({
+const ShelleyJsCryptoProvider = async ({
   walletSecretDef: {rootSecret, derivationScheme},
   network,
   config,
-}) => {
+  // eslint-disable-next-line require-await
+}): Promise<CryptoProvider> => {
```
DavidTranDucVL, line 29: Is this actually returning Promise?

PeterBenc: I changed it to async, so yes it does, but the eslint doesn't like that it doesn't have an await expression and it doesn't have to be async. I changed it because the other crypto providers are async, so afaik, either all crypto providers must async or `ShelleyCryptoProviderFactory` needs to be async.

app/frontend/wallet/shelley/shelley-crypto-provider-factory.ts
```diff
@@ -3,9 +3,13 @@ import ShelleyJsCryptoProvider from './shelley-js-crypto-provider'
 import ShelleyTrezorCryptoProvider from './shelley-trezor-crypto-provider'
 import ShelleyLedgerCryptoProvider from './shelley-ledger-crypto-provider'
 import NamedError from '../../helpers/NamedError'
+import {CryptoProvider} from '../../types'
 
 const ShelleyCryptoProviderFactory = (() => {
-  const getCryptoProvider = (cryptoProviderType, options) => {
+  const getCryptoProvider = (
+    cryptoProviderType: CRYPTO_PROVIDER_TYPES,
```
refi93, line 10: `CryptoProviderTypeEnum` is imho a less "eyebrow raising" type name than `CRYPTO_PROVIDER_TYPES` which looks rather like a collection of something

app/frontend/types.ts
```diff
@@ -1,6 +1,78 @@
...
+  deriveXpub: (derivationPath: BIP32Path) => Promise<Buffer>
+  isHwWallet: () => boolean
+  getHdPassphrase: () => Buffer | void
+  _sign: (message: HexString, absDerivationPath: BIP32Path) => void
+  ensureFeatureIsSupported: (feature: CryptoProviderFeatures) => void
+  isFeatureSupported: (feature: CryptoProviderFeatures) => boolean
+  displayAddressForPath: (absDerivationPath: BIP32Path, stakingPath: BIP32Path) => void
+}
+
+export const enum CERTIFICATES_ENUM {
```
refi93, line 22: `CertificateEnum` is imho fine as well as looks better in function signatures

app/frontend/wallet/constants.ts
```diff
@@ -96,10 +97,10 @@ export const NETWORKS = {
 
 export const DELAY_AFTER_TOO_MANY_REQUESTS = 2000
 export const ADALITE_SUPPORT_EMAIL = 'adalite@vacuumlabs.com'
-export const CRYPTO_PROVIDER_TYPES = {
-  LEDGER: 'LEDGER',
-  TREZOR: 'TREZOR',
-  WALLET_SECRET: 'WALLET_SECRET',
+export const enum CRYPTO_PROVIDER_TYPES {
```
refi93, line 100: the `const` can give a false sense of "contantness" of the object, for enums it's generally advisable to freeze the object as well: https://medium.com/@jdgelok/javascript-enums-with-object-freeze-e3c1cca54de8

refi93: ah sorry - that advice applies sense for JS, not for typescript :) But anyway - it appears that `const enum` may not be what we want as those are removed from the compiled JS: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums - is there any strong reason to have the enums as const?

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -129,14 +130,18 @@ const ShelleyJsCryptoProvider = ({
     return ShelleySignedTransactionStructured(txAux, witnesses, meta)
   }
 
-  function isFeatureSupported(feature) {
+  function isFeatureSupported(feature: CryptoProviderFeatures) {
     return true
   }
 
-  function ensureFeatureIsSupported(feature) {
+  function ensureFeatureIsSupported(feature: CryptoProviderFeatures) {
```
refi93, line 137: `CryptoProviderFeatures` looks like a collection of something at first sight and it looks odd for `feature` to be of such type, so I'd consider renaming the type to `CryptoProviderFeatureEnum` for the sake of clarity

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -79,9 +78,9 @@ const CardanoTrezorCryptoProvider = async ({network, config}) => {
     certificatePointer?: CardanoCertificatePointer
   }
 
-  async function displayAddressForPath(absDerivationPath, stakingPath?) {
+  async function displayAddressForPath(absDerivationPath: BIP32Path, stakingPath?: BIP32Path) {
     const addressParameters: CardanoAddressParameters = {
-      addressType: 0,
+      addressType: 0, // TODO: get from cardano-crypto
```
refi93, line 83: looks like a good opportunity to introduce AddressType enum - I'd not rely on cardano-crypto as it does not have proper typescript definitions, so I would not be afraid to declare our own enum here

refi93 (changes requested):
only minor remarks

app/frontend/wallet/constants.ts
ppershing: const enum in typescript is rather a beast. I basically boils down to compile-time replacement. We generally encourage const enums unless you need to export the type ouside of the project (in which case const enum is no-go). Normal enums are rather a mess because for some inexplicable reason they figured out that enum should contain *both* forward and inverse mapping. This will shoot you in the foot if you are doing some funky stuff like `enum { a='b', b='c', c='a'}`

ppershing: I would also use CamelCaps for enum types. 
`const enum CryptoProviderType` sounds very reasonable to me in this case

Same goes for enum keys although you may want it to keep consistent with the rest of the project so maybe ALL_CAPS is ok there.

PeterBenc: Actually, the reason I used const enums in multiple places was that Pershing recommended it to us :D

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
PeterBenc: Sure I ll add the enum, but the comment meant that we should get the addressType out of the address it self using `cardano-crypto` and not hardcode it here hoping it will always be a Base address.

app/frontend/wallet/shelley/shelley-crypto-provider-factory.ts
PeterBenc: You are right we are not really consistent with enum naming, I noticed but I just basically copy pasted or reusused exististing enums, not making the order in them an issue of this PR, but I ll look into it :+1:

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
PeterBenc: I ll refactor the enum names, thanks :+1:

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
refi93: I see, then this should perhaps be a general comment for the function that it should be extended to accept the expected address type as that's not being passed to it currently - in theory, you can figure it out from the presence/absence of the staking/spending path but seems rather fragile

PeterBenc: I ll comment it better :+1:

refi93 (approved):

PR #794, opened 2021-01-20
Delegate by ticker

PeterBenc:


PeterBenc (commented):
Other than the comments, I noticed that while we are getting the pool info, or if there is an error the space for the pool info shrinks, which wasn't happening before. 
image

PeterBenc:
Also a note, some of the functions related to updating the poolID are a bit messy and done in a hurry shortly after the Shelley release, so if you want you can fix them but I don't want to bother you with it since its mainly my fault, I think you can implement the other changes regardless and then I could fix it, or you can, your call.

DavidTranDucVL:
> Other than the comments, I noticed that while we are getting the pool info, or if there is an error the space for the pool info shrinks, which wasn't happening before.
> image

The issue is that now we are showing either ticker or pool ID, both of which have different length and `pool ID` usually takes 2 rows, so question is, if we even want to keep the height and if we do, then which one.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,8 +26,34 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
+      </Fragment>
+    )
+  } else {
+    return (
+      <Fragment>
+        Ticker or <span className="valid">Stake Pool ID</span>
+      </Fragment>
+    )
+  }
+}
```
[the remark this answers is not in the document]
DavidTranDucVL: Note: with current behaviour, when it's not validated ticker or pool id, neither is highlighted.

DavidTranDucVL:
> Other than the comments, I noticed that while we are getting the pool info, or if there is an error the space for the pool info shrinks, which wasn't happening before.
> image

Yes, I changed this when I added behaviour where we show either Ticker/Pool Has ID, since one is usually longer then other. If we are gonna show both, it makes sense to keep the height.

PeterBenc:
Reopening this since it got close probably accidentally by deleting the base branch.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -101,20 +130,28 @@ interface Props {
...
       dropShadow: false,
+      poolHashOrTicker: '',
     }
   }
 
   componentDidMount() {
-    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') this.props.selectAdaliteStakepool()
+    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') {
+      this.props.selectAdaliteStakepool()
+      this.setState({poolHashOrTicker: this.props.poolRecommendation.recommendedPoolHash})
```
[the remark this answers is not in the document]
DavidTranDucVL: I'm not sure about this one. Shouldn't we aim to keep stuff as modular as possible? It already feels like global is so big and probably some fields used there could be refactored to local state.

Moving it to global state has downside of broadening scope where you have to check what and where the field is used for and. (When you are not familiar with code or just get back to it later.) Meanwhile I don't really see upside of moving it there.

app/frontend/actions.ts
```diff
@@ -1013,7 +1016,8 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
      * pool validation must happen before debouncing
      * but pool info shown after
      */
-    const poolHash = hash || e.target.value
+    const poolByTicker = e && e.target.value && state.validStakepoolsByTicker[e.target.value]
+    const poolHash = hash || (poolByTicker && poolByTicker.poolHash) || e.target.value
     const validationError = poolIdValidator(poolHash, state.validStakepools)
```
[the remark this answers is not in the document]
DavidTranDucVL: I refactored this function in a slightly different way. It doesn't make much sense to forward the responsibility of handling the input event to actions. Now it's only taking in a `poolHash`.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,8 +26,34 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
+  }
+}
+
+const TickerOrHashField = ({inputValue, ticker, poolHash}) => {
+  if (inputValue.length <= MAX_TICKER_LENGTH) {
+    return <div style="word-break: break-all;">{`Stake Pool ID: ${poolHash || ''}`}</div>
+  } else {
+    return <div>{`Ticker: ${ticker || ''}`}</div>
+  }
+}
```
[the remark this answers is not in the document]
DavidTranDucVL: Fixed

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -101,20 +130,28 @@ interface Props {
...
-    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') this.props.selectAdaliteStakepool()
+    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') {
+      this.props.selectAdaliteStakepool()
+      this.setState({poolHashOrTicker: this.props.poolRecommendation.recommendedPoolHash})
+    }
+  }
+
+  handleOnInput = (event) => {
+    this.setState({poolHashOrTicker: event.target.value})
+    this.props.updateStakePoolIdentifier(event)
```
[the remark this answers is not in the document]
DavidTranDucVL: Fixed

app/frontend/components/pages/delegations/delegatePage.tsx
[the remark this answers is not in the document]
DavidTranDucVL: Fixed

DavidTranDucVL:
@PeterBenc please re-review

app/frontend/actions.ts
```diff
@@ -553,10 +553,8 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
   const resetDelegation = () => {
     setState({
       shelleyDelegation: {
-        delegationFee: 0,
-        selectedPool: {
-          poolHash: '',
-        },
+        delegationFee: 0 as Lovelace,
+        selectedPool: null,
```
PeterBenc, line 557: Maybe we could reset the delegation to something more sensible. For example if we reset it to 
```
{
    poolId: '',
    ticker: '',
    ...
}
```
Not sure this wont cause some issues but at least we would not have to check whether this are defined or selectedPool is null etc. but we would instead always display these.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,8 +35,26 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
-const StakePoolInfo = ({pool, gettingPoolInfo}) => {
-  const {fixedCost, homepage, margin, ticker, name, validationError} = pool
+type StakePoolLabelProps = {
+  isTicker: boolean
+  isPoolHash: boolean
+}
+
+const StakePoolLabel = ({isTicker, isPoolHash}: StakePoolLabelProps): h.JSX.Element => (
+  <Fragment>
+    <span className={isTicker ? 'valid' : ''}>Ticker</span> or{' '}
```
PeterBenc, line 45: Is `valid` a good name? I think something like `highlight' would be bit better. Css doesnt need to know about validity.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,8 +35,26 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
+)
+
+type StakePoolInfoProps = {
+  pool?: Stakepool
+  gettingPoolInfo: boolean
+}
+
+const StakePoolInfo = ({pool, gettingPoolInfo}: StakePoolInfoProps): h.JSX.Element => {
+  const {fixedCost, homepage, margin, ticker, name, poolHash, validationError} =
+    pool === null ? ({} as Stakepool) : pool
```
PeterBenc, line 57: If we reset the `selectedPool` to dummy empty string we dont have to check here if its null.
Notice in all these places we do something like 
```
${ticker || ''}
```
and we also would't have to do that.

app/frontend/actions.ts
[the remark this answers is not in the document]
PeterBenc: I totally agree, lets move the logic from actions to the actual component, in that case, tho, we should also do the validation in the component. As mentioned before, the search by ticker might not work and we need a specific error for that. So we can either move the validation to the component and pass the resulting error to action along with the poolhash or do it in actions as we used to.

Its crucial to handle this logic in one place, so either we should get the pool info in the component and pass everything to actions which would just set it or handle everything in actions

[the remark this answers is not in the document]
PeterBenc: I totally agree, lets move the logic from actions to the actual component, in that case, tho, we should also do the validation in the component. As mentioned before, the search by ticker might not work and we need a specific error for that. So we can either move the validation to the component and pass the resulting error to action along with the poolhash or do it in actions as we used to. 

Its crucial to handle this logic in one place, so either we should get the pool info in the component and pass everything to actions which would just set it or handle everything in actions

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,45 +112,65 @@ const BigDelegatorOffer = () => (
...
-class Delegate extends Component<Props, {dropShadow: boolean}> {
-  constructor(props) {
+interface DelegateState {
+  dropShadow: boolean
+  fieldValue: string
+  isTicker: boolean
+  isPoolHash: boolean
+}
+
+class Delegate extends Component<Props, DelegateState> {
```
PeterBenc, line 141: This might be a good chance to refactor this to functional component.

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -376,9 +380,34 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     return response
   }
 
-  function getValidStakepools(): Promise<any> {
-    const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/stakePools`
-    return request(url)
+  async function getValidStakepools(): Promise<Stakepools> {
```
PeterBenc, line 383: Please rebase, these have been already typed so you might need to fix this a bit.

app/frontend/actions.ts
DavidTranDucVL: I don't like this approach that much. Idea is that when you see object that is `null`, in this case `selectedPool` you know that it doesn't exists - you don't have anything selected. Having a non-null field called `selectedPool` is giving impression, that you in have a pool selected.

Also, maybe not that related to this, but I also quite dislike that we have `validationError` inside of this object.

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: See comment above. Although I'm not really happy with this line of code.

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Refactored

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Refactored

app/frontend/wallet/blockchain-explorer.ts
DavidTranDucVL: I rebased, not sure if I resolved everything correctly. I made these resolutions, for pool info:
- `Stakepool` - is used also outside of block chain explorer and has slightly different fields, so I kept it
- `StakePoolInfo` - is used in block chain explorer and not even exported outside of file, so I kept it
For mappings:
- `Stakepools` - since we changed the way mapping works, as in having an underlying object with two accessors `getPoolInfoByTicker` and `getPoolInfoByPoolHash`, I kept this type and removed the other type: `ValidStakePoolsMapping`, because this refactor cause to not be used anywhere anymore

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,148 +113,150 @@ const BigDelegatorOffer = () => (
...
+  pool,
+  isBigDelegator,
+  validStakepools,
+  withAccordion,
+  title,
+  updateStakePoolIdentifier,
+  confirmTransaction,
+  selectAdaliteStakepool,
+}: Props): h.JSX.Element => {
+  if (!stakePool && ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') selectAdaliteStakepool()
```
PeterBenc, line 154: We should probably use `useEffect` here other wise the page wont re-render when changing tabs.

DavidTranDucVL: Fixed

app/frontend/actions.ts
PeterBenc: I agree about the null and selected pool. In case we want it to rest it to empty string we should at least rename it to `selectedPoolInfo` or something similar. Anyway lets keep the null here. 
regarding the error inside the pools, previously there used to be multiple pools you could delegate to so this is what is left of it. Feel free to refactor this but probably in another PR since we don't want to prolong this even more.

app/frontend/components/pages/delegations/delegatePage.tsx
PeterBenc: We could create a dummy `Stakepool` object with all the fields empty, and that way we could remove all the `|| '' ` and also the `as Stakepool` since it would be a valid stakepool.

app/frontend/wallet/blockchain-explorer.ts
PeterBenc: Makes sense.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,148 +113,152 @@ const BigDelegatorOffer = () => (
...
 
-  componentDidMount() {
-    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') this.props.selectAdaliteStakepool()
+  const [fieldValue, setFieldValue] = useState(
+    ADALITE_CONFIG.ADALITE_STAKE_POOL_ID ? poolRecommendation.recommendedPoolHash : ''
+  )
+  const [isTicker, setIsTicker] = useState(false)
+  const [isPoolHash, setIsPoolHash] = useState(!!ADALITE_CONFIG.ADALITE_STAKE_POOL_ID)
+
+  const handleOnInput: (event: any) => void = (event: any): void => {
```
PeterBenc, line 164: Probably `const handleOnInput = (event: any): void => ` would work.

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Refactored

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Refactored

app/frontend/actions.ts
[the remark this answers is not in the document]
DavidTranDucVL: I moved validation to component. I also changed behaviour of search. If we have tickers from BE, it works like previously. If we don't have tickers from BE, the label shows only `Stake Pool ID` instead of `Ticker or Stake Pool ID`, same change in principle for showing error.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,148 +133,168 @@ const BigDelegatorOffer = () => (
...
-                  onBlur={() => this.setState({dropShadow: false})}
-                />
-              </div>
-              <StakePoolInfo pool={stakePool} gettingPoolInfo={gettingPoolInfo} />
-              <div />
-            </li>
-          </ul>
-        </div>
+  const delegationHandler = async (): Promise<void> => {
+    await confirmTransaction('delegate')
```
xdzurman, line 206: is this rebased? I think TxType.DELEGATE can already be used

app/frontend/translations.ts
```diff
@@ -23,7 +23,8 @@ const translations = {
   DonationAmountTooLow: () => `Minimum donation is ${ADALITE_MIN_DONATION_VALUE} ADA`,
   DonationInsufficientBalance: () => 'Insufficient balance for the donation.',
 
-  InvalidStakepoolIdentifier: () => 'Stakepool id is invalid.',
+  InvalidStakepoolIdentifier: ({hasTickerMapping}) =>
+    `Enter valid ${hasTickerMapping ? 'ticker or ' : ''}stakepool id.`,
```
xdzurman, line 27: Enter **a** valid..

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -257,7 +256,9 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     ])
 
     const extractUrl = (poolHash) =>
-      validStakepools[poolHash] ? validStakepools[poolHash].url : null
+      validStakepools.getPoolInfoByPoolHash(poolHash)
+        ? validStakepools.getPoolInfoByPoolHash(poolHash).url
+        : null
```
xdzurman, line 261:
```suggestion
      validStakepools.getPoolInfoByPoolHash(poolHash)?.url || null
```

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -367,9 +368,12 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
             ...nextRewardDetail,
             distributionEpoch: nextRewardDetail.forEpoch + epochsToRewardDistribution,
             pool:
-              validStakepools[poolHash] && validStakepools[poolHash].name
-                ? validStakepools[poolHash]
-                : await getPoolInfo(validStakepools[poolHash].url).catch(() => UNKNOWN_POOL_NAME),
+              validStakepools.getPoolInfoByPoolHash(poolHash) &&
+              validStakepools.getPoolInfoByPoolHash(poolHash).name
+                ? validStakepools.getPoolInfoByPoolHash(poolHash)
```
xdzurman, line 373:
```suggestion
              validStakepools.getPoolInfoByPoolHash(poolHash)?.name
                ? validStakepools.getPoolInfoByPoolHash(poolHash)
```

xdzurman (commented):
Good job, I really like what this PR brings. I suggested some minor quality of life changes

app/frontend/actions.ts
```diff
@@ -1014,27 +1019,31 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
     /**
      * this has to be redone,
      * pool validation must happen before debouncing
      * but pool info shown after
      */
-    const poolHash = hash || e.target.value
-    const validationError = poolIdValidator(poolHash, state.validStakepools)
+    const newPool = state.validStakepools.getPoolInfoByPoolHash(poolHash)
+    const oldPool = state.shelleyDelegation.selectedPool
+    if (newPool && oldPool && newPool.poolHash === oldPool.poolHash) return
```
refi93, line 1040: We can use optional chaining here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

refi93: to be clear it's not 100% equivalent - but if both newPool and oldPool are null, setting the newPool doesn't really make sense

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -45,7 +96,7 @@ const StakePoolInfo = ({pool, gettingPoolInfo}) => {
             >
               <span className="delegation show-info">{''}</span>
             </a>
-            {`Tax: ${margin * 100 || ''}`}%
+            {`Tax: ${margin !== null ? margin * 100 : ''}`}%
```
refi93, line 99: is margin guaranteed to not be undefined? If not, a safer comparison would be `margin != null` and also it should fallback to `n/a` or something like that, not an empty string

app/frontend/types.ts
```diff
@@ -121,3 +123,21 @@ export const enum TxType {
...
+  fixedCost: string
+  url: string
+  name: string
+  ticker: string
+  homepage: string
+  poolHash: string
+  validationError?: any
+}
+
+export type Stakepools = {
```
refi93, line 139: the name is dangerously close to `Stakepool` and may seem to be some simple collection of them rather than an object with such non-trivial interface. IMHO a name `StakepoolsView` (borrowing the term "view" from databases) would be more descriptive and less missleading

refi93: https://www.tutorialsteacher.com/typescript/typescript-interface

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,148 +133,168 @@ const BigDelegatorOffer = () => (
...
+  validStakepools,
+  withAccordion,
+  title,
+  updateStakePoolIdentifier,
+  confirmTransaction,
+  selectAdaliteStakepool,
+}: Props): h.JSX.Element => {
+  useEffect(
+    () => {
+      if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') selectAdaliteStakepool()
```
refi93, line 176: I know this code predates this PR but do we really need to check the config here?

I'm checking the logic of `selectAdaliteStakepool()` (which is removed from this config variable by a few layers of abstraction so one is wondering if it actually does what it's supposed to) and it would imho be simpler to refactor/generalize the function to `tryPreselectingRecommendedPool()` and set it if there is actually any available recommendation, otherwise it would do nothing

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: I assume the reason why it falls back to `''` is so the field is empty, like other fields, when input is empty. But I agree we are mixing two different flows here. I will fix this.
Screenshot_2021-02-15 AdaLite - Cardano Wallet

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -9,11 +9,23 @@ import ConfirmTransactionDialog from '../../pages/sendAda/confirmTransactionDial
...
+  <div className="validation-message send">Calculating fee...</div>
+)
 
-const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
+type DelegationValidationProps = {
+  delegationValidationError: any
+  txSuccessTab: string
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
DavidTranDucVL: Removed this line

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,18 +36,57 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
-        <div>{getTranslation(validationError.code)}</div>
+        <div>{getTranslation(validationError.code, validationError.params)}</div>
       ) : gettingPoolInfo ? (
         <div>Getting pool info..</div>
       ) : (
         <div>
-          <div>{`Name: ${name || ''}`}</div>
-          <div>{`Ticker: ${ticker || ''}`}</div>
+          <div>{`Name: ${name}`}</div>
+          <div style="word-break: break-all;">{`Stake Pool ID: ${poolHash}`}</div>
```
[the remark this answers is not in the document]
DavidTranDucVL: Fixed

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -82,148 +133,168 @@ const BigDelegatorOffer = () => (
...
+  const [fieldValue, setFieldValue] = useState(
+    ADALITE_CONFIG.ADALITE_STAKE_POOL_ID ? poolRecommendation.recommendedPoolHash : ''
+  )
+  const [isTicker, setIsTicker] = useState(false)
+  const [isPoolHash, setIsPoolHash] = useState(!!ADALITE_CONFIG.ADALITE_STAKE_POOL_ID)
 
-  componentDidMount() {
-    if (ADALITE_CONFIG.ADALITE_STAKE_POOL_ID !== '') this.props.selectAdaliteStakepool()
+  const handleOnInput = (event: any): void => {
+    const fieldValue: string = event && event.target && event.target.value
```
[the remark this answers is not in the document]
DavidTranDucVL: Optional chaining doesn't work with our setup of prettier. Don't know how to make it work.

app/frontend/translations.ts
DavidTranDucVL: Fixed.

app/frontend/wallet/blockchain-explorer.ts
DavidTranDucVL: Refactored

app/frontend/wallet/blockchain-explorer.ts
DavidTranDucVL: Refactored

app/frontend/actions.ts
DavidTranDucVL: Optional chaining doesn't work with our setup of prettier. Don't know how to make it work. For now I just ignored prettier errors.

app/frontend/types.ts
DavidTranDucVL: Refactored

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Good point, fixed.

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -427,9 +430,36 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     return response
   }
 
-  function getValidStakepools(): Promise<ValidStakePoolsMapping> {
-    const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/v2/stakePools`
-    return request(url)
+  async function getValidStakepools(): Promise<StakepoolDataProvider> {
+    const createValidStakePools = async (): Promise<StakepoolDataProvider> => {
```
refi93, line 434: sorry to bother further with it but if we renamed the type to StakepoolDataProvider (which makes sense given its use) the old naming of the related functions/variables - validStakePools seems to be still misleadingly suggesting at first sight it's a collection so I would rename them to `poolDataProvider` or something like that

app/frontend/actions.ts
xdzurman: @refi93 I tried to get this working as well - I updated all dependencies, re-installed node modules, etc., without success. I updated the node too. Do you know how to get this to work?

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: I chcked `confirmTransaction` function and it's logic is checking against strings, not `TxType` type. I don't want to refactor that in this PR, it's already big enough.

app/frontend/wallet/blockchain-explorer.ts
DavidTranDucVL: Yes, sorry, I got distracted with some other stuff yesterday and didn't refactor everything that I wanted. This change should be reflected everywhere now.

app/frontend/components/pages/delegations/delegatePage.tsx
DavidTranDucVL: Fixed

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -24,18 +36,57 @@ const DelegationValidation = ({delegationValidationError, txSuccessTab}) =>
...
+    <span className={isPoolHash ? 'highlight' : ''}>Stake Pool ID</span>
+  </Fragment>
+)
+
+type StakePoolInfoProps = {
+  pool?: Stakepool
+  gettingPoolInfo: boolean
+}
+
+const StakePoolInfo = ({pool, gettingPoolInfo}: StakePoolInfoProps): h.JSX.Element => {
```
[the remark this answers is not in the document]
DavidTranDucVL: Refactored

xdzurman (approved):
LGTM, nice. Squash into fewer commits if you can

DavidTranDucVL:
There is still the problem with prettier that I'm not sure how to solve.

app/frontend/actions.ts
refi93: we can look into it, I see `yarn fix` is failing with it, this issue might be related: https://github.com/prettier/prettier/issues/6609

PR #796, opened 2021-01-22
Refactor dashboard page

PeterBenc:
Changes: 

- refactored dashboard page to a functional component
- added types and enums for main tabs, and subtabs,
- created a `constants.ts` file where we can store general purpose enums and constants, previously there was nothing similar and we just used to constants file inside the wallet directory, which might not suit some of the constants.
- updated preact lib so we can use preact dev tools to debug slow rendering

### DIFF
--- a/app/frontend/components/pages/dashboard/dashboardPage.tsx
+++ b/app/frontend/components/pages/dashboard/dashboardPage.tsx
@@ -1,4 +1,4 @@
-import {h, Component, Fragment} from 'preact'
+import {h, Fragment} from 'preact'
 import {connect} from '../../../helpers/connect'
 import actions from '../../../actions'
 import Balance from '../../common/balance'
@@ -10,7 +10,6 @@ import DelegatePage from '../delegations/delegatePage'
 import CurrentDelegationPage from '../delegations/currentDelegationPage'
 import StakingHistoryPage from '../delegations/stakingHistoryPage'
 import ShelleyBalances from '../delegations/shelleyBalances'
-import {ADALITE_CONFIG} from '.././../../config'
 import {MainTab, SubTab} from './tabs'
 import InfoModal from '../../common/infoModal'
 import NotShelleyCompatibleDialog from '../login/nonShelleyCompatibleDialog'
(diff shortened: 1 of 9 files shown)


app/frontend/walletApp.js
```diff
@@ -1,3 +1,4 @@
+// import 'preact/debug'
```
xdzurman, line 1: is this left-over?

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -22,201 +22,210 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
+      {ADALITE_CONFIG.ADALITE_CARDANO_VERSION === 'shelley' && (
+        <ul className="tabinator">
+          {Object.values(MainTabs).map((name, i) => (
+            <MainTab
               key={i}
               name={name}
-              selectedTab={selectedSubTab}
-              selectTab={this.selectSubTab}
+              isSelected={name === selectedMainTab}
+              selectTab={() => selectMainTab(name)}
```
xdzurman, line 185: shouldn't this be just `selectMainTab(name)` instead of `() => selectMainTab(name)` due to `(name: MainTabs) => void` type?

app/frontend/components/pages/dashboard/tabs.tsx
```diff
@@ -1,21 +1,32 @@
...
-    className={`dashboard-tab ${name === selectedTab ? 'selected' : ''}`}
-    onClick={() => selectTab(name)}
-  >
+type SubTabProps = {
+  name: SubTabs
+  selectTab: (name: SubTabs) => void
+  isSelected: boolean
+}
+
+export const SubTab = ({name, isSelected, selectTab}: SubTabProps) => (
```
xdzurman, line 28: Switching the order of isSelected and selectTab arguments would be more consistent with MainTab function arguments

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -22,201 +22,210 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
-  selectedMainTab: any
-  selectMainTab?: (value: string) => void
-}
+import {useState} from 'preact/hooks'
+import {SubTabs, MainTabs} from '../../../constants'
 
 const StakingPage = () => {
+  const subTabs = [SubTabs.DELEGATE_ADA, SubTabs.CURRENT_DELEGATION, SubTabs.STAKING_HISTORY]
+  const defaultSubTab = SubTabs.DELEGATE_ADA
+  const mainSubTab = SubTabs.SHELLEY_BALANCES
```
xdzurman, line 31: subTabs, defaultSubTab and mainSubTab and defined in all of these pages separately, although these variables are shared among them. I would therefore extract them to props and pass their values at their instantiation, so the sub and defaults are in one place and it would be more readable

xdzurman (changes requested):

app/frontend/walletApp.js
PeterBenc: Nope, this is a purposefully left comment, so If a developer needs to use preact dev tools, he just uncomments this.

app/frontend/components/pages/dashboard/dashboardPage.tsx
PeterBenc: You are partly right, In a way, what I need is to pass there a function, either `() => selectMainTab(name)` or `selectMainTab` but the second one the better option since it agrees with the type. Thnx.

app/frontend/components/pages/dashboard/tabs.tsx
PeterBenc: Fair point :+1:

app/frontend/components/pages/dashboard/dashboardPage.tsx
PeterBenc: If I understand you correctly, you would prefer the `DashboardMobile` being called only once, and instead of passing the values from its parent component, we would pass them like `subtabs={MainTabs[selectedMainTab].subtabs}` `defaultSubTab={MainTabs[selectedMainTab].defaultSubTab}` etc. If this is what you mean, I previously planned to do it this way, but I found it a lot less readable to create this massive object in which for every tab I had to define a list of subtabs, defaultSubTab, and mainTab even thou as you pointed out, the binding happens multiple times in that case. 

It's just a matter of creating a big object with subtabs, defaultSubTab, and mainTab for every MainTab or defining these in the respective components, in which the second choice seemed a lot more readable to me, that's it.

app/frontend/walletApp.js
xdzurman: In that case I would add an additional comment after this

app/frontend/components/pages/dashboard/dashboardPage.tsx
xdzurman: Sure but that way it's all in one place and it would be better typed, plus the exact same code
```
<DashboardMobileContent
          subTabs={subTabs}
          defaultSubTab={defaultSubTab}
          mainSubTab={mainSubTab}
        />
```
appear 3 times in the code, which doesn't seem like a good refactoring direction to me

app/frontend/walletApp.js
PeterBenc: Fair point :+1:

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -22,201 +22,210 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
+  const [selectedSubTab, selectSubTab] = useState(defaultSubTab)
+  return (
+    <div className="dashboard-content">
+      {mainSubTab && SubPages[mainSubTab]}
+      <ul className="dashboard-tabs">
+        {subTabs.map((name, i) => (
+          <SubTab
+            key={i}
+            name={name}
+            isSelected={name === selectedSubTab}
```
xdzurman, line 217: Also I'd rename selectedSubTab and selectedMainTab everywhere to active[Main/Sub]Tab, it's super distracting right now

xdzurman (changes requested):

app/frontend/components/pages/dashboard/dashboardPage.tsx
PeterBenc: The fact that some code appears 3 times is a bad smell for sure, but not necessarily a bad decision. First of all, it doesn't really make sense to me to detach the logic of what is displayed on each main tab to some object, secondly this has other implications in terms of UX. I decided to remove that duplicate code as you pointed out and implement it like this.

```
const DashboardMobileContent = ({activeMainTab}: DashboardMobileProps) => {
  const [activeSubTab, setActiveSubTab] = useState(MainPages[activeMainTab].defaultSubTab)
  const subTabs = MainPages[activeMainTab].subTabs
  const mainSubTab = MainPages[activeMainTab].mainSubTab
  useEffect(
    () => {
      setActiveSubTab(MainPages[activeMainTab].defaultSubTab)
    },
    [activeMainTab]
  )
  ```
 I had to use `useEffect` to rerender the default sub tab if the main tab changes, which is terms of UX pretty terrible, (causes the tab to change right await, and it's content only on the second render which is quite noticeable that instead of the content changing all at once, it changes basically twice.)
 
So in this case, it is more than beneficial to have a mobile view not as one component, but to have a mobile view for every tab, even though it happens to be a bit of duplicate code in this example.
 
 I also decided to add a hook for viewport size so we don't render mobile and desktop view at once and I ll be probably reverting to the previous approach even tho it might make a bigger mess out of it.

File: app/frontend/components/pages/dashboard/dashboardPage.tsx (hunk 5 above)
Comment (PeterBenc) [reply app/frontend/components/pages/dashboard/dashboardPage.tsx:217]: I ll change it to `active` and `setActive`, good point :+1:

File: app/frontend/components/pages/dashboard/dashboardPage.tsx (hunk 4 above)
Comment (PeterBenc) [reply app/frontend/components/pages/dashboard/dashboardPage.tsx:49]: I pushed two versions, if you want to, pull it and check out it behaves.

File: app/frontend/components/pages/dashboard/dashboardPage.tsx [hunk 6]
```diff
@@ -22,202 +21,249 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
+        </div>
+      )}
+    </Fragment>
   )
 }
 
-const SendingPage = ({shouldShowExportOption}) => {
+const SendingPage = connect((state: State) => ({
+  shouldShowExportOption: state.shouldShowExportOption,
+}))(
```
xdzurman, line 78: I'd perhaps not create a HOC out of this and have 'shouldShowExportOption' passed as a prop from higher, what do u think? Also, this component has the "isViewportSmall: boolean" prop typed, others do not

app/frontend/components/pages/dashboard/tabs.tsx
```diff
@@ -1,21 +1,32 @@
...
 }
 
-export const SubTab = ({name, selectedTab, selectTab}) => (
-  <li
-    className={`dashboard-tab ${name === selectedTab ? 'selected' : ''}`}
-    onClick={() => selectTab(name)}
-  >
+type SubTabProps = {
+  name: SubTabs
+  setActive: (name: SubTabs) => void
```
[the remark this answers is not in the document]
PeterBenc: Thanks, fixed :+1:

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -22,202 +21,249 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
-interface Props {
-  selectedMainTab: any
-  selectMainTab?: (value: string) => void
+// we could export this and use it in other places, in that case
+// we should optimalize it with context
+// https://blog.logrocket.com/developing-responsive-layouts-with-react-hooks/
+const useViewport = () => {
+  const [isViewportSmall, setIsViewportSmall] = useState(false)
+
+  const checkScreenSize = () => {
```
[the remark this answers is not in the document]
PeterBenc: Fixed.

app/frontend/components/pages/dashboard/dashboardPage.tsx
PeterBenc: I didn't want to connect the dashboard page to `shouldShowExportOption` just so it passes it elsewhere, but doesn't matter, fixed, and typed.

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -22,209 +21,260 @@ import AccountsDashboard from '../accounts/accountsDashboard'
...
-interface Props {
-  selectedMainTab: any
-  selectMainTab?: (value: string) => void
+// we could export this and use it in other places, in that case
+// we should optimalize it with context
+// https://blog.logrocket.com/developing-responsive-layouts-with-react-hooks/
+const useViewport = () => {
+  const [isViewportSmall, setIsViewportSmall] = useState(false)
+
+  const handleScreeResize = () => {
```
xdzurman, line 33: typo

xdzurman (approved):

PR #798, opened 2021-01-25
Fix tests and add tests for getting account keys

PeterBenc:
Fixes unhandled mocks, add new ones, and adds tests for getting the account keys.

### DIFF
--- a/app/frontend/wallet/account.ts
+++ b/app/frontend/wallet/account.ts
@@ -4,11 +4,10 @@ import {MAX_INT32} from './constants'
 import NamedError from '../helpers/NamedError'
 import {CryptoProvider, Lovelace} from '../types'
 import {
-  getStakingAddressHex,
   getAccountXpub as getAccoutXpubShelley,
   ShelleyStakingAccountProvider,
   ShelleyBaseAddressProvider,
-  getStakingKeyCborHex,
+  getStakingXpub,
 } from './shelley/shelley-address-provider'
 
 import {
@@ -151,6 +150,10 @@ const MyAddresses = ({
     )
   }
 
+  async function getStakingAddress() {
+    return await accountAddrManager._deriveAddress(accountIndex)
+  }
+
   return {
     getAddressToAbsPathMapper,
     fixedPathMapper,
@@ -160,6 +163,7 @@ const MyAddresses = ({
     accountAddrManager,
     legacyExtManager,
     areAddressesUsed,
+    getStakingAddress,
   }
 }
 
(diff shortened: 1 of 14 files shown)


xdzurman (approved):
LGTM, but I didn't quite fancy walletSettings and cryptoProviderSettings being arrays. Giving them proper key names in objects would be more readable than referencing it by [0], [1], etc., and having to look in the original array to figure out what it actually is. The proposed solution would look like this: "walletSettings.shelley15Word" instead of "walletSettings[0]". We can do that in another PR

PeterBenc:
> LGTM, but I didn't quite fancy walletSettings and cryptoProviderSettings being arrays. Giving them proper key names in objects would be more readable than referencing it by [0], [1], etc., and having to look in the original array to figure out what it actually is. The proposed solution would look like this: "walletSettings.shelley15Word" instead of "walletSettings[0]". We can do that in another PR

I agree definitely, I did this initially to be compliant with the previous tests but we should have the setting as objects for sure.

app/frontend/wallet/account.ts
```diff
@@ -500,6 +500,7 @@ const Account = ({
     getPoolRecommendation,
     isAccountUsed,
     ensureXpubIsExported,
+    _getKeys: getKeys,
```
refi93, line 503: I checked the `getKeys()` implementation:
```
async function getKeys() {
    const shelleyAccountXpub = await getAccoutXpubShelley(cryptoProvider, accountIndex)
    const byronAccountXpub = await getAccoutXpubByron(cryptoProvider, accountIndex)
    const stakingKeyCborHex = await getStakingKeyCborHex(cryptoProvider, accountIndex)
    const stakingAddress = await myAddresses.accountAddrManager._deriveAddress(accountIndex)
    const stakingAddressHex = await getStakingAddressHex(cryptoProvider, accountIndex)
    return {
      shelleyAccountXpub,
      byronAccountXpub,
      stakingKeyCborHex,
      stakingAddress,
      stakingAddressHex,
    }
  }
```

 and it seems to me that it should be named differently or better, broken down into e.g. `getAccountKeys()` and `getStakingCredentials()`. And `stakingKeyCborHex` would probably be neater if it was just the hex of the public key (possibly even xpub) and do the cbor-encoding wherever needed with some simple utility function (e.g. xpubToCborHex) or something like that as it is a matter of presentation and is preferable to hold data in the state in a standardized format and not bloat it with different presentations of the same piece of information which can easily get out of sync when implementing changes. Same goes for `stakingAddressHex` as the hex representation of the address could easily be extracted with a helper function bech32ToHex wherever needed, avoiding duplication and bloating in the state.

PeterBenc: You are right, I left this half-refactored and targeted mostly on fixing the names of the variables and I united keys and addresses into one function, but I ll fix it, and regarding the state pollution, the reason I put multiple versions of the same key, like hex and cbor into state was to separate the encoding logic from the `keys.tsx` component, but you are right, we could create the mentioned helper functions and do it that way. Thanks.

refi93 (approved):
LGTM, just fix the eslint error please

PR #799, opened 2021-01-26
Type address provider and manager

PeterBenc:


app/frontend/types.ts
```diff
@@ -56,11 +65,6 @@ export type _PubKeyCbor = {
   cborHex: HexString
 }
 
-export type _Address = {
```
refi93, line 59: seems to me like "DerivedAddress" could be a good name for this type as it would convey the information that we derived the address and therefore it's natural to expect some additional context like the derivation path and possibly encoding as well

app/frontend/types.ts
```diff
@@ -1,5 +1,14 @@
 export type BIP32Path = number[]
 
+export type _Address = string & {__typeAddress: any}
```
refi93, line 3: (probably noob) question: Do I understand correctly that `& {__typeAddress: any}` is a way to "tag" the given type leveraging type intersection so the type-checking system doesn't mix it with arbitrary strings or is there a different rationale behind this pattern? Does this pattern have a name by which I could look it up in docs/tutorials?

app/frontend/wallet/address-manager.ts
```diff
@@ -59,10 +67,11 @@ const AddressManager = ({addressProvider, gapLimit, blockchainExplorer}) => {
     })
   }
 
-  function getAddressToAbsPathMapping() {
+  // this is supposed to return {[key: _Address]: BIP32Path} but ts does support
```
refi93, line 70: I think this is a feature rather than a "bug" given that arbitrary types may not be hashable/comparable in a way that would make the mapping work as expected, so I would not put here the comment. Maybe a type alias "AddressString" could do to make the intent clear instead of a comment

refi93: an alternative would be returning a `Map` which I'd guess typescript would allow to have arbitrary entries as that's what's in the specs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map - but I can see it having unintended side effects if we refactor the _Address type to actually be an object so I'd rather not go down that path

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -32,17 +32,21 @@ export const base58AddressToHex = (address: string): HexString => {
...
+export const stakingAddressHexFromXpub = (stakeXpub: Buffer, networkId: number): HexString => {
   const addrBuffer: Buffer = packRewardAddress(xpub2blake2b224Hash(stakeXpub), networkId)
   return Buffer.from(addrBuffer).toString('hex')
 }
 
-export const baseAddressFromXpub = (spendXpub: Buffer, stakeXpub: Buffer, networkId): string => {
+export const baseAddressFromXpub = (
+  spendXpub: Buffer,
+  stakeXpub: Buffer,
+  networkId: number
```
refi93, line 48: it seems sensible to declare a dedicated type for networkId as well which could serve as a guard against passing there some garbage/mixing parameters (eg. protocolMagic with networkId)

refi93 (commented):
LGTM, just a few questions/idea for adding a separate type for networkId

refi93 (approved):

app/frontend/types.ts
ppershing: Yes. At nozdormu we even have a helper type `export type Nominal<Brand, T> = T & {'__ brand': Brand}`

app/frontend/wallet/shelley/helpers/addresses.ts
PeterBenc: Sure, to do that I have to type networks, I ve already started working on it but if you don't mind I ll make it into another PR.

PR #800, opened 2021-01-26
Refactor login page and its sidebar

xdzurman:
- Converts loginPage and and loginPageSidebar from classes to components with hooks
- Removes now unnecessary ENABLE_HW_WALLETS constant
- Introduces types
- Renames misleading variables
- Uses useViewport to render components depending on the screen size instead of hiding desktop/mobile components
- Solves mobile bug of HW wallets login not having padding
- Prepares ground for private key login

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -104,7 +104,8 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
   const setAuthMethod = (state: State, option: AuthMethodEnum): void => {
     setState({
       authMethod: option,
-      shouldShowExportOption: option === 'mnemonic' || option === 'file',
+      shouldShowExportOption:
+        option === AuthMethodEnum.Mnemonic || option === AuthMethodEnum.KeyFile,
     })
   }
 
@@ -273,7 +274,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
       },
       walletLoadingError: undefined,
       shouldShowWalletLoadingErrorModal: false,
-      authMethod: 'mnemonic',
+      authMethod: AuthMethodEnum.Mnemonic,
       shouldShowExportOption: true,
     })
   }
@@ -326,7 +327,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
         formIsValid: false,
       },
       shouldShowGenerateMnemonicDialog: true,
-      authMethod: 'mnemonic',
+      authMethod: AuthMethodEnum.Mnemonic,
       shouldShowMnemonicInfoAlert: true,
     })
   }
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -37,7 +37,7 @@ import {parseUnsignedTx} from './helpers/cliParser/parser'
 import {TxPlan, unsignedPoolTxToTxPlan} from './wallet/shelley/shelley-transaction-planner'
 import getDonationAddress from './helpers/getDonationAddress'
 import {localStorageVars} from './localStorage'
-import {AccountInfo, Ada, Lovelace, CryptoProviderFeature, _Address, TxType} from './types'
+import {AccountInfo, Ada, Lovelace, CryptoProviderFeature, _Address, TxType, AuthMethodType} from './types'
 import {MainTabs} from './constants'
 
 let wallet: ReturnType<typeof ShelleyWallet>
@@ -101,11 +101,11 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     }
   }
 
-  const setAuthMethod = (state: State, option: AuthMethodEnum): void => {
+  const setAuthMethod = (state: State, option: AuthMethodType): void => {
     setState({
       authMethod: option,
       shouldShowExportOption:
-        option === AuthMethodEnum.Mnemonic || option === AuthMethodEnum.KeyFile,
+        option === AuthMethodType.MNEMONIC || option === AuthMethodType.KEY_FILE,
     })
   }
 
@@ -274,7 +274,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
       },
       walletLoadingError: undefined,
       shouldShowWalletLoadingErrorModal: false,
-      authMethod: AuthMethodEnum.Mnemonic,
+      authMethod: AuthMethodType.MNEMONIC,
       shouldShowExportOption: true,
     })
   }
@@ -327,7 +327,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
         formIsValid: false,
       },
       shouldShowGenerateMnemonicDialog: true,
-      authMethod: AuthMethodEnum.Mnemonic,
+      authMethod: AuthMethodType.MNEMONIC,
       shouldShowMnemonicInfoAlert: true,
     })
   }
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -101,7 +101,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     }
   }
 
-  const setAuthMethod = (state, option) => {
+  const setAuthMethod = (state: State, option: AuthMethodEnum): void => {
     setState({
       authMethod: option,
       shouldShowExportOption: option === 'mnemonic' || option === 'file',
--- a/app/frontend/components/pages/dashboard/dashboardPage.tsx
+++ b/app/frontend/components/pages/dashboard/dashboardPage.tsx
@@ -21,35 +21,18 @@ import AccountsDashboard from '../accounts/accountsDashboard'
 import {State} from '../../../state'
 import PoolOwner from '../advanced/poolOwner'
 import ErrorModals from './errorModals'
-import {useEffect, useState} from 'preact/hooks'
+import {useState} from 'preact/hooks'
 import {SubTabs, MainTabs} from '../../../constants'
+import {useViewport, isSmallerThanDesktop} from '../../common/viewPort'
+import {ScreenType} from '../../../types'
 
-// we could export this and use it in other places, in that case
-// we should optimalize it with context
-// https://blog.logrocket.com/developing-responsive-layouts-with-react-hooks/
-const useViewport = () => {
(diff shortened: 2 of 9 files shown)


app/frontend/actions.ts
```diff
@@ -100,7 +108,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     }
   }
 
-  const setAuthMethod = (state, option) => {
+  const setAuthMethod = (state, option: AuthMethodEnum): void => {
```
PeterBenc, line 111: Since we are already modifying this, could we type the state as well?

app/frontend/components/pages/login/loginPage.tsx
```diff
@@ -16,190 +17,198 @@ import Tag from '../../common/tag'
...
+    authMethod,
+    shouldShowDemoWalletWarningDialog,
+    logoutNotificationOpen,
+    walletLoadingError,
+    shouldShowGenerateMnemonicDialog,
+    shouldShowWalletLoadingErrorModal,
+    shouldShowStakingBanner,
+    autoLogin,
+    errorBannerContent,
+  } = useSelector((state) => ({
```
PeterBenc, line 211: We could type the state here as well.

app/frontend/components/pages/login/loginPageSidebar.tsx
```diff
@@ -164,23 +164,16 @@ const FileContent = () => (
   </div>
 )
 
-interface Props {
-  authMethod: '' | 'mnemonic' | 'hw-walet' | 'file'
+const LoginPageSidebar = () => {
+  const {authMethod} = useSelector((state) => ({authMethod: state.authMethod}))
```
PeterBenc, line 168: Here as well.

app/frontend/components/pages/login/loginPage.tsx
```diff
@@ -16,190 +17,198 @@ import Tag from '../../common/tag'
...
+    window.addEventListener('resize', handleScreenResize)
+
+    return () => window.removeEventListener('resize', handleScreenResize)
+  }, [])
+
+  return isViewportSmall
+}
+
+// TODO: extract to app/frontend/constants.ts after rebase
 const AUTH_METHOD_NAMES = {
```
PeterBenc, line 40: Im not sure if you meant to do this after the rebase as well, but we could make this into an enum and use it in this file a bit more. Most of the components here use the literar strings like `'mnemonic'` and `'hw-wallet'`...
Also it might be nice to check where else we are using the literal values, like in `actions.ts` in functions like `setAuthMethod`,  `loadDemoWallet` and `openGenerateMnemonicDialog`.

PeterBenc (commented):
Thanks, LGTM. It would be nice to be able to reuse the enum of auth methods everywhere but we could also make this into a separate issue.

app/frontend/components/pages/login/loginPage.tsx
xdzurman: Yeah it's used with string literals in many places, I stated it as possible future changes, but I can do it now

app/frontend/actions.ts
xdzurman: Sure, we can type it bit by bit with every PR

app/frontend/components/pages/login/loginPage.tsx
PeterBenc: Sorry, I wasnt sure if you just wanted to extract it to `constants.ts` or also do this, but still, If it doesn't bother you that much we should do it now, I can hardly imagine somebody getting back to it later just to fix this.

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -139,12 +122,12 @@ const AdvancedPage = ({isViewportSmall}: {isViewportSmall: boolean}) => {
...
 }
 
-const AccountsPage = ({isViewportSmall}) => {
+const AccountsPage = ({screenSize}: {screenSize: ScreenSize}) => {
   const subTabs = [SubTabs.ACCOUNTS]
   const defaultSubTab = SubTabs.ACCOUNTS
   return (
     <Fragment>
-      {isViewportSmall ? (
+      {screenSize < ScreenSize.DESKTOP ? (
```
refi93, line 130: comparing enums with `<` is kind of fragile. I would perhaps preserve the isViewportSmall function but it would take the screenSize as an argument and I would make it exportable from useViewport.js, perhaps renaming that file to just "viewPort"

app/frontend/types.ts
```diff
@@ -55,7 +55,19 @@ export type _XPubKey = {
   xpubHex: HexString
 }
 
-export type AuthMethod = '' | 'hw-wallet' | 'mnemonic' // TODO
+export enum AuthMethodEnum {
+  INITIAL,
```
refi93, line 59: this looks to me like an abuse of an enum that is meant to represent actual AuthMethods. If I understand correctly, this represents the fact that no AuthMethod has been selected yet which is kind of counterintuitive to encapsulate within the AuthMethod itself and can expose logic handling different actual AuthMethods to deal with this weird case.

I would find it more natural to not have this "initial" type at all and instead have the logic that expects that to have an interface expecting an optional type (?AuthMethodEnum) and logic that is expected to handle valid AuthMethods to expect a non-optional type. And the initial AuthMethod in the state would just be null, idiomatically representing the fact that no auth method has been selected

app/frontend/types.ts
```diff
@@ -55,7 +55,19 @@ export type _XPubKey = {
   xpubHex: HexString
 }
 
-export type AuthMethod = '' | 'hw-wallet' | 'mnemonic' // TODO
+export enum AuthMethodEnum {
```
refi93, line 58: AuthMethodType would imho be a more suitable name - "enum" is a low-level concept not really worth exposing via naming

app/frontend/components/pages/login/loginPageSidebar.tsx
```diff
@@ -164,23 +166,16 @@ const FileContent = () => (
   </div>
 )
 
-interface Props {
-  authMethod: '' | 'mnemonic' | 'hw-walet' | 'file'
+const LoginPageSidebar = () => {
+  const {authMethod} = useSelector((state: State) => ({authMethod: state.authMethod}))
+  return (
+    <aside className="sidebar">
+      {authMethod === AuthMethodEnum.INITIAL && <InitialContent />}
```
refi93, line 173: This seems like a good opportunity to leverage case exhaustiveness checking by typescript: https://stackoverflow.com/questions/59685270/typescript-ensure-all-cases-are-handled-for-union-type

app/frontend/constants.ts
```diff
@@ -18,3 +20,9 @@ export enum SubTabs {
   BALANCE = 'Balance',
   SHELLEY_BALANCES = 'Shelley Balances',
 }
+
+export const AuthMethodNames = {
+  [AuthMethodEnum.MNEMONIC]: 'Mnemonic',
```
refi93, line 25: I would directly put this into the getAuthMethodName function and leverage typescript to ensure no auth method is left out

refi93 (changes requested):
see comments - mostly improvement ideas to take more advantage of TypeScript

app/frontend/components/pages/dashboard/dashboardPage.tsx
xdzurman: The motivation for this kind of comparison was that different pages have a different "splitting point" for their logic. For example, dashboardPage needs to have one layout for (mobile, tablet) together and another layout for the desktops. On the other hand, loginPage needs to have one layout for mobile and the other for (tablet, desktop) together.

I would therefore shoot for two functions to retain these "intervals". The first would be the one you mentioned but renamed
```
  function isSmallerThanDesktop(screenSize: ScreenSize) {
    return screenSize in [ScreenSize.MOBILE, ScreenSize.TABLET]
}
```
and the other one would work for the other splitting point (I believe we won't need more splits)
```
  function isBiggerThanMobile(screenSize: ScreenSize) {
    return screenSize in [ScreenSize.TABLET, ScreenSize.DESKTOP]
}
```

What do you think?

app/frontend/components/pages/login/loginPageSidebar.tsx
xdzurman: This solution is seems a little bit hard to pull off with out current settings. It implicitly assumes strictNullChecks tsconfig option to be set "true". If you go ahead to the playground and turn the option to off, this solution no longer works.

If I turn this on, it introduces a billion errors from all parts of code. The other solution - with strictNullChecks off - is to have an assert at the end of such exhaustive functions https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript . But imho that pollutes the codebase if we were to do that in every such an exhaustive checking function. What do you think?

app/frontend/components/pages/dashboard/dashboardPage.tsx
refi93: I see, makes sense, what you suggest looks good to me

app/frontend/components/pages/login/loginPageSidebar.tsx
xdzurman: I introduced the assert function and exhaustive switches for the time being. I handled the most obvious auth method functionalities, although it could be refactored further for DropdownItems and AuthTabs, but that would require iterating a typescript enum, which is not that pretty.

refi93 (approved):
LGTM, just squash & rebase

app/frontend/components/pages/login/loginPageSidebar.tsx
```diff
@@ -164,23 +167,28 @@ const FileContent = () => (
   </div>
 )
 
-interface Props {
-  authMethod: '' | 'mnemonic' | 'hw-walet' | 'file'
+const SidebarContentByAuthMethod = ({authMethod}: {authMethod: AuthMethodType}) => {
+  switch (authMethod) {
+    case null:
```
refi93, line 172: hmm, shouldn't AuthMethodType be optional? I would expect typescript checks to catch it

xdzurman: Same as with setting AuthMethodType to null in the initial state. Perhaps I'd type the auth method as 
`authMethod: AuthMethodType | null` to be more precise

PR #810, opened 2021-01-31
Separate css into multiple files by media queries

PeterBenc:
Motivation:
Styling anything used to be a huge pain since you always had to make sure you in the right part of the file (in the right media query). This PR separates the styles into multiple files so it's at least easier to navigate and to know what part of the view was edited on which screen.


xdzurman (changes requested):
Different css files have diffent indent

PeterBenc:
> Different css files have diffent tab indent

Thanks, should be fixed.

xdzurman (approved):

PR #811, opened 2021-02-01
Test tx plannig

PeterBenc:
Motivation: 
Test for tx planning were not present at all.

Changes:
- refactor tests and refactors settings to objects
- adds tests for calculating fee
- adds tests for getting tx plan and tx plan settings

Gotcha:

- test for calculating fee is divided into three parts, sendAda, delegation and reward withdrawal, all three containing only one test for now and not having tx plan in its settings. Testing the contents of the plan is the responsibility of the account and these tests are supposed to just test if the fee was calculated, an error was thrown or a behavior caused by the success of getting the tx plan.

### DIFF
--- a/app/tests/src/actions/transaction-actions.js
+++ b/app/tests/src/actions/transaction-actions.js
@@ -1,4 +1,4 @@
-import {setStateFn, setupInitialState} from './actions'
+import {setMockState, setupInitialMockState} from './actions'
 import {ADALITE_CONFIG} from '../../../frontend/config'
 import mockNetwork from '../common/mock'
 import {CryptoProviderType} from '../../../frontend/wallet/constants'
@@ -9,7 +9,7 @@ import {walletSettings} from '../common/wallet-settings'
 let state, action
 
 beforeEach(() => {
-  ;[state, action] = setupInitialState()
+  ;[state, action] = setupInitialMockState()
 })
 
 before(() => {
@@ -30,11 +30,12 @@ before(() => {
   mockNet.mockPoolRecommendation()
 })
 
-const loadTestWallet = async () => {
+const loadTestWallet = async (mockState) => {
   await action.loadWallet(state, {
     cryptoProviderType: CryptoProviderType.WALLET_SECRET,
     walletSecretDef: await mnemonicToWalletSecretDef(walletSettings.Shelley15Word.secret),
   })
+  setMockState(state, mockState)
 }
 
 const sendAdaTxSettings = {
@@ -95,8 +96,7 @@ const withdrawalSettings = {
 describe('Send ADA fee calculation', () => {
   Object.entries(sendAdaTxSettings).forEach(([name, setting]) =>
     it(`should calculate fee for tx with ${name}`, async () => {
-      await loadTestWallet()
-      setStateFn(state, setting.state)
+      await loadTestWallet(setting.state)
       await action.calculateFee()
       assert.deepEqual(state.sendTransactionSummary.fee, setting.sendTransactionSummary.fee)
     })
@@ -106,8 +106,7 @@ describe('Send ADA fee calculation', () => {
 describe('Delegation fee calculation', () => {
   Object.entries(delegationSettings).forEach(([name, setting]) =>
     it(`should calculate fee for tx with ${name}`, async () => {
-      await loadTestWallet()
-      setStateFn(state, setting.state)
+      await loadTestWallet(setting.state)
       await action.calculateDelegationFee()
       assert.deepEqual(state.sendTransactionSummary.fee, setting.sendTransactionSummary.fee)
     })
@@ -117,8 +116,7 @@ describe('Delegation fee calculation', () => {
 describe('Withdrawal fee calculation', () => {
   Object.entries(withdrawalSettings).forEach(([name, setting]) =>
     it(`should calculate fee for tx with ${name}`, async () => {
-      await loadTestWallet()
-      setStateFn(state, setting.state)
+      await loadTestWallet(setting.state)
       await action.withdrawRewards(state)
       assert.deepEqual(state.sendTransactionSummary.fee, setting.sendTransactionSummary.fee)
     })
--- a/app/tests/src/actions/transaction-actions.js
+++ b/app/tests/src/actions/transaction-actions.js
@@ -31,7 +31,7 @@ it('Calculate fee - shelley', async () => {
 
   await action.loadWallet(state, {
     cryptoProviderType: CryptoProviderType.WALLET_SECRET,
-    walletSecretDef: await mnemonicToWalletSecretDef(walletSettings[0].secret),
+    walletSecretDef: await mnemonicToWalletSecretDef(walletSettings.Shelley15Word.secret),
   })
 
   state.sendAddress.fieldValue =
--- a/app/tests/src/actions/transaction-actions.js
+++ b/app/tests/src/actions/transaction-actions.js
@@ -1,4 +1,4 @@
-import {setupInitialState} from './actions'
+import {setStateFn, setupInitialState} from './actions'
 import {ADALITE_CONFIG} from '../../../frontend/config'
 import mockNetwork from '../common/mock'
 import {CryptoProviderType} from '../../../frontend/wallet/constants'
@@ -12,7 +12,7 @@ beforeEach(() => {
   ;[state, action] = setupInitialState()
 })
 
-it('Calculate fee - shelley', async () => {
+before(() => {
   ADALITE_CONFIG.ADALITE_CARDANO_VERSION = 'shelley'
   ADALITE_CONFIG.ADALITE_NETWORK = 'MAINNET'
   const mockNet = mockNetwork(ADALITE_CONFIG)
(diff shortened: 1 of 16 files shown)


app/tests/src/wallet/account-manager.js
```diff
@@ -45,62 +40,40 @@ const initAccountManager = async (settings, i) => {
...
-  })
-  it('should not discover accounts if bulk export is disabled', async () => {
-    const mockNet = mockNetwork(ADALITE_CONFIG)
-    mockNet.mockBulkAddressSummaryEndpoint()
-    const accounts = await accountManagers[1].discoverAccounts()
-    // console.log(accounts, accountManagerSettings[1])
-    assert.equal(accounts.length, accountManagerSettings[1].numberOfDiscoveredAccounts)
-    mockNet.clean()
-  })
-  it('should not discover accounts if wallet is not shelley compatible', async () => {
```
xdzurman, line 77: sorry, why are these two getting removed?

app/tests/src/common/crypto-provider-settings.js
```diff
@@ -37,21 +37,32 @@ const cryptoProviderSettings = [
     type: 'mnemonic',
     isShelleyCompatible: true,
   },
-  {
+]
```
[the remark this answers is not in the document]
PeterBenc: Of course, we will, I just didn't do it now and left the removal of byron tests and byron code to a separate PR.

app/tests/src/wallet/account-manager.js
PeterBenc: They havent really got removed, I am iterating over the settings, and that way I don't have to define a test for each one of them.
image

xdzurman (approved):

app/tests/src/actions/transaction-actions.js
```diff
@@ -1,4 +1,4 @@
-import {setupInitialState} from './actions'
+import {setStateFn, setupInitialState} from './actions'
```
refi93, line 1: I'd rename this to setupInitialMockState and setMockState (without the 'Fn' suffix as it doesn't convey anything important) to make it clear to the reader that we are not reusing functions from the actual frontend which is otherwise easy to confuse given the relative import of actions from the "tests" folder

app/tests/src/wallet/account-manager.js
```diff
@@ -45,62 +40,40 @@ const initAccountManager = async (settings, i) => {
...
-    mockNet.clean()
-  })
+describe('Account discovery', () => {
+  Object.entries(accountManagerSettings).forEach(([name, setting]) =>
+    it(`should discover the right amount of accounts ${name}`, async () => {
+      const accountManager = await accountManagers[name]
+      const mockNet = mockNetwork(ADALITE_CONFIG)
+      mockNet.mockBulkAddressSummaryEndpoint()
+      const accounts = await accountManager.discoverAccounts()
+      assert.equal(accounts.length, setting.numberOfDiscoveredAccounts)
```
refi93, line 59: settings.expectedNumberOfDiscoveredAccounts would be a more suitable to name to make it clear at first sight what's the returned value and what's the expected one

app/public/css/0-1366px.css
```diff
@@ -0,0 +1,28 @@
+@media screen and (max-height: 1366px) {
```
refi93, line 1: seems like those css files got merged by accident from https://github.com/vacuumlabs/adalite/pull/810 as it wasn't merged yet

app/tests/src/actions/transaction-actions.js
```diff
@@ -28,22 +28,99 @@ it('Calculate fee - shelley', async () => {
...
+      assert.deepEqual(state.sendTransactionSummary.fee, setting.sendTransactionSummary.fee)
+    })
+  )
+})
 
-  mockNet.clean()
+describe('Withdrawal fee calculation', () => {
+  Object.entries(withdrawalSettings).forEach(([name, setting]) =>
+    it(`should calculate fee for tx with ${name}`, async () => {
+      await loadTestWallet()
```
refi93, line 120: setting.state could be passed into loadTestWallet as a parameter and perform the `setStateFn` within it, removing that repetitiveness from the tests themselves

app/tests/src/wallet/account-manager.js
```diff
@@ -26,7 +21,7 @@ const initAccountManager = async (settings, i) => {
     maxAccountIndex,
   } = settings
   const config = {...ADALITE_CONFIG, isShelleyCompatible, shouldExportPubKeyBulk}
-  // console.log(JSON.stringify(settings))
+
```
refi93, line 24: redundant empty space

app/tests/src/wallet/account-manager.js
```diff
@@ -45,62 +40,40 @@ const initAccountManager = async (settings, i) => {
...
-  )
+  return AccountManager({config, cryptoProvider, blockchainExplorer, maxAccountIndex})
 }
 
-before(async () => {
-  for (const accountManagerSetting of accountManagerSettings) {
-    await initAccountManager(accountManagerSetting)
-  }
+before(() => {
+  Object.entries(accountManagerSettings).forEach(([name, setting]) => {
```
refi93, line 47: `initAccountManager` seems to be async so this can lead to flaky tests if I understand correctly. `await Promise.all(Object.entries(accountManagerSettings).map(etc))` should address that

app/tests/src/wallet/account-manager.js
```diff
@@ -109,8 +82,9 @@ describe('account exploration', () => {
   it('should add consequent new account if previous was used', async () => {
     const mockNet = mockNetwork(ADALITE_CONFIG)
     mockNet.mockBulkAddressSummaryEndpoint()
-    const accountsLength = (await accountManagers[1].discoverAccounts()).length
-    const nextAccount = await accountManagers[1].exploreNextAccount()
+    const accountManager = await accountManagers.withDisabledBulkExport
```
refi93, line 85: I would add to the test description a remark clarifying that this concerns the case when bulk export is disabled as with enabled bulk export this wouldn't make much sense given that all accounts should be discovered on the first go

app/tests/src/wallet/account.js
```diff
@@ -39,27 +41,54 @@ const initAccount = async (settings, i) => {
...
+  mockNet.mockPoolMeta()
+  mockNet.mockGetAccountState()
+  mockNet.mockAccountDelegationHistory()
+  mockNet.mockAccountStakeRegistrationHistory()
+  mockNet.mockWithdrawalHistory()
+  mockNet.mockRewardHistory()
+  mockNet.mockPoolRecommendation()
+
+  Object.entries(accountSettings).forEach(([name, setting]) => {
+    accounts[name] = initAccount(setting)
```
refi93, line 72: since initAccount() is async, this may not always work so should be still wrapped in Promise.all and await as I've written above

app/tests/src/actions/transaction-actions.js
PeterBenc: Will do :+1:

app/tests/src/wallet/account-manager.js
PeterBenc: Will do :+1:

app/public/css/0-1366px.css
PeterBenc: Yeah, you started reviewing just as I was rebasing it on the css PR, that's all.

app/tests/src/actions/transaction-actions.js
PeterBenc: Good idea :+1:

app/tests/src/wallet/account-manager.js
PeterBenc: I ll fix it.

app/tests/src/wallet/account-manager.js
PeterBenc: Yes, the `initAccountManager` is still async, but I decided not to use `Promise.all()` as you suggest, since it returns an array and also expects an array of promises, and that would forfeit the whole idea behind having the settings as an object. Therefore I decided to store the AccountManagers as Promises. Anyway I could refactor it to for cycle to keep the object behaviour and store awaited values instead of promises. What do you think?

app/tests/src/wallet/account.js
PeterBenc: Im not sure what you mean by 'this may not always work`. I decided to store Accounts as promises and they need to be awaited when the test is executing but in that case, everything should work always..right? :D

app/tests/src/wallet/account-manager.js
PeterBenc: Will do.

app/tests/src/wallet/account-manager.js
refi93: Promise.all awaits on an array of promises, the returned result is just a byproduct which we don't really care about here as you are modifying an object from the outer scope anyway, but a for loop with awaits is also an option of course and arguably more readable though the linter may complain if this rule is enabled which may be as in production this usually results in suboptimal performace given that the promises are chained instead of executed in parallel which doesn't really matter here: https://eslint.org/docs/rules/no-await-in-loop

app/tests/src/wallet/account.js
refi93: ah, I see - what's the benefit of that though if you are awaiting on it anyway in tests below? Also seems like a chance to shoot ourselves in the foot in the future especially since the tests have not been migrated to ts yet so we can't add an explicit type annotation to the accounts object yet.

app/tests/src/wallet/account-manager.js
PeterBenc: Okay, `Promise.all` it is :+1: Thnx

app/tests/src/wallet/account.js
PeterBenc: You are right, I should have at least renamed it to something like `accountPromises`, anyway I ll refactor it to Promise.all and we don't need to await anything,

app/tests/src/wallet/account.js
```diff
@@ -39,27 +41,57 @@ const initAccount = async (settings, i) => {
...
+  mockNet.mockAccountDelegationHistory()
+  mockNet.mockAccountStakeRegistrationHistory()
+  mockNet.mockWithdrawalHistory()
+  mockNet.mockRewardHistory()
+  mockNet.mockPoolRecommendation()
+
+  await Promise.all(
+    Object.entries(accountSettings).map(async ([name, setting]) => {
+      accounts[name] = await initAccount(setting)
+      return accounts[name]
```
refi93, line 74: I would not return anything to make it clear we are interested only about the completion of the promise and saving a redundant statement

app/tests/src/common/address-manager-settings.js
```diff
@@ -51,23 +51,28 @@ const addressManagerSettings = [
   },
 ]
 
-const addressManagerSettings2 = [
-  {
+// these are used for the new tests
+const addressManagerSettings2 = {
```
refi93, line 55: "new" is a very relative term in the codebase - I'd rather name this `addressManagerSettings` and the "old" `addressManagerSettings` to `addressManagerSettings_deprecated` or something - if we really want to get rid of those. And if not, we should probably find appropriate name for both separately as they even have different fields.

In general, "settings" is quite an overloaded term and I find it more suitable to name it "fixtures" and split the "parameters" from "expected" results as currently it's really confusing having both mixed, but that's probably worth addressing separately.

app/tests/src/wallet/account.js
PeterBenc: Hmm, you are right, I was of the thought that I have to return the promise for `Promise.all` to be able to await it but surprisingly not. Thnx.

app/tests/src/common/address-manager-settings.js
PeterBenc: The comment was mostly for me tbh since I expected that if anyone, I ll be working on this further.. :D but I ll update the comment and fix the names. Regarding the `settings` I was only following the previous implementation and I agree, we can address that separately.

PeterBenc: I decided to remove the older Shelley tests altogether since they are testing the same stuff as the new ones anyway.

refi93 (approved):

PR #815, opened 2021-02-01
Type cardano network

PeterBenc:
Motivation: 
Previously the `constants.ts` file in wallet dir contained all kinds of networks that are no longer used and none of it was typed.

Changes:

- removes old networks from the `NETWORKS` constant
- add a type for `Network`, `NetworkId` and `ProtocolMagic`


app/frontend/wallet/constants.ts
```diff
@@ -15,88 +16,36 @@ export const TX_WITNESS_SIZES = {
...
-    },
+
+export const NETWORKS: {[key: string]: Network} = {
+  MAINNET: {
+    name: 'mainnet',
+    networkId: NetworkId.MAINNET,
+    protocolMagic: ProtocolMagic.MAINNET,
+    eraStartSlot: 4492800, // 21600 slot x 208 epochs
+    eraStartDateTime: Date.parse('29 Jul 2020 21:44:51 UTC'),
+    ttl: 3600,
```
refi93, line 39: isn't that rather something like `defaultTtl`?

app/frontend/wallet/types.ts
```diff
@@ -0,0 +1,12 @@
+import {NetworkId, ProtocolMagic} from './constants'
```
refi93, line 1: I'd keep the NetworkId and ProtocolMagic types here together with other enums in constants to separate true constans from typing-related abstractions. This would also get us rid of the current cyclic import between types.ts ans constants.ts

PeterBenc: Will do, I was having a dilemma myself whether to define enums in `constants` file or `types` file. I agree let's have enums in `constants` file since as you said, they are typing-related abstractions and also remove the cyclic import. :+1: Thnx.

app/frontend/wallet/constants.ts
PeterBenc: I agree, I was thinking it doesnt have to be bound to a specific network, but then I thought, `ttlSeconds` might not be the best name since it just only happens that the current network has one slot per second but some other network may have another number of slots per second. Additionally the current `ttl` represents number of slot we add to the current slot number so if it makes sense to you I ll rename it to `ttlSlots`. Im not that happy with this name either so Im really open to suggestions. What do you think? :D

PeterBenc: Hm, but instead of defining the number of slots we add to the current slot we could define how many slots are per second (in this case 1) and have another constant for seconds and helper function that returns slotNumber appropriate for one hour from now. Anyway we can do this in a another PR.

PeterBenc: also here 
```
return cryptoProvider.network.eraStartSlot + timePassed + cryptoProvider.network.ttl
```
we are adding the number of passed seconds + ttl which would break if the `slot per second` ratio was different. We could fix that as well.

PeterBenc:
PS: I ll squash the commits before merging.

refi93 (approved):

PR #817, opened 2021-02-02
Fix error handling

PeterBenc:
Motivation: 

- Lot of ledger related errors come to sentry daily, and all we can do is to advice users to try using webUsb or to our trouble shooting guide. 
- If a network error happens during the fee calculation it results in unexpected error since its not cached.

Changes: 

- catch errors when calculating fee
- add translations for general `Error` and modify the `captureBySentry` helper to send also the params into translations so if the for example the `Error` is present in the translations but doest have a translation matching its params it returns undefined thus resulting in the error being considered unexpected.
- add the ledger errors to `errorsWithHelp` so it automatically recommends user to read the troubleshooting guide, and if that doesn't help he can send us the error if he really wants to.

Gotcha: 

- We should probably refactor and type the translations and also sentry capturing logic a bit but this is mostly a hotfix so we don't have to respond manually to every user.
- I also update our troubleshooting guide and added a ledger section
- the `SyntaxError: Failed to execute 'postMessage' on 'Window'` is a issue for quite some time and I havent been able to find any fix or reasoning behind it happening, only that it has to do something with some build in chrome crypto extension. Anyway the only thing I  usually respond to the users is to checkout the trouble shooting guide anyway.

What user will see from now on:
 
image

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -603,17 +603,24 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
 
   const prepareTxPlan = async (args) => {
     const state = getState()
-    const plan: any = await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
-    // FIXME: this cant be any
-    if (plan.error) {
-      stopLoadingAction(state, {})
-      resetDelegationWithoutHash(state)
-      setState({
-        calculatingDelegationFee: false,
-        calculatingFee: false,
-      })
+    try {
+      const plan: any = await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
+      // FIXME: this cant be any
+      if (plan.error) {
+        stopLoadingAction(state, {})
+        resetDelegationWithoutHash(state)
+        setState({
+          calculatingDelegationFee: false,
+          calculatingFee: false,
+        })
+      }
+      return plan
+    } catch (e) {
+      return {
+        estimatedFee: 0,
+        error: {code: e.name},
+      }
     }
-    return plan
   }
 
   const calculateFee = async () => {
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -616,6 +616,10 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
       }
       return plan
     } catch (e) {
+      if (e.name !== 'NetworkError' || e.name !== 'ServerError') {
+        throw e
+      }
+
       return {
         estimatedFee: 0,
         error: {code: e.name},
--- a/app/frontend/translations.ts
+++ b/app/frontend/translations.ts
@@ -6,6 +6,9 @@ import {Lovelace, CryptoProviderFeature} from './types'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 
+const ledgerTroubleshootingSuggestion =
+  'If you are using Ledger, please try connecting your device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger"). For more information please read the section concerning Ledger in our troubleshooting suggestions.'
+
 const translations = {
   SendAddressInvalidAddress: () => 'Invalid address',
   SendAmountIsNan: () => 'Invalid format: Amount has to be a number',
@@ -55,6 +58,10 @@ const translations = {
     }
     return `TransportInterfaceNotAvailable: ${message} ${errors[message] || ''}`
   },
+  DisconnectedDeviceDuringOperation: () =>
+    `DisconnectedDeviceDuringOperation: ${ledgerTroubleshootingSuggestion}`,
+  TransportWebUSBGestureRequired: () =>
+    `TransportWebUSBGestureRequired: ${ledgerTroubleshootingSuggestion}`,
 
   TransactionRejectedByNetwork: () =>
     'TransactionRejectedByNetwork: Submitting the transaction into Cardano network failed. We received this error and we will investigate the cause.',
@@ -130,6 +137,19 @@ const translations = {
   PoolRegNoTtl: () =>
     'TTL parameter is missing in the transaction. It is explicitly required even for the Allegra era.',
   PoolRegTxParserError: ({message}) => `Parser error: ${message}`,
+
+  Error: ({message}) => {
+    const errors = {
+      'NotFoundError: The device was disconnected.': `${message}${ledgerTroubleshootingSuggestion}`,
+      'AbortError: The transfer was cancelled.': `${message}${ledgerTroubleshootingSuggestion}`,
+      // an issue with CryptoToken extension allowing 2-step verification
+      // https://askubuntu.com/questions/844090/what-is-cryptotokenextension-in-chromium-extensions
+      "SyntaxError: Failed to execute 'postMessage' on 'Window': Invalid target origin 'chrome-extension://kmendfapggjehodndflmmgagdbamhnfd' in a call to 'postMessage'": `${message}${ledgerTroubleshootingSuggestion}`,
+    }
+    // we return undefined in case of unmached message on purpose since we
+    // want to treat such errors as unexpected
+    return errors[message]
+  },
 }
 
 function getTranslation(code, params = {}) {
--- a/app/frontend/components/pages/advanced/keys.tsx
+++ b/app/frontend/components/pages/advanced/keys.tsx
@@ -64,7 +64,7 @@ const Keys = ({byronAccountXpub, shelleyAccountXpub, stakingAddress, stakingXpub
       </div>
       <div className="advanced-value">{stakingAddress}</div>
       <div className="advanced-label">
-        Staking key hex <LinkIconToKey stakeKey={bechAddressToHex(stakingAddress)} />
+        Staking key hex <LinkIconToKey stakeKey={stakingAddress} />
       </div>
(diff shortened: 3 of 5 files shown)


app/frontend/translations.ts
```diff
@@ -6,6 +6,9 @@ import {Lovelace, CryptoProviderFeature} from './types'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 
+const ledgerConnectError =
+  ' Please try connecting you device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger"). For more information please follow the Ledger part of our troubleshooting suggestions.'
```
refi93, line 10: probably redundant space at the beginning and typo - `you` -> `your`

PeterBenc: Fixed, thnx :+1:

app/frontend/translations.ts
```diff
@@ -130,6 +136,15 @@ const translations = {
   PoolRegNoTtl: () =>
     'TTL parameter is missing in the transaction. It is explicitly required even for the Allegra era.',
   PoolRegTxParserError: ({message}) => `Parser error: ${message}`,
+
+  Error: ({message}) => {
+    const errors = {
+      'NotFoundError: The device was disconnected.': `${message}${ledgerConnectError}`,
+      'AbortError: The transfer was cancelled.': `${message}${ledgerConnectError}`,
+      "SyntaxError: Failed to execute 'postMessage' on 'Window': Invalid target origin 'chrome-extension://kmendfapggjehodndflmmgagdbamhnfd' in a call to 'postMessage'": `${message}`,
```
refi93, line 144: where does that error message come from? Not sure if that extension identifier is global

app/frontend/actions.ts
```diff
@@ -603,17 +603,24 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-    const plan: any = await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
-    // FIXME: this cant be any
-    if (plan.error) {
-      stopLoadingAction(state, {})
-      resetDelegationWithoutHash(state)
-      setState({
-        calculatingDelegationFee: false,
-        calculatingFee: false,
-      })
+    try {
```
refi93, line 606: what's the purpose of this try/catch? what kind of error can happen here that it's required?

app/frontend/actions.ts
```diff
@@ -603,17 +603,24 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+      if (plan.error) {
+        stopLoadingAction(state, {})
+        resetDelegationWithoutHash(state)
+        setState({
+          calculatingDelegationFee: false,
+          calculatingFee: false,
+        })
+      }
+      return plan
+    } catch (e) {
```
refi93, line 618: I would restrict this to network errors if that's our concern to not silence other exceptions that may be e.g. related to some bug or other kind of malfuncion

app/frontend/translations.ts
PeterBenc: I mentioned it briefly in the PR description, its pretty rare and regarding the extension identifier, I haven't checked it but I was researching it couple of times I always came to that one crypto-something extension in chrome that somehow causes this.

app/frontend/translations.ts
```diff
@@ -6,6 +6,9 @@ import {Lovelace, CryptoProviderFeature} from './types'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 
+const ledgerConnectError =
```
refi93, line 9: ledgerTroubleshootingSuggestion is imho a more accurate name as it's not an error per se

app/frontend/translations.ts
```diff
@@ -130,6 +136,15 @@ const translations = {
...
     'TTL parameter is missing in the transaction. It is explicitly required even for the Allegra era.',
   PoolRegTxParserError: ({message}) => `Parser error: ${message}`,
+
+  Error: ({message}) => {
+    const errors = {
+      'NotFoundError: The device was disconnected.': `${message}${ledgerConnectError}`,
+      'AbortError: The transfer was cancelled.': `${message}${ledgerConnectError}`,
+      "SyntaxError: Failed to execute 'postMessage' on 'Window': Invalid target origin 'chrome-extension://kmendfapggjehodndflmmgagdbamhnfd' in a call to 'postMessage'": `${message}`,
+    }
+    return errors[message]
```
refi93, line 146: this can get resolved to undefined if the message is lacking which is probably not desired, something like

```
return (message in errors) ? errors[message] ? message
```

PeterBenc: Actually, the undefined behavior is desired, we check whether an error has a translation and if not we display unexpectedErrorModal, so its crucial this returns undefined, otherwise all generic errors with the name `Error` would be "expected".

app/frontend/actions.ts
PeterBenc: You are right, I didnt get to respond to the previous comment but our concern are the network errors indeed. All the other errors that might occur would still be displayed somewhere but you are right that they wouldn't be "unexpected". I ll add  
```
If (e.name !== 'NetworkError` || e.name !== 'ServerError') throw e
```

app/frontend/translations.ts
refi93: I see - that's probably worth explicitly acknowledging with a comment so nobody is tempted to "fix" the function interface

app/frontend/translations.ts
refi93: I see, it's apparently the CryptoToken extension allowing 2-step verification, likely via U2F: https://askubuntu.com/questions/844090/what-is-cryptotokenextension-in-chromium-extensions
From sentry it seems that the users that had that issue so far were using Ledger but of course, not necessarily. Maybe I would reword the troubleshooting suggestion to "If you are using Ledger Nano, ..." and put it here as well

app/frontend/actions.ts
PeterBenc: I definitely plan to refactor this before the multi-asset so this whole tx plan creation works a bit better :D

app/frontend/translations.ts
refi93: and please clarify with a comment where that arbitrary string comes from and why it's here

app/frontend/translations.ts
PeterBenc: You are right, thanks for the suggestion, we should refactor this as well, but as mentioned in the description, this is rather a hotfix.

app/frontend/translations.ts
PeterBenc: Thanks, I should have commented it. :+1:

app/frontend/translations.ts
refi93: hotfixes without appropriate TODO comments have high chances of becoming permanent fixes :D

PeterBenc: True.. :D

PeterBenc:
PS: I ll squash the commits before merging.

app/frontend/translations.ts
```diff
@@ -6,6 +6,9 @@ import {Lovelace, CryptoProviderFeature} from './types'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 
+const ledgerTroubleshootingSuggestion =
+  'If you are using Ledger, please try connecting your device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger").For more information please follow section concerning Ledger in our troubleshooting suggestions.'
```
refi93, line 10:
```suggestion
  'If you are using Ledger, please try connecting your device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger"). For more information please check the section concerning Ledger in our troubleshooting suggestions.'
```

app/frontend/translations.ts
```diff
@@ -130,6 +137,19 @@ const translations = {
...
+  Error: ({message}) => {
+    const errors = {
+      'NotFoundError: The device was disconnected.': `${message}${ledgerTroubleshootingSuggestion}`,
+      'AbortError: The transfer was cancelled.': `${message}${ledgerTroubleshootingSuggestion}`,
+      // an issue with CryptoToken extension allowing 2-step verification
+      // https://askubuntu.com/questions/844090/what-is-cryptotokenextension-in-chromium-extensions
+      "SyntaxError: Failed to execute 'postMessage' on 'Window': Invalid target origin 'chrome-extension://kmendfapggjehodndflmmgagdbamhnfd' in a call to 'postMessage'": `${message}${ledgerTroubleshootingSuggestion}`,
+    }
+    // we return undefined in case of unmached message on purpose since we
+    // want to threat such errors as unexpected
```
refi93, line 150:
```suggestion
    // want to treat such errors as unexpected
```

app/frontend/actions.ts
```diff
@@ -603,17 +603,25 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+      if (e.name !== 'NetworkError' || e.name !== 'ServerError') throw e
```
refi93, line 619: I would rather wrap the statement in curly braces so the return statement right below doesn't get confused as the logic triggered by the condition

```suggestion
      if (e.name !== 'NetworkError' || e.name !== 'ServerError') {
        throw e
      }

```

refi93 (approved):
last few nitpicks, otherwise LGTM

PR #819, opened 2021-02-03
Type blockchain explorer

xdzurman:
Assigns types to data retrieved from BE endpoints. Leaves space for future refactors. Only patches up new typescript errors due to inconsistency of the state, however, such changes and refactors are out of scope of this PR.

Possible changes:
- WHERE (which file) to store the new types
- Split up purely BE types and any follow-up types of stuff created freshly on FE

### DIFF
--- a/app/frontend/wallet/blockchain-explorer.ts
+++ b/app/frontend/wallet/blockchain-explorer.ts
@@ -15,6 +15,24 @@ import distinct from '../helpers/distinct'
 import {UNKNOWN_POOL_NAME} from './constants'
 import {HexString, Lovelace} from '../types'
 import {captureMessage} from '@sentry/browser'
+import {
+  BulkAddressesSummaryResponse,
+  TxSummaryResponse,
+  SubmitResponse,
+  BulkAdressesUtxoResponse,
+  HostedPoolMetadata,
+  DelegationHistoryEntry,
+  RewardsHistoryEntry,
+  WithdrawalsHistoryEntry,
+  StakeRegistrationHistoryEntry,
+  NextRewardDetail,
+  ValidStakePools,
+  NextRewardDetailsFormatted,
+  RewardWithMetadata,
+  PoolRecommendationResponse,
+  AccountInfoResponse,
+  BestSlotResponse,
+} from './explorer-types'
 
 const cacheResults = (maxAge: number, cache_obj: Object = {}) => <T extends Function>(fn: T): T => {
   const wrapped = (...args) => {
@@ -34,24 +52,30 @@ const cacheResults = (maxAge: number, cache_obj: Object = {}) => <T extends Func
 const blockchainExplorer = (ADALITE_CONFIG) => {
   const gapLimit = ADALITE_CONFIG.ADALITE_GAP_LIMIT
 
-  async function _fetchBulkAddressInfo(addresses) {
+  async function _fetchBulkAddressInfo(addresses: Array<string>) {
     const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/bulk/addresses/summary`
-    const result = await request(url, 'POST', JSON.stringify(addresses), {
-      'Accept': 'application/json',
-      'Content-Type': 'application/json',
-    })
+    const result: BulkAddressesSummaryResponse = await request(
+      url,
+      'POST',
+      JSON.stringify(addresses),
+      {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+      }
+    )
+    // @ts-ignore (TODO, handle 'Left')
     return result.Right
   }
 
-  const getAddressInfos = cacheResults(5000)(_fetchBulkAddressInfo)
+  const _getAddressInfos = cacheResults(5000)(_fetchBulkAddressInfo)
 
-  async function getTxHistory(addresses) {
+  async function getTxHistory(addresses: Array<string>) {
     const transactions = []
     const chunks = range(0, Math.ceil(addresses.length / gapLimit))
     const cachedAddressInfos = (await Promise.all(
       chunks.map(async (index) => {
         const beginIndex = index * gapLimit
-        return await getAddressInfos(addresses.slice(beginIndex, beginIndex + gapLimit))
+        return await _getAddressInfos(addresses.slice(beginIndex, beginIndex + gapLimit))
       })
     )).reduce(
       (acc, elem) => {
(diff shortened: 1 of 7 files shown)


app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -87,26 +111,28 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
   async function fetchTxRaw(txId) {
     // eslint-disable-next-line no-undef
     const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/txs/raw/${txId}`
     const result = await request(url)
     return Buffer.from(result.Right, 'hex')
   }
 
-  async function isSomeAddressUsed(addresses) {
-    return (await getAddressInfos(addresses)).caTxNum > 0
+  async function isSomeAddressUsed(addresses): Promise<boolean> {
```
refi93, line 130:
```suggestion
  async function isSomeAddressUsed(addresses: Array<string>): Promise<boolean> {
```

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -357,7 +391,10 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     }
   }
 
-  function getPoolRecommendation(poolHash: string, stake: number): Promise<any> {
+  function getPoolRecommendation(
+    poolHash: string,
+    stake: number
```
refi93, line 396: do you know what's this "stake" parameter? seems pretty cryptic to me, so maybe it would be nice to give it some more descriptive name

app/frontend/wallet/explorer-types.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export type StakeRegistrationHistoryEntry = {
+  txHash: string
+  time: string
+  epochNo: number
+  action: 'registration' | 'deregistration'
+}
+
+export type ValidStakePools = {
```
refi93, line 51: I'd name this `ValidStakePoolsMapping` to make it clear it's not an array/collection and I would create a separate type for the object value, something like `StakePoolProps`

The ideal endgame would imho be just returning a list of stake pools and whoever needs to look them up by id could just sequentially search it or delegate it to some function which would be optimized for that should that be an actual bottleneck

refi93 (changes requested):
see comments, mostly minor ideas for improvement

app/frontend/wallet/explorer-types.ts
xdzurman: Done. Originally, we went for the Map due to many places which needed to access it, and at that time poolId was sufficient. Therefore we decided that O(1) lookup by just the poolId would be ideal for the time being.

PeterBenc: btw @DavidTranDucVL is already working on a refactor regarding this validStakePools Object btw, so we address any suggestions in that PR then.

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -87,26 +111,28 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
-    return (await getAddressInfos(addresses)).caTxNum > 0
+  async function isSomeAddressUsed(addresses: Array<string>): Promise<boolean> {
+    return (await _getAddressInfos(addresses)).caTxNum > 0
   }
 
   // TODO: we should have an endpoint for this
-  async function filterUsedAddresses(addresses: Array<String>) {
+  async function filterUsedAddresses(addresses: Array<string>) {
     const txHistory = await getTxHistory(addresses)
     const usedAddresses = new Set()
```
PeterBenc, line 137: This is currently returning a `Set<unknown>` 
```suggestion
    const usedAddresses = new Set<string>()
```

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -372,16 +409,16 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     const url = `${
       ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL
     }/api/account/info/${stakingKeyHashHex}`
-    const response = await request(url)
+    const response: AccountInfoResponse = await request(url)
```
PeterBenc, line 412: Maybe someone more with more typescript can say, but personally I would not type the `const`, I would type the return statement of the function. I get that `request` returns `Promise<any> and there is not much to do about that so we have to type the const, but still I would type the return statement on all functions as well. But as I said, someone more experienced could comment :D

PeterBenc (changes requested):
Can we do something about `_fetchBulkAddressInfo`, `getTxHistory` ,`fetchTxInfo`,`getBalance`,`fetchUnspentTxOutputs` returning `Promise<any>` ? 

Also, I have to admit I didn't investigate whats the reasoning behind typing something with the `Response` suffix and other things without. For example, it makes sense to me to type the result of transaction submission as `TxSubmissionResponse` but it doesn't make sense for validStakePools for example. I am not saying that anything is wrong but just a note to think about. Maybe if you have to access the result by `.Right` the type has the `Response` suffix, otherwise not. :thinking:

xdzurman:
> Can we do something about `_fetchBulkAddressInfo`, `getTxHistory` ,`fetchTxInfo`,`getBalance`,`fetchUnspentTxOutputs` returning `Promise<any>` ?
> 
> Also, I have to admit I didn't investigate whats the reasoning behind typing something with the `Response` suffix and other things without. For example, it makes sense to me to type the result of transaction submission as `TxSubmissionResponse` but it doesn't make sense for validStakePools for example. I am not saying that anything is wrong but just a note to think about. Maybe if you have to access the result by `.Right` the type has the `Response` suffix, otherwise not.

As stated in the description, the goal of this PR is to type things coming from the backend. Also the whole backend explorer could use a lot of refactoring, that's why I did not want to touch _fetchBulkAddressInfo, getTxHistory, etc. Everything with a Response suffix comes from backend.

app/frontend/wallet/blockchain-explorer.ts
xdzurman: Check the other comment

PeterBenc:
> > Can we do something about `_fetchBulkAddressInfo`, `getTxHistory` ,`fetchTxInfo`,`getBalance`,`fetchUnspentTxOutputs` returning `Promise<any>` ?
> > Also, I have to admit I didn't investigate whats the reasoning behind typing something with the `Response` suffix and other things without. For example, it makes sense to me to type the result of transaction submission as `TxSubmissionResponse` but it doesn't make sense for validStakePools for example. I am not saying that anything is wrong but just a note to think about. Maybe if you have to access the result by `.Right` the type has the `Response` suffix, otherwise not.
> 
> As stated in the description, the goal of this PR is to type things coming from the backend. Also the whole backend explorer could use a lot of refactoring, that's why I did not want to touch _fetchBulkAddressInfo, getTxHistory, etc., and just type the things coming from the BE for a start. Perhaps the PR name is misleading in that way.
> 
> Everything with a Response suffix comes from backend. Although not all endpoints return Right/Left wrapper. I stated it as a possible change up to discussion - to separate truly BE types and the new FE types after data transformation. In that scenario it makes sense to type the consts, so that it's clear what is supposed to come from BE and into what we're transforming it (separate FE type).

I see, sorry then for misunderstanding the scope of this PR, if I understand correctly then, everything coming as a return statement of `request` should have the `Response` suffix (that's great :+1: )  and we should type the return types of the functions in another PR so we can use them in other components like wallet, account etc.  Thnx for the explanation.

refi93 (approved):

PR #825, opened 2021-02-09
Optimilize-account-info

PeterBenc:
Changes: 
Removes unnessary calls to poolInfo and uses it only as a fallback
When exploring accounts, only external addresses are checked 

Gotcha: 
These are just hot fixed so we dont send so many requests to poolMeta, (which can slow the page a lot). Since both @xdzurman and @DavidTranDucVL  are working on blockchain explorer I didnt try to refactor any of the current behavior and left it for their PRs.

### DIFF
--- a/app/frontend/wallet/blockchain-explorer.ts
+++ b/app/frontend/wallet/blockchain-explorer.ts
@@ -230,20 +230,19 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     return response
   }
 
-  async function getStakingHistory(accountPubkeyHex, validStakepools) {
-    // Urls
+  async function getStakingHistory(stakingKeyHashHex: HexString, validStakepools) {
     const delegationsUrl = `${
       ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL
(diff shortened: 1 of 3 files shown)


app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -261,7 +260,9 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
       validStakepools[poolHash] ? validStakepools[poolHash].url : null
 
     const poolMetaUrls = distinct(
-      [...delegations, ...rewards].map(({poolHash}) => extractUrl(poolHash))
+      [...delegations, ...rewards]
+        .filter(({name}) => !name)
+        .map((historyEntry) => extractUrl(historyEntry.poolHash))
```
DavidTranDucVL, line 265: Isn't object destructing more clear and shorter syntax?

DavidTranDucVL (approved):
LGTM

app/frontend/wallet/blockchain-explorer.ts
PeterBenc: It is, my bad I changed it during the refactor and forgot to change it back. I ll fix it.

xdzurman (approved):

PR #828, opened 2021-02-10
Update ledger lib

PeterBenc:
Changes: 
- Updates ledger lib and ledger transport libraries
- Capture message to sentry if the transport creation fails, for debugging purposes.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -24,11 +24,11 @@
     "postinstall": "node ./helpers/removeChainLibsWasmImport"
   },
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "^2.1.0",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "^2.2.0",
     "@emurgo/js-chain-libs": "https://github.com/SebastienGllmt/js-chain-libs-pkg",
-    "@ledgerhq/hw-transport": "^5.36.0",
+    "@ledgerhq/hw-transport": "^5.43.0",
     "@ledgerhq/hw-transport-u2f": "^5.36.0-deprecated",
-    "@ledgerhq/hw-transport-webusb": "^5.36.0",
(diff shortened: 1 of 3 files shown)


app/package.json
```diff
@@ -24,11 +24,11 @@
     "postinstall": "node ./helpers/removeChainLibsWasmImport"
   },
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "^2.1.0",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "https://github.com/vacuumlabs/ledgerjs-cardano-shelley/releases/download/v2.2.0/cardano-foundation-ledgerjs-hw-app-cardano-2.2.0.tgz",
```
refi93, line 27: you can already use the npm version

app/package.json
```diff
@@ -24,11 +24,11 @@
...
     "@emurgo/js-chain-libs": "https://github.com/SebastienGllmt/js-chain-libs-pkg",
-    "@ledgerhq/hw-transport": "^5.36.0",
+    "@ledgerhq/hw-transport": "^5.43.0",
     "@ledgerhq/hw-transport-u2f": "^5.36.0-deprecated",
```
refi93, line 30: wondering for how long we would be able to keep this deprecated transport version drifting from the rest :thinking: but as far as I tested it works

refi93 (approved):
LGTM, just update to the npm version of ledgerjs please

app/package.json
PeterBenc: I checked and also myEtherWallet uses this deprecated version. We should definitely dedicate some time researching these transports tho.

refi93: yes, I checked independently and it's the latest version which makes sense as it's deprecated, so as long as it works fine, just wanted to point out that eventually, we would likely need to remove it altogether due to slippage from the rest of transport libraries, hopefully until then some of the other transports gets more widely adopted

PR #829, opened 2021-02-10
Support testnet

PeterBenc:
Changes: 
- Refactors address encoding so it works with testnet addresses and happens in one place. 
- Adds new network for mary

Todo:
- Fix all parameters in the network constant.
- Fix failing tests

### DIFF
--- a/app/frontend/wallet/account.ts
+++ b/app/frontend/wallet/account.ts
@@ -1,6 +1,6 @@
 import AddressManager from './address-manager'
 import PseudoRandom from './helpers/PseudoRandom'
-import {MAX_INT32} from './constants'
+import {DEFAULT_TTL_SLOTS, MAX_INT32} from './constants'
 import NamedError from '../helpers/NamedError'
 import {
   AddressToPathMapping,
@@ -148,7 +148,9 @@ const MyAddresses = ({
     }
   }
 
-  async function areAddressesUsed() {
+  async function areAddressesUsed(): Promise<boolean> {
+    // we check only the external addresses since internal should not be used before external
+    // https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit
     const baseExt = await baseExtAddrManager._deriveAddresses(0, gapLimit)
     return await blockchainExplorer.isSomeAddressUsed(baseExt)
   }
@@ -206,7 +208,7 @@ const Account = ({
     blockchainExplorer,
   })
 
-  async function ensureXpubIsExported() {
+  async function ensureXpubIsExported(): Promise<void> {
     // get first address to ensure that public key was exported
     await myAddresses.baseExtAddrManager._deriveAddress(0)
   }
@@ -215,10 +217,10 @@ const Account = ({
     // TODO: move to wallet
     try {
       const bestSlot = await blockchainExplorer.getBestSlot().then((res) => res.Right.bestSlot)
-      return bestSlot + cryptoProvider.network.ttlSlots
+      return bestSlot + DEFAULT_TTL_SLOTS
     } catch (e) {
       const timePassed = Math.floor((Date.now() - cryptoProvider.network.eraStartDateTime) / 1000)
-      return cryptoProvider.network.eraStartSlot + timePassed + cryptoProvider.network.ttlSlots
+      return cryptoProvider.network.eraStartSlot + timePassed + DEFAULT_TTL_SLOTS
     }
   }
 
@@ -315,6 +317,9 @@ const Account = ({
   }
 
   function isAccountUsed(): Promise<boolean> {
+    //TODO: we should decouple shelley compatibility from usedness
+    // in case the wallet is not shelley compatible we consider the
+    // the first account not used
     return config.isShelleyCompatible && myAddresses.areAddressesUsed()
   }
 
--- a/app/frontend/wallet/constants.ts
+++ b/app/frontend/wallet/constants.ts
@@ -26,11 +26,21 @@ export const NETWORKS: {[key: string]: Network} = {
     protocolMagic: ProtocolMagic.MAINNET,
     eraStartSlot: 4492800, // 21600 slot x 208 epochs
     eraStartDateTime: Date.parse('29 Jul 2020 21:44:51 UTC'),
-    ttlSlots: 3600,
+    epochsToRewardDistribution: 4,
+    minimalOutput: 1000000,
+  },
+  MARY_TESTNET: {
+    name: 'mary-testnet',
+    networkId: NetworkId.TESTNET,
+    protocolMagic: ProtocolMagic.MARY_TESTNET,
+    eraStartSlot: 0,
+    eraStartDateTime: Date.parse('24 Jul 2019 20:20:16 UTC'),
     epochsToRewardDistribution: 4,
     minimalOutput: 1000000,
   },
 }
+
+export const DEFAULT_TTL_SLOTS = 3600
 
 export const DELAY_AFTER_TOO_MANY_REQUESTS = 2000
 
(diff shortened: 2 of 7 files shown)


xdzurman (commented):
I'll try this out with testnet BE tomorrow and et you know if I have more comments

app/frontend/wallet/constants.ts
```diff
@@ -30,6 +30,16 @@ export const NETWORKS: {[key: string]: Network} = {
     epochsToRewardDistribution: 4,
     minimalOutput: 1000000,
   },
+  MARY_TESTNET: {
+    name: 'mary-testnet',
+    networkId: NetworkId.TESTNET,
+    protocolMagic: ProtocolMagic.MARY_TESTNET,
+    eraStartSlot: 4492800, // 21600 slot x 208 epochs
+    eraStartDateTime: Date.parse('29 Jul 2020 21:44:51 UTC'),
+    ttlSlots: 3600,
```
xdzurman, line 39: I think we can drop all ttl-related consts, since it's optional since Allegra, but I'll let u know

PeterBenc: Tbh, I think it's a nice feature to be able to know for sure that if anything happens with the tx, one hour later it's definitely not valid. I wouldn't remove this even tho its optional.

app/frontend/wallet/types.ts
```diff
@@ -1,11 +1,12 @@
 export const enum NetworkId {
   MAINNET = 1,
-  HASKELL_TESTNET = 0,
+  TESTNET = 0,
 }
 
 export const enum ProtocolMagic {
   MAINNET = 764824073,
   HASKELL_TESTNET = 42,
+  MARY_TESTNET = 3,
```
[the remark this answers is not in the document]
PeterBenc: Thnx.

app/tests/src/shelley.js
```diff
@@ -28,11 +29,11 @@ describe('shelley address derivation', () => {
   const mnemonic15Words =
     'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon address'
   it('should derive base address from 15-words mnemonic', async () => {
-    const cp = await getCryptoProvider(mnemonic15Words, 0) //TODO: change discriminator to networkId for shelley
+    const cp = await getCryptoProvider(mnemonic15Words, NetworkId.TESTNET) //TODO: change discriminator to networkId for shelley
     const addrGen = ShelleyBaseAddressProvider(cp, 0, false)
     const {address} = await addrGen(0)
     const expected =
-      'addr1qzz6hulv54gzf2suy2u5gkvmt6ysasfdlvvegy3fmf969y7r3y3kdut55a40jff00qmg74686vz44v6k363md06qkq0qk0f2ud'
+      'addr_test1qzz6hulv54gzf2suy2u5gkvmt6ysasfdlvvegy3fmf969y7r3y3kdut55a40jff00qmg74686vz44v6k363md06qkq0qy0adz0'
```
PeterBenc, line 36: Seems strange I know but it seems the test was wrong all along, creating testnet address without the proper prefix.

app/frontend/wallet/constants.ts
refi93: arguably, we can extract that from the config though into a regular constant like defaultTtlSlots or something, I don't really see it being tweaked across network, getting just duplicated all over the different network configs

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -69,4 +86,5 @@ export const isByron = (address: HexString): boolean => {
 }
 
 export const addressToHex = (address: string): HexString =>
+  // TODO: this is pretty fragile, we should check for the validity of the address
```
refi93, line 89: not sure if validate here necessarily - we should be able to leverage typescript to narrow the type of the given string to _Address which this function would expect and then this should no longer be as fragile

app/frontend/wallet/constants.ts
```diff
@@ -30,6 +30,16 @@ export const NETWORKS: {[key: string]: Network} = {
     epochsToRewardDistribution: 4,
     minimalOutput: 1000000,
   },
+  MARY_TESTNET: {
+    name: 'mary-testnet',
+    networkId: NetworkId.TESTNET,
+    protocolMagic: ProtocolMagic.MARY_TESTNET,
+    eraStartSlot: 4492800, // 21600 slot x 208 epochs
+    eraStartDateTime: Date.parse('29 Jul 2020 21:44:51 UTC'),
```
refi93, line 38: are we sure eraStartSlot and eraStartDateTime  constants are the same for Mary testnet?  If it's not relevant, do we need them anymore?

app/frontend/wallet/account.ts
```diff
@@ -206,9 +208,11 @@ const Account = ({
     blockchainExplorer,
   })
 
-  async function ensureXpubIsExported() {
+  async function ensureXpubIsExported(): Promise<void> {
     // get first address to ensure that public key was exported
     await myAddresses.baseExtAddrManager._deriveAddress(0)
+    // in case the wallet is not shelley compatible we consider the
```
refi93, line 214: mixing shelley compatibility and "usedness" seems like a potential great source of confusion :thinking:

do you see some way to reframe/refactor the logic to not have to "bend" the meaning of "used account"?

refi93 (changes requested):
a few clarifying questions/improvement suggestions

app/frontend/wallet/shelley/helpers/addresses.ts
PeterBenc: I agree. Thnx.

app/frontend/wallet/constants.ts
PeterBenc: Its in the PR description as TODO :D, I ll fix those.

app/frontend/wallet/account.ts
PeterBenc: I agree, I don't like this as well, that's why I added the comment, I think we are already discovering accounts only if the wallet is `ShelleyCompatible`. Och I just noticed, I misplaced this comment and it should be somewhere else..I ll fix it. Thnx.

PeterBenc: This probably deserves even more comment, since this logic is pretty crucial and the whole account discovery and xpub exporting depends on it.

app/frontend/wallet/account.ts
```diff
@@ -315,6 +317,8 @@ const Account = ({
   }
 
   function isAccountUsed(): Promise<boolean> {
+    // in case the wallet is not shelley compatible we consider the
```
refi93, line 320: this deserves a TODO comment instructing to decouple shelley compatibility from usedness as the current state is very confusing as discussed before

app/frontend/wallet/constants.ts
refi93: I see, sorry I missed that

app/frontend/wallet/shelley/helpers/addresses.ts
refi93: I would clarify in the comment what we think needs to be done (restrict type of the input to _Address) to not be confused into adding validation here

app/frontend/wallet/account.ts
```diff
@@ -148,7 +148,9 @@ const MyAddresses = ({
     }
   }
 
-  async function areAddressesUsed() {
+  async function areAddressesUsed(): Promise<boolean> {
```
ppershing, line 151: shouldn't this be "isAccountUsed" ?

ppershing: oh, now I see below. This is madness

refi93 (approved):

PR #843, opened 2021-02-19
Refactor tx building 2

PeterBenc:
Changes: 
- refactors tx building so functions like `ShelleyTxInput` doesn't return encodeCbor function but a transactions inputs ready for encoding, also changes the interface so it receives array of inputs instead of a single input to put encoding logic in one place, same for all parts of the transaction
- Adds types for Tx entries, like inputs and outputs
- This allowed me to refactor `prepareTxAux` function in wallet as well as the fee calculation logic

In `shelley-transaction-planner.ts`
- Add types for tx plan and its parts (_Input, _Output....) 
- Refactor error handling when calculating plan, we ll probably change the logic a bit due to multi assets but its more readable and functional already
- Refactor computing tx plan, 
- Introduces a new type TxPlanResult

In `actions.ts`
- Simplifies the logic in `calculateFee`, `calculateDelegationFee`, `withdrawRewards` and `convertNonStakingUtxos` a lot
- Adds txPlan types
- Types necessary fields

Gotcha: 
- I typed only necessary stuff and used `as` a lot since its not really in the scope of this PR to refactor everything, only the part of workflow from "creating the txplan" to "submitting the tx", 
- I left plenty of todos but please comment if you find them important enough for me to implement now, 
- I didn't get to type and refactor the cryptoproviders, but since the tx plan and Tx aux are ready that's the next step
- Some of the tests are failing since the fee calculation logic has changed
- I didn't really test if the fee calculation computes a fee big enough, I will test that and add missing parts in this PR.
- I didn't refactor or ficx anything related to pool registration, currently not working

TODO: 
- address all TODOs left
- refactor and type crypto providers
- refactor and fix pool registration
- test fee calculation and tx submitting

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -34,10 +34,24 @@ import {State, GetStateFn, SetStateFn, getSourceAccountInfo} from './state'
 import ShelleyCryptoProviderFactory from './wallet/shelley/shelley-crypto-provider-factory'
 import {ShelleyWallet} from './wallet/shelley-wallet'
 import {parseUnsignedTx} from './helpers/cliParser/parser'
-import {TxPlan, unsignedPoolTxToTxPlan} from './wallet/shelley/shelley-transaction-planner'
+import {
+  TxPlan,
+  TxPlanResult,
+  TxPlanResultType,
+  unsignedPoolTxToTxPlan,
+} from './wallet/shelley/shelley-transaction-planner'
 import getDonationAddress from './helpers/getDonationAddress'
 import {localStorageVars} from './localStorage'
-import {AccountInfo, Ada, Lovelace, CryptoProviderFeature, _Address, TxType, AuthMethodType} from './types'
+import {
+  AccountInfo,
+  Ada,
+  Lovelace,
+  CryptoProviderFeature,
+  _Address,
+  TxType,
+  TxPlanArgs,
+  AuthMethodType,
+} from './types'
 import {MainTabs} from './constants'
 
 let wallet: ReturnType<typeof ShelleyWallet>
@@ -479,20 +493,25 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
 
   /* SEND ADA */
 
-  const setTransactionSummary = (tab, plan, coins?, donationAmount?) => {
+  const setTransactionSummary = (
+    tab, // TODO: add enum for tabs
+    plan: TxPlan,
+    coins?: Lovelace,
+    donationAmount?: Lovelace
+  ) => {
     setState({
       sendTransactionSummary: {
-        amount: coins || 0,
-        donation: donationAmount || 0,
-        fee: plan.fee != null ? plan.fee : plan.estimatedFee,
-        plan: plan.fee != null ? plan : null,
+        amount: coins || (0 as Lovelace),
+        donation: donationAmount || (0 as Lovelace),
+        fee: plan.fee,
+        plan,
         tab,
         deposit: plan.deposit,
       },
     })
   }
 
-  const resetTransactionSummary = (state) => {
+  const resetTransactionSummary = (state: State) => {
     setState({
       sendTransactionSummary: {
         amount: 0 as Lovelace,
@@ -500,7 +519,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
         donation: 0 as Lovelace,
         plan: null,
         tab: state.sendTransactionSummary.tab,
-        deposit: 0,
+        deposit: 0 as Lovelace,
       },
     })
   }
@@ -564,33 +583,18 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     setErrorState('sendAddressValidationError', sendAddressValidator(state.sendAddress.fieldValue))
     setErrorState(
       'sendAmountValidationError',
-      sendAmountValidator(
-        state.sendAmount.fieldValue,
-        state.sendAmount.coins,
-        getSourceAccountInfo(state).balance
-      )
+      sendAmountValidator(state.sendAmount.fieldValue, state.sendAmount.coins, getSourceAccountInfo(
+        state
+      ).balance as Lovelace)
     )
     setErrorState(
       'donationAmountValidationError',
       donationAmountValidator(
         state.donationAmount.fieldValue,
         state.donationAmount.coins,
-        getSourceAccountInfo(state).balance
+        getSourceAccountInfo(state).balance as Lovelace
       )
     )
-  }
-
-  const resetDelegationWithoutHash = (state) => {
-    const newState = getState()
-    setState({
-      shelleyDelegation: {
-        ...state.shelleyDelegation,
-        selectedPool: {
-          ...newState.shelleyDelegation.selectedPool,
-        },
-        delegationFee: 0,
-      },
-    })
   }
 
   const isSendFormFilledAndValid = (state) =>
@@ -600,27 +604,18 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
     !state.sendAmountValidationError &&
     !state.donationAmountValidationError
 
-  const prepareTxPlan = async (args) => {
+  const prepareTxPlan = async (args: TxPlanArgs): Promise<TxPlanResult> => {
     const state = getState()
     try {
-      const plan: any = await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
-      // FIXME: this cant be any
-      if (plan.error) {
-        stopLoadingAction(state, {})
-        resetDelegationWithoutHash(state)
-        setState({
-          calculatingDelegationFee: false,
-          calculatingFee: false,
-        })
-      }
-      return plan
+      return await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
     } catch (e) {
-      if (e.name !== 'NetworkError' || e.name !== 'ServerError') {
+      // TODO: refactor setErrorState to check all errors if there unexpected
+      if (e.name !== 'NetworkError' && e.name !== 'ServerError') {
         throw e
       }
-
       return {
-        estimatedFee: 0,
+        type: TxPlanResultType.FAILURE,
+        estimatedFee: 0 as Lovelace,
         error: {code: e.name},
       }
     }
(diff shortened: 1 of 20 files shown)


app/frontend/actions.ts
```diff
@@ -479,28 +493,33 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+    coins?: Lovelace,
+    donationAmount?: Lovelace
+  ) => {
     setState({
       sendTransactionSummary: {
-        amount: coins || 0,
-        donation: donationAmount || 0,
-        fee: plan.fee != null ? plan.fee : plan.estimatedFee,
-        plan: plan.fee != null ? plan : null,
+        amount: coins || (0 as Lovelace),
```
ppershing, line 504: prefer ?? instead of ||

app/frontend/actions.ts
```diff
@@ -479,28 +493,33 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
+        donation: donationAmount || (0 as Lovelace),
+        fee: plan.fee,
+        plan,
         tab,
```
ppershing, line 508: wtf is tab doing in transaction summary?

app/frontend/actions.ts
```diff
@@ -34,10 +34,24 @@ import {State, GetStateFn, SetStateFn, getSourceAccountInfo} from './state'
...
+import {
+  TxPlan,
+  TxPlanResult,
+  TxPlanResultType,
+  unsignedPoolTxToTxPlan,
+} from './wallet/shelley/shelley-transaction-planner'
 import getDonationAddress from './helpers/getDonationAddress'
 import {localStorageVars} from './localStorage'
-import {AccountInfo, Ada, Lovelace, CryptoProviderFeature, _Address, TxType, AuthMethodType} from './types'
+import {
```
ppershing, line 45: Tooling: There is an eslint plugin that converts type imports to `import type {TypeA, ...}` format which is quite preferred. We should set it up...

app/frontend/actions.ts
```diff
@@ -564,63 +583,39 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-      if (plan.error) {
-        stopLoadingAction(state, {})
-        resetDelegationWithoutHash(state)
-        setState({
-          calculatingDelegationFee: false,
-          calculatingFee: false,
-        })
-      }
-      return plan
+      return await wallet.getAccount(state.sourceAccountIndex).getTxPlan(args)
```
ppershing, line 610: For the future: I hate that this contains side-effects. In reality, we should split out the functionality into fetching of available inputs and "coin selection" algorithm

app/frontend/actions.ts
PeterBenc: I didnt refactor this, but its rather a bad naming, it should be `type` and `TxType` enum.

app/frontend/actions.ts
```diff
@@ -564,63 +583,39 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
     } catch (e) {
-      if (e.name !== 'NetworkError' || e.name !== 'ServerError') {
+      // TODO: refactor setErrorState to check all errors if there unexpected
+      if (e.name !== 'NetworkError' && e.name !== 'ServerError') {
```
ppershing, line 613: we should at least have Enum for errors

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -1,315 +1,334 @@
...
+  const fee = txFeeFunction(estimateTxSize(inputs, outputs, certificates, withdrawals))
   return fee
 }
 
-function computeRequiredDeposit(certs: Array<Cert>): Lovelace {
-  let deposit = 0
-  for (const {type} of certs) {
-    if (type === 0) deposit += 2000000
+function computeRequiredDeposit(certificates: Array<_Certificate>): Lovelace {
+  const CertificateDeposit: {[key in CertificateType]: number} = {
```
ppershing, line 135: shouldn't this be in network constants?

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -1,315 +1,334 @@
...
+    inputs.push(utxo)
+    try {
+      const plan = computeTxPlan(inputs, outputs, change, certificates, withdrawals)
+      validateTxPlan(plan)
+      return {
+        type: TxPlanResultType.SUCCESS,
+        txPlan: plan,
+      }
+    } catch (e) {
+      if (inputs.length === utxos.length) {
```
ppershing, line 319: wtf are we doing here?
This at minimum should be just after the cycle though a deeper question is whether compute/validate tx plan needs to throw ...

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -227,15 +154,66 @@ function ShelleySignedTransactionStructured(txAux, witnesses, meta) {
...
+      poolParams.poolOwners.map((ownerObj) => {
+        if (ownerObj.stakingKeyHashHex) {
+          return Buffer.from(ownerObj.stakingKeyHashHex, 'hex')
+        }
+        // else is a path owner and has pubKeyHex
+        return Buffer.from(ownerObj.pubKeyHex, 'hex')
+      }),
+      poolParams.relays.map((relay) => {
+        switch (relay.type) {
+          case 0:
```
ppershing, line 184: do we really want hard-coded numbers?

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: There is a comment above, that we should refactor this, I just made it so it works for now.. :D. Regarding the throwing in computing tx plan, we need information about what caused the plan to not be created. So I intentionally made it to throw errors.

app/frontend/wallet/shelley/shelley-transaction.ts
PeterBenc: As mentioned in the PR description, I wasn't touching the pool registration-related functionality. I just move this into a function as it was. I ll refactor the pool registration.

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: I suppose we can drop the the throwing in the plan computation and validate it in the `validateTxPlan`. in that case tho, we would do a lot of stuff twice, (like computing the size of tx in bytes...) but I guess it would separate the concert of tx plan validation... :thinking:

app/frontend/actions.ts
```diff
@@ -564,63 +583,39 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-      )
+      sendAmountValidator(state.sendAmount.fieldValue, state.sendAmount.coins, getSourceAccountInfo(
+        state
+      ).balance as Lovelace)
     )
     setErrorState(
       'donationAmountValidationError',
       donationAmountValidator(
         state.donationAmount.fieldValue,
         state.donationAmount.coins,
```
refi93, line 594: couldn't this also be `as Lovelace`?

app/tests/src/index.js
```diff
@@ -1,24 +1,24 @@
...
-  describe('Blockchain Explorer', () => {
-    require('./blockchain-explorer')
-  })
-  describe('Cardano Wallet', () => {
-    require('./cardano-wallet')
-  })
-  describe('Import/Export Wallet as JSON', () => {
-    require('./keypass-json')
-  })
+  // describe('CBOR', () => {
```
refi93, line 4: don't forget to uncomment these tests

app/frontend/wallet/shelley/types.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export type TxStakingKeyDeregistrationCert = [
+  TxCertificateKey.STAKING_KEY_DEREGISTRATION,
+  [number, Buffer]
+]
+
+export type TxDelegationCert = [TxCertificateKey.DELEGATION, [number, Buffer], Buffer]
+
+// prettier-ignore
+export type TxSingleHostIPRelay = [
```
refi93, line 88: worth adding at least comments stating what those numbers and buffers are supposed to be so the code is at least readable until we actually get to refactor it

app/frontend/actions.ts
```diff
@@ -637,31 +632,42 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-    setErrorState('sendAmountValidationError', validationError)
-    if (!validationError) {
-      setTransactionSummary('send', plan, coins, donationAmount)
+      donationAmount,
+      txType: TxType.SEND_ADA,
+    })
+    const balance = getSourceAccountInfo(state).balance as Lovelace
+
+    if (txPlanResult.type === TxPlanResultType.SUCCESS) {
+      const newState = getState() // if the values changed meanwhile
```
refi93, line 646: I'm not that confident this check actually works - if I'm not mistaken, our state is currently mutable so the "old" state's nested keys will get modified regardless

refi93: ok, no, I checked and the state storage implementation copies the top-level key-value pairs into a new object each time they are modified: https://github.com/vacuumlabs/adalite/blob/eebae33b7bbb339d96d02c1c320f9547778134bf/app/frontend/libs/unistore/index.js#L34

PeterBenc: :D I don't think it works either.. I ll test it and remove it. :+1:

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -30,35 +52,57 @@ function ShelleyTxAux(inputs, outputs, fee, ttl, certs?, withdrawals?) {
...
     withdrawals,
     encodeCBOR,
   }
 }
 
-function ShelleyFee(fee) {
-  function encodeCBOR(encoder) {
-    return encoder.pushAny(fee)
-  }
+function ShelleyTxInputs(inputs: _Input[]): TxInput[] {
```
refi93, line 61: I would not pollute the general namespace with types/functions meant for translating our higher-level objects like tx inputs/outputs into their low-level versions ready to be serialized and don't serve a purpose outside of tx serialization which is one of the lowest-level concerns of the wallet

I would name this function something like `cborizeTxShelleyInputs` and call the returned type `cborizedTxInput` to make it obvious it's kind of a low-level type. Similarly for other types being "cborized" here

PR #845, opened 2021-02-21
Fix "Cannot read property 'url'" for pools missing from validStakepools mapping

refi93:
Motivation: A user reported running into a `Cannot read property 'url' of undefined` error due to given pool not being among the "valid stake pools"

Changes:
* fix the logic fetching pool info to gracefully fall back to `UNKNOWN_POOL_NAME` if pool URL is missing

Testing:
* I was not able to reproduce the exact error, but I tested the wallet loads fine and given logic is triggered by logging from the respective function

### DIFF
--- a/app/frontend/wallet/blockchain-explorer.ts
+++ b/app/frontend/wallet/blockchain-explorer.ts
@@ -33,6 +33,7 @@ import {
   PoolRecommendationResponse,
   StakingInfoResponse,
   BestSlotResponse,
+  StakePoolInfo,
 } from './explorer-types'
 
 const cacheResults = (maxAge: number, cache_obj: Object = {}) => <T extends Function>(fn: T): T => {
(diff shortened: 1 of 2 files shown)


app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -360,17 +361,28 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     validStakepools: ValidStakePoolsMapping,
     epochsToRewardDistribution: number
   ): Promise<NextRewardDetailsFormatted> {
+    const getPool = async (
+      poolHash: string
+    ): Promise<StakePoolInfo | HostedPoolMetadata | string> => {
```
refi93, line 366: this type ambiguity seems very confusing, a proper fix would probably involve significantly more changes/refactoring, so at least I explicitly put here the types to "document" the current behavior

PeterBenc (approved):
Thnx for fixing this, we should probably fix the eslint error before merging.. :)

refi93:
@PeterBenc - yes, dumb mistake, I forgot to await the result of the newly introduced async helper function :facepalm:

PR #848, opened 2021-02-22
Type blockchain explorer endpoints

xdzurman:
- Separates backend types and frontend types regarding blockchain explorer
- Assigns types to each blockchain explorer endpoint
- Removes mutable logic from getTxsHistory

### DIFF
--- a/app/frontend/wallet/backend-types.ts
+++ b/app/frontend/wallet/backend-types.ts
@@ -1,12 +1,4 @@
-export type HostedPoolMetadata = {
-  name: string
-  description: string
-  ticker: string
-  homepage: string
-  extended?: string
-}
-
-export type StakePoolInfo = {
+type StakePoolInfo = {
   pledge: string
   margin: number
   fixedCost: string
@@ -15,7 +7,7 @@ export type StakePoolInfo = {
   homepage: string
 }
 
-export type StakePoolInfoExtended = StakePoolInfo & {
+type StakePoolInfoExtended = StakePoolInfo & {
   poolHash: string
   liveStake: string
   roa: string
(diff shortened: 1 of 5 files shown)


app/frontend/wallet/backend-types.ts
```diff
@@ -105,8 +86,15 @@ export type PoolRecommendationResponse = StakePoolInfoExtended & {
   isInRecommendedPoolSet: boolean
 }
 
+export declare type Token = {
+  policyId: string
+  assetName: string
+  quantity: string // possibly huge amounts
+}
+
 export type CoinObject = {
```
refi93, line 95: I would name this AmountEntry as agreed in adalite-backend

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -424,10 +435,11 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     }))
   }
 
-  async function getStakingInfo(stakingKeyHashHex: HexString) {
+  async function getStakingInfo(stakingKeyHashHex: HexString): Promise<StakingInfoResponse> {
     const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/account/info/${stakingKeyHashHex}`
-    const response: StakingInfoResponse = await request(url)
+    const response = await request(url)
     // if we fail to recieve poolMeta from backend
+    // TODO: IMHO we shouldn't freestyle append poolInfo here, it breaks types easily
```
refi93, line 442: agree

app/frontend/wallet/backend-types.ts
```diff
@@ -105,8 +86,15 @@ export type PoolRecommendationResponse = StakePoolInfoExtended & {
   isInRecommendedPoolSet: boolean
 }
 
+export declare type Token = {
+  policyId: string
+  assetName: string
+  quantity: string // possibly huge amounts
```
refi93, line 92: we should introduce bigint for that - native is probably a bit too soon as Safari switched only recently, so I'd use some library for that

PR #854, opened 2021-02-25
[Mary] Add asset data provider

DavidTranDucVL:
- Added AssetDataProvider to state, currently it updates from UI when selecting asset from dropdown.
- Refactored StakepoolDataProvider


app/frontend/actions.ts
```diff
@@ -229,7 +228,7 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
         shouldShowGenerateMnemonicDialog: false,
         shouldShowAddressVerification: usingHwWallet,
         // send form
-        sendAmount: {fieldValue: '', coins: 0 as Lovelace},
+        sendAmount: {isLovelace: true, fieldValue: '', coins: 0 as Lovelace},
```
refi93, line 231: this looks shady - I'd keep a separate attribute of the form aside from sendAmount to track what kind of asset is being sent and perhaps generalize the type of the `coins` field from Lovelace to CoinAmount

refi93: or keep it inside, in the end it probably doesn't matter that much, especially, if it's convenient but not have just a flag whether it's ADA or not, but have the assetType explicitly - I'd imagine a union type like:
```
AssetType = {
    type: ADA
} | {
    type: NATIVE_ASSET;
    policyIdHex: string,
    assetNameHex: string 
}
```

app/frontend/types.ts
```diff
@@ -256,3 +251,20 @@ export type Balance = {
   coins: Lovelace
   tokens: Token[]
 }
+
+export const enum AssetType {
+  ADA,
+  TOKEN,
```
refi93, line 257: wouldn't it make sense to have type to describe the exact assetName and policyId of the given token? And have this enum instead as "AssetFamily" or "AssetKind" . The AssetType type then could be something like:

```
```
AssetType = {
    family: AssetFamily.ADA
} | {
    family: AssetFamily.TOKEN;
    policyIdHex: string,
    assetNameHex: string 
}
```
```

app/frontend/actions.ts
DavidTranDucVL: I had similar idea initially, look at https://github.com/vacuumlabs/adalite/pull/854/commits/454301bba27a0ee96f98ae2bce784db2f1df0532#diff-3fb8e9893dfb9f69a160bd6ba11cd0ce4827c717c0285d2a928c15c27ff8f238 but then we steered away from this idea. I'm fine with whichever. @PeterBenc what do you think.

DavidTranDucVL:
This PR is outdated.

PR #858, opened 2021-03-01
Multiasset support

refi93:


app/frontend/wallet/backend-types.ts
```diff
@@ -105,8 +88,15 @@ export type PoolRecommendationResponse = StakePoolInfoExtended & {
   isInRecommendedPoolSet: boolean
 }
 
+export type TokenObject = {
+  policyId: string
+  assetName: string
+  quantity: string // possibly huge amounts
```
refi93, line 94: I would add a TODO and backlog issue to transition to BigInt (probably not the native one as Safari added it only recently) - anybody can easily mint/have tokens with huge amounts well beyond JS max safe int

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -1,315 +1,384 @@
...
+  const change = {
+    ...possibleChange,
+    tokens: tokenDifference,
+  }
+
   const feeWithChange = computeRequiredTxFee(
     inputs,
-    [...outputs, possibleChange],
-    certs,
+    [...outputs, change],
```
refi93: we should watch out for big amounts of tokens overflowing max uint64 - not a showstopper for the release but should be a part of the overall migration to bigint. Similar issue in cardano-wallet for context: https://github.com/input-output-hk/cardano-wallet/issues/2532

app/frontend/wallet/backend-types.ts
```diff
@@ -105,8 +88,15 @@ export type PoolRecommendationResponse = StakePoolInfoExtended & {
   isInRecommendedPoolSet: boolean
 }
 
+export type TokenObject = {
```
refi93, line 91: the `Object` suffix is IMO too cryptic, ideally we should have a single type using solely BigInts, but meanwhile fr the sake of clarity I'd use an explicit name like `TokenWithStringifiedQuantity`

app/frontend/wallet/backend-types.ts
PeterBenc: I agree, I intentionally dropped the support for bigints to not support them in some places, and don't support them in other places, we should handle the migration to bigint in separate issue dedicate to that so nothing is misses. I ll create an issue for it.

app/frontend/wallet/backend-types.ts
PeterBenc: I plan to refactor the types before the release, as we talked, add the cborized prefix and remove the _, I ll include this as well :+1:

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -40,208 +46,364 @@ interface Props {
...
+
+  const adaAsset: DropdownAssetItem = {
+    type: AssetFamily.ADA,
+    policyId: null,
+    assetName: 'ADA',
+    assetNameHex: null,
+    quantity: balance,
+    star: true,
+  }
+  const dropdownAssetItems: Array<DropdownAssetItem> = useMemo(
```
refi93, line 145: will the data cached by `useMemo()` be updated after the user sends/receives tokens?

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -40,208 +46,364 @@ interface Props {
...
+        wrapperClassName="no-margin"
+        defaultItem={selectedAsset}
+        displaySelectedItem={({assetName}: DropdownAssetItem) => assetName}
+        displaySelectedItemClassName="input dropdown"
+        items={dropdownAssetItems}
+        displayItem={showDropdownAssetItem}
+        onSelect={handleDropdownOnSelect}
+        showSearch={dropdownAssetItems.length >= 4}
+        searchPredicate={searchPredicate}
+        searchPlaceholder={`Search from ${dropdownAssetItems.length} assets by name or hash`}
```
refi93, line 277:
```suggestion
        searchPlaceholder={`Search from ${dropdownAssetItems.length} assets by name or policy ID`}
```

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -40,208 +46,364 @@ interface Props {
...
+        htmlFor={`${isModal ? 'account' : ''}send-amount`}
+      >
+        Amount
+      </label>
+      <div className="input-wrapper">
+        <input
+          className="input send-amount"
+          id={`${isModal ? 'account' : ''}send-amount`}
+          name={`${isModal ? 'account' : ''}send-amount`}
+          placeholder="0.000000"
```
refi93, line 297: this should be "0" if it's a token

app/frontend/helpers/cliParser/parser.tsx
```diff
@@ -41,11 +45,51 @@ const parseTxInputs = (txInputs: any[]): _Input[] => {
...
 
+const parseAssets = (assets: any): _Asset[] => {
+  if (!(assets instanceof Map)) {
+    throw Error(Errors.TxAssetParseError)
+  }
+  return Array.from(assets).map(([assetName, amount]) => {
+    if (!Buffer.isBuffer(assetName)) {
+      throw Error(Errors.AssetNameParseError)
+    }
+    // TODO: is lovelace is not the best name since its not a lovelace
```
refi93, line 56: I guess this comment is outdated

app/frontend/helpers/cliParser/parser.tsx
```diff
@@ -172,21 +211,58 @@ const parseTxCerts = (txCertificates: any[]): _Certificate[] => {
...
+
+const parseMetaDataHash = (metaDataHash: any): Buffer | undefined => {
+  if (metaDataHash && !Buffer.isBuffer(metaDataHash)) {
+    throw Error(Errors.MetaDataHashParseError)
+  }
+  return metaDataHash
 }
 
 const parseUnsignedTx = ([txBody, meta]: _UnsignedTxDecoded): _UnsignedTxParsed => {
+  if (txBody.get(TxBodyKeys.MINT)) {
```
refi93, line 252: wouldn't it be a more resilient/simpler approach to serialize the tx again and compare if it matches byte-by-byte with the originally supplied tx?

app/frontend/helpers/cliParser/types.ts
```diff
@@ -20,15 +22,23 @@ export const enum TxCertificateKeys {
   STAKEPOOL_REGISTRATION = 3,
 }
 
+export type Lovelace = BigInt
```
refi93, line 25: If we are going with native BigInt, we should add a check for this in our initial "feature check" before the page gets loaded: https://github.com/vacuumlabs/adalite/blob/d7c1656937f36bc3331d06cdc4f1efae8264eecb/app/public/js/init.js#L2 as browsers such as Safari adopted it only recenly: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt - for this advanced feature it's currently not a big deal, so this is rather a note for the future when we update the wallet logic to handle big ints.

Probably worth considering though if we wouldn't be better off just using some library to not leave out users of older/non-mainstream browsers, but maybe it's too low of a percentage to bother as the majority of Safari users appears to have the 14 version already: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

app/frontend/wallet/account.ts
```diff
@@ -225,91 +222,68 @@ const Account = ({
...
-      donationAmount,
-      changeAddress,
-      stakingAddress,
-      poolHash,
-      !stakingKeyRegistered,
-      rewards
-    )
-    return plan
+    // TODO: ideally we want to do nonStaking, with token, only ada, the rest
+    if (txType === TxType.SEND_ADA || txType === TxType.CONVERT_LEGACY) {
```
refi93, line 268: what's the point in having this condition? Is it addressing some edge case that would otherwise result in a suboptimal tx plan? Probably worth elaborating with a comment if it isn't actually redundant

app/frontend/wallet/account.ts
```diff
@@ -225,91 +222,68 @@ const Account = ({
...
+      return [
+        ...shuffleArray(nonStakingUtxos, randomGenerator),
+        ...shuffleArray(baseAddressUtxos, randomGenerator),
+      ]
+    }
+    const adaUTxOs = baseAddressUtxos.filter(({tokens}) => tokens.length === 0)
```
refi93, line 274: `adaOnlyUtxos` would be a more accurate name

app/frontend/components/pages/sendAda/sendAdaPage.tsx
refi93: I tested it manually by sending a new token and appears that yes :thinking:

refi93: ahhh, I see, it's a preact hooks native helper and `[adaAsset, tokenBalance]` below specifies that the cache is invalidated any time any of those entries changes: https://preactjs.com/guide/v10/hooks/#usememo

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: Fixed

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: Fixed

PR #864, opened 2021-03-03
Add cardano scan links to multi-asset tokens

DavidTranDucVL:
This also fixes bug when user clicks on dropdown button and it reopens instead of toggle.


PeterBenc (approved):
LGTM

PR #866, opened 2021-03-04
Move and redesign "My Addresses" to "Receive" tab

DavidTranDucVL:
Adds receive tab with my addresses
Removes my addresses from send tab


xdzurman:
image
naming of the tabs seems kind of inconsistent right now as of the -ing suffix, although if discussed with Michal then I'm fine with it

xdzurman:
image
I think we should make the "expanded" logic more user friendly - perhaps an arrow indicating that others are expandable as well, or highlight the background slightly? What do you think?

xdzurman:
Can we make the whole span clickable? I don't know if it's just my sloppy clicking skills, but I hit the outer span in 25% of the cases and the address didn't expand even though the pointer makes it intuitive that it should work.

https://user-images.githubusercontent.com/22454337/110321546-1e3ccd80-8012-11eb-9a64-54795ead3fee.mp4

DavidTranDucVL:
> I think we should make the "expanded" logic more user friendly - perhaps an arrow indicating that others are expandable as well, or highlight the background slightly? What do you think?

I was thinking about this as well, but couldn't settle with a solution. If we add arrow like in delegate stake card, it might be weird that clicking on other items in list automatically contracts previous item. Also so far we don't contract the item you have already opened if you click on it again - so this opens up discussion on to if we would want that.

xdzurman:
> > I think we should make the "expanded" logic more user friendly - perhaps an arrow indicating that others are expandable as well, or highlight the background slightly? What do you think?
> 
> I was thinking about this as well, but couldn't settle with a solution. If we add arrow like in delegate stake card, it might be weird that clicking on other items in list automatically contracts previous item. Also so far we don't contract the item you have already opened if you click on it again - so this opens up discussion on to if we would want that.

I kinda fancy the highlight solution in the Account tab: 
image

app/frontend/components/pages/receiveAda/addressItem.tsx
```diff
@@ -1,125 +1,127 @@
...
+      {name}
+    </a>
+  ) : (
+    <Fragment>
+      View on{' '}
+      <a className="address-link" href={url} target="_blank" rel="noopener">
+        {name}
+      </a>
+    </Fragment>
+  )
```
xdzurman, line 25: Although small in size, I'd move it to a separate file. I see we use also "ViewOnCardanoScan", so it would perhaps make sense to merge it and use just that

app/frontend/components/pages/receiveAda/addressItem.tsx
```diff
@@ -1,125 +1,127 @@
...
-  constructor(props) {
-    super(props)
-    this.updateDimensions = this.updateDimensions.bind(this)
-    this.openAddressDetail = this.openAddressDetail.bind(this)
-    this.state = {onSmallDevice: false}
-  }
+const AddressItem = ({address, bip32path, isExpanded, expand}: Props): h.JSX.Element => {
+  const explorerLinksInline = (
+    <div>
+      {ADALITE_CONFIG.ADALITE_CARDANO_VERSION === 'byron' && (
```
xdzurman, line 37: I won't pester you with refactoring "ADALITE_CARDANO_VERSION === 'byron'" into an enum right now, I'm adding this into the backlog though

app/frontend/components/pages/receiveAda/addressItem.tsx
```diff
@@ -1,125 +1,127 @@
...
+            <CopyOnClick value={address} elementClass="address-link copy">
+              <a className="copy-text">Copy Address</a>
+            </CopyOnClick>
+          </div>
+          <div className="verification">
+            <AddressVerification address={address} />
           </div>
-          <a
-            className="address-link more"
-            onClick={() => this.openAddressDetail({address, bip32path}, this.state.onSmallDevice)}
```
xdzurman, line 112: openAddressDetail is now completely unused

xdzurman: also most likely the whole AddressDetailDialog class and closeAddressDetail

DavidTranDucVL: I think the whole pop verify address modal logic is now unused, not sure if want to keep it or delete it all.

DavidTranDucVL: Removed dead code.

app/frontend/components/pages/receiveAda/addressItem.tsx
DavidTranDucVL: I moved it to separate file, but haven't merged it. I think it complicate things more then make it clearer.

app/public/css/styles.css
```diff
@@ -2521,58 +2522,71 @@ li.main-tab input + label.selected:after {
...
+  display: flex;
+  cursor: pointer;
   color: var(--color-grey);
   word-break: break-all;
-  margin-bottom: 16px;
+}
+
+.address.expanded {
+  background-color: var(--color-selected);
+}
```
PeterBenc, line 2555:
```suggestion
address.expanded {
    border-top: none;
    background-color: var(--color-selected);
    box-shadow: 0px 8px 16px 0px rgba(0, 0, 0, 0.08);
}
```
Just a quick suggestion but I would somehow utilize the box-shadowing to be consistent with other tabs.
image

PeterBenc:
> image
> naming of the tabs seems kind of inconsistent right now as of the -ing suffix, although if discussed with Michal then I'm fine with it

In that case I would make the set the min width of the tabs, I don't really like that now the "Send" tab is so skinny. They dont need to be the exact size but at least seem similar.

app/public/css/styles.css
PeterBenc: But tbh I dont really like that the selected address and the one above are not separated by a line so keeping the `border-top` might be better.

And also this might something to discuss with the designers, especially since the accounts tab has a somewhat similar structure but looks totally different. :smile:

PeterBenc: Also in the picture the `min-width` for the tabs is set.

PeterBenc: (Using this thread for more comments since the "global" comments are painful to reply to :smile: )
If possible, I would utilize the accordion behavior we use on `DelegatePage`. Imho that would look really nice.

DavidTranDucVL: I updated whole behaviour, check it out.

app/public/css/styles.css
```diff
@@ -2521,58 +2524,93 @@ li.main-tab input + label.selected:after {
...
+  max-height: 208px;
+  overflow: hidden;
+  transition: max-height 0.5s ease-out;
+}
+
+.address .hide {
+  max-height: 0px;
+  overflow: hidden;
+  transition: max-height 0.5s ease-out;
+}
```
xdzurman, line 2596: I played around with the transitions a little bit, 0.5s seems a little laggy and slow. I tried 0.25s, also `transition: background-color 0.25s ease-out 0.25s, box-shadow 0.25s ease-out 0.25s;`  and it felt better for me, what do you think?

DavidTranDucVL: Sounds good, changed it to 0.25s

xdzurman: `transition: max-height 0.5s ease-out;` in 0-767px.css needs to be changed as well, now it behaves strangely on mobile

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -85,6 +86,7 @@ const SendingPage = ({
           </div>
           <div className="dashboard-column">
             <SendAdaPage />
+            <ReceiveReddirect />
```
xdzurman, line 89: typo

app/public/css/styles.css
DavidTranDucVL: Fixed

xdzurman (approved):
LGTM, just squash into fewer commits

PR #868, opened 2021-03-05
Add change coins smaller than min utxo value to fee

PeterBenc:
Previously we were adding the change amount that was lower than the minimal utxo amount to the firts output. This is unwanted behaviour in case a user wants to send an exact amount.

Changes: 
- we add the change amount to fee


DavidTranDucVL (approved):
LGTM

PR #871, opened 2021-03-06
Add wanted account modal

xdzurman:
Adds wanted account modal for specific addresses

This is supposed to be a temporary solution, not the most modular or beautiful addition

image


PeterBenc (approved):
Thnx

PR #876, opened 2021-03-09
Mary fixes

PeterBenc:


app/frontend/components/pages/accounts/accountDropdown.tsx
```diff
@@ -27,7 +20,7 @@ const AccountDropdown = ({
         onBlur={() => hideAccountDropdown(true)}
         onClick={() => toggleAccountDropdown()}
       >
-        {shouldNumberAccountsFromOne ? `Account #${accountIndex + 1}` : `Account ${accountIndex}`}
+        {`Account ${printAccountIndex(accountIndex)}`}
```
refi93, line 23: for the sake of consistency, I'd name the function `formatAccountIndex`

app/frontend/wallet/helpers/tokenFormater.ts
```diff
@@ -7,13 +7,12 @@ import {TokenObject} from '../backend-types'
 export const arraySum = (numbers: Array<number>): number =>
   numbers.reduce((acc: number, val) => acc + val, 0)
 
-export const formatToken = (token: TokenObject, multiplier: number = 1): Token => ({
-  policyId: token.policyId,
-  assetName: token.assetName,
-  quantity: multiplier * parseInt(token.quantity, 10),
+export const formatToken = (token: TokenObject): Token => ({
```
refi93, line 10: this is really a misleading name - until I read the code I though it formats the token into its string representation. Perhaps `parseToken` would be more accurate

app/frontend/wallet/helpers/tokenFormater.ts
```diff
@@ -25,7 +24,7 @@ const aggregateTokensForPolicy = (policyGroup: Array<Token>, policyId: string) =
     )
     .value()
 
-export const aggregateTokens = (tokens: Array<Array<Token>>): Array<Token> =>
+export const aggregateTokens = (tokens: Token[][]): Token[] =>
```
refi93, line 27: `aggregateTokensByPolicy` perhaps would be a better name, especially given that the output type is so opaque

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -151,24 +152,43 @@ function computeRequiredDeposit(certificates: Array<TxCertificate>): Lovelace {
...
+
+  const numAssets = distinctAssets.length
+  const numPIDs = distinct(tokens.map(({policyId}) => policyId)).length
+
+  const sumAssetNameLengths = distinctAssets.reduce(
+    (acc, assetName) => acc + Math.max(Buffer.from(assetName, 'hex').byteLength, 1),
+    0
+  )
+
+  const pidSize = 28
```
refi93, line 180: better be explicit and name it `policyIdSize`, a reader without context may get confused and think first of some "process id" or other totally misleading/unrelated concept

app/frontend/wallet/helpers/tokenFormater.ts
refi93: ah I see - it actually just takes a list of list of tokens -> I'd create an alias type "TokenBundle" equivalent to "Token[]" to make it clear, and name the function `aggregateTokenBundles` which would imho make it clearer to the reader why nested lists of tokens are being used (actually concealing that technicality from them)

app/frontend/wallet/helpers/tokenFormater.ts
```diff
@@ -39,3 +38,8 @@ export const groupTokensByPolicyId = (tokens: Token[]): {[policyId: string]: Tok
     .groupBy(({policyId}) => policyId)
     .value()
 }
+
+export const getTokenDifference = (tokens1: Token[], tokens2: Token[]): Token[] => {
```
refi93, line 42: related to the suggestion above, I'd name this `getTokenBundlesDifference`

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: Sure, np, but I followed the specification linked above and wanted to be consistent with it as much as possible. A possible reader without context would need to check that specification anyway. Either way, I ll rename it to `policyIdSize`.

app/frontend/wallet/helpers/tokenFormater.ts
PeterBenc: I agree :+1:

app/frontend/components/pages/accounts/accountDropdown.tsx
PeterBenc: Sure, np, but with what exactly you mean it should be consistent with? We use the `printAda` helper in a very similar way so my initial idea was to be consistent with it.

app/frontend/wallet/helpers/tokenFormater.ts
PeterBenc: Good idea :+1:

app/frontend/components/pages/accounts/accountDropdown.tsx
refi93: I think I was referring to `formatToken` which appeared in this PR as well

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -126,24 +126,22 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     for (const [address, amount] of tx.ctbInputs || []) {
       if (addresses.includes(address)) {
         effect -= +amount.getCoin
-        const formatedInputTokens = amount.getTokens.map((token) => formatToken(token, -1))
+        const formatedInputTokens = amount.getTokens.map(formatToken)
```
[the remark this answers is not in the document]
PeterBenc: Fixed.

app/frontend/components/pages/accounts/accountDropdown.tsx
PeterBenc: You mean the one you suggested to be renamed to `parseToken` ? :D Anyway, I renamed it to format token. :+1:

app/frontend/wallet/helpers/tokenFormater.ts
PeterBenc: Fixed.

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: Anyway, I renamed it to `policyIdSize` and left comment stating that it represents `pid` from the specification.

app/frontend/wallet/helpers/tokenFormater.ts
PeterBenc: Fixed.

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -119,30 +119,31 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
+    const outputTokenBundle: TokenBundle[] = []
+    const inputTokenBundle: TokenBundle[] = []
 
     let effect = 0 //effect on wallet balance accumulated
     for (const [address, amount] of tx.ctbInputs || []) {
       if (addresses.includes(address)) {
         effect -= +amount.getCoin
-        const formatedInputTokens = amount.getTokens.map((token) => formatToken(token, -1))
-        inputTokens.push(formatedInputTokens)
+        const formatedInputTokenBundle = amount.getTokens.map((token) => parseToken(token))
```
refi93, line 129: I guess this should now be `parsedInputTokenBundle` after formatToken has been renamed

app/frontend/components/pages/accounts/accountDropdown.tsx
refi93: I see :D I think got confused by that formatToken token function and forgot to update this comment

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -119,30 +119,31 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
+        inputTokenBundle.push(formatedInputTokenBundle)
       }
     }
     for (const [address, amount] of tx.ctbOutputs || []) {
       if (addresses.includes(address)) {
         effect += +amount.getCoin
-        const formatedOutputTokens = amount.getTokens.map((token) => formatToken(token, 1))
-        outputTokens.push(formatedOutputTokens)
+        const formatedOutputTokenBundle = amount.getTokens.map((token) => parseToken(token))
```
refi93, line 136: same as above

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -149,24 +150,45 @@ function computeRequiredDeposit(certificates: Array<TxCertificate>): Lovelace {
...
-    address,
-    coins,
-    tokens,
+  // NOTE: should be 2, but a bug in Haskell set this to 0
+  const coinSize = 0
+  const txOutLenNoVal = 14
+  const txInLen = 7
+  const utxoEntrySizeWithoutVal = 6 + txOutLenNoVal + txInLen // 27
+
+  // NOTE: should be 29 but a bug in Haskell set this to 27
```
refi93, line 165: would be nice to reference the code/issue as this seems pretty obscure and hard to track later on

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -149,24 +150,45 @@ function computeRequiredDeposit(certificates: Array<TxCertificate>): Lovelace {
...
+  const adaOnlyUtxoSize = utxoEntrySizeWithoutVal + coinSize
+
+  const distinctAssets = distinct(tokenBundle.map(({assetName}) => assetName))
+
+  const numAssets = distinctAssets.length
+  // number of unique policyIds
```
refi93, line 171: this comment seems avoidable if we just rename the variable to numPolicyIds and from the assignment it's pretty obvious those are unique

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -117,9 +117,9 @@ function cborizeTxInputs(inputs: TxInput[]): CborizedTxInput[] {
   return txInputs
 }
 
-function cborizeTxOutputTokens(tokens: Token[]): CborizedTxTokens {
+function cborizeTxOutputTokens(tokenBundle: TokenBundle): CborizedTxTokens {
```
refi93, line 120: this begs for renaming to `cborizeTxOutputTokenBundle` and the type would be `CborizedTxTokenBundle` then

app/frontend/helpers/formatAccountIndex.ts
```diff
@@ -0,0 +1 @@
+export const formatAccountIndex = (accountIndex: number) => `#${accountIndex + 1}`
```
refi93, line 1: it seems like an overkill to me to create a file just to declare this helper - I'd consider adding it to an index.ts file together with other such trivial helpers

app/frontend/helpers/distinct.ts
```diff
@@ -1 +1 @@
-export default (array) => Array.from(new Set(array))
+export default <T>(array: T[]) => Array.from(new Set(array)) as T[]
```
refi93, line 1: such default export hampers searchability by function name which is a bit annoying at least

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: OK, but I still dont really understand why we want to change this implementation in anyway since it specifically points to the specification, and anyone wondering can just open the link and understand.

app/frontend/wallet/shelley/shelley-transaction.ts
PeterBenc: thnx :+1:

app/frontend/helpers/formatAccountIndex.ts
PeterBenc: Do you min addressing this in another PR? This would also go for other helpers which is a bit out of scope for this PR.

app/frontend/wallet/blockchain-explorer.ts
PeterBenc: Thnx fixed.

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: It would be nice to reference the issue but I think its sufficient enough for now. Who knows if there even is an issue for this, I think the time spent looking for the cause in cardano node respo or looking through all the issues in that repository is not worth it imo.

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: It makes more sense to me to follow the spec to the letter rather than rename stuff just so it makes a bit more sense in the scope of the function, but it doesn't match whats in the spec.

app/frontend/helpers/distinct.ts
PeterBenc: Fixed.

PR #879, opened 2021-03-10
Create multi assets page

rizip1:
* Display "multi-assets" data from "sendAdaPage" in a standalone card & make it scrollable
* Adjust for all resolutions
* Create `FormattedAssetItem` for shared Asset formatting, with the freedom to use in any layout

// desktop
a1

// mobile
a2

### DIFF
--- a/app/frontend/components/pages/dashboard/dashboardPage.tsx
+++ b/app/frontend/components/pages/dashboard/dashboardPage.tsx
@@ -5,6 +5,7 @@ import Balance from '../../common/balance'
 import TransactionHistory from '../txHistory/transactionHistory'
 import ExportCard from '../exportWallet/exportCard'
 import SendAdaPage from '../sendAda/sendAdaPage'
+import MultiAssetsPage from '../sendAda/multiAssetsPage'
 import MyAddresses from '../receiveAda/myAddresses'
 import DelegatePage from '../delegations/delegatePage'
 import CurrentDelegationPage from '../delegations/currentDelegationPage'
@@ -86,8 +87,9 @@ const SendingPage = ({
             <Balance />
             <TransactionHistory />
           </div>
-          <div className="dashboard-column">
+          <div className="dashboard-column shrinkable">
             <SendAdaPage />
+            <MultiAssetsPage />
             <ReceiveRedirect />
             {shouldShowExportOption && <ExportCard />}
           </div>
@@ -161,11 +163,20 @@ const AccountsPage = ({screenType}: {screenType: ScreenType}) => {
   )
 }
 
+const MobileSendAdaPage = () => (
+  <Fragment>
+    <SendAdaPage />
+    <div className="mobile-multi-assets-page-wrapper">
+      <MultiAssetsPage />
+    </div>
+  </Fragment>
+)
+
 const SubPages: {[key in SubTabs]: any} = {
   [SubTabs.DELEGATE_ADA]: <DelegatePage />,
   [SubTabs.CURRENT_DELEGATION]: <CurrentDelegationPage />,
   [SubTabs.STAKING_HISTORY]: <StakingHistoryPage />,
-  [SubTabs.SEND_ADA]: <SendAdaPage />,
+  [SubTabs.SEND_ADA]: <MobileSendAdaPage />,
   [SubTabs.TRANSACTIONS]: <TransactionHistory />,
   [SubTabs.ADDRESSES]: <MyAddresses />,
   [SubTabs.KEYS]: <Keys />,
--- a/app/frontend/components/pages/sendAda/multiAssetsPage.tsx
+++ b/app/frontend/components/pages/sendAda/multiAssetsPage.tsx
@@ -1,0 +1,73 @@
+import {h} from 'preact'
+import {useSelector} from '../../../helpers/connect'
+import {getSourceAccountInfo, State} from '../../../state'
+import {AssetFamily, Token} from '../../../types'
+import {assetNameHex2Readable} from '../../../wallet/shelley/helpers/addresses'
+import {FormattedAssetItem} from '../../common/asset'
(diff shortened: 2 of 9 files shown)


app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -160,11 +162,20 @@ const AccountsPage = ({screenType}: {screenType: ScreenType}) => {
   )
 }
 
+const MobileSendAdaPage = () => (
+  <Fragment>
+    <SendAdaPage />
+    <div className="mobile-multi-assets-page-wrapper">
```
rizip1, line 168: Are there some conventions for naming classes?

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -76,14 +76,25 @@ type DropdownAssetItem = Token & {
...
+
+export const Policy = ({policyId}: PolicyProps) => (
+  <div className="multi-asset-hash">
+    <span className="ellipsis">{policyId.slice(0, -6)}</span>
+    <span>{policyId.slice(-6)}</span>
+  </div>
+)
+
+// TODO: consider exporting to a standalone file
+export const AssetItem = ({
```
rizip1, line 89: Simple "reuse" till we decide what we want to show.

app/public/css/styles.css
```diff
@@ -3775,6 +3775,23 @@ input:checked + .slider:before {
   margin-bottom: 32px;
 }
 
+/* MULTI ASSETS SCREEN: START */
```
rizip1, line 3778: What do you think about those "start/end" comments? To make the "end" more explicit

app/public/css/styles.css
```diff
@@ -696,21 +696,21 @@ p.hey {
   display: none;
 }
 
-.searchable-select-item .multi-asset-item {
+.multi-asset-item {
```
rizip1, line 699: I removed `.searchable-select-item` from selectors as I am also reusing component with those classes outside of the `select`

app/frontend/components/pages/dashboard/dashboardPage.tsx
DavidTranDucVL: I'm not aware of any.

app/frontend/components/pages/sendAda/multiAssetsPage.tsx
```diff
@@ -0,0 +1,43 @@
...
+        assetNameHex: token.assetName,
+        assetName: assetNameHex2Readable(token.assetName),
+        type: AssetFamily.TOKEN,
+        star: false,
+      })),
+  ]
+
+  return (
+    <div className="card">
+      <h2 className="card-title">Multi assets</h2>
```
DavidTranDucVL, line 25: We probably want to call it more user friendly, up for discussion. Maybe `Tokens`?

app/public/css/styles.css
```diff
@@ -3775,6 +3775,23 @@ input:checked + .slider:before {
...
+  margin-bottom: 32px;
+}
+
+.multi-assets-page-list {
+  max-height: 300px;
+  overflow: auto;
+}
+
+.multi-asset-page-policy {
+  /* TODO: rethink, larger width causes cards on the left to shrink too much */
```
DavidTranDucVL, line 3790: I was thinking about this issue as well. Maybe we can remove flex grow/shrink from the columns?

app/frontend/components/pages/sendAda/multiAssetsPage.tsx
rizip1: Michal suggested "Digital assets"

PeterBenc (commented):
Is a deliberate decision to not make MultiAssetPage into a separate tab in mobile view? I think its ok but I can imagine some user never finding out that there is something under the "Send" page.  Maybe something work discussing with Michael but I would make it into a separate tab. 

image

Regarding the token policyIds, I think its ok to use an ellipsis (I think yoroi does too) but we probably want a `CopyOnClick` functionality.

rizip1:
> Is a deliberate decision to not make MultiAssetPage into a separate tab in mobile view? I think its ok but I can imagine some user never finding out that there is something under the "Send" page. Maybe something work discussing with Michael but I would make it into a separate tab.
> 
> image
> 
> Regarding the token policyIds, I think its ok to use an ellipsis (I think yoroi does too) but we probably want a `CopyOnClick` functionality.

Thanks, I understand your point, lets discuss this again with @MichalPetro

PeterBenc (commented):
LGTM. tho my personal preference would be to have the assetName and policyId in the middle between the two lines (similarly to transaction history) (already edited in the picture below)
image

Also this happens for the first asset when copying ( the popup is cut )
image

Also I would suggest we use similar icon as the address has (in the picture below) for the "Copy on click". I basically guessed that clicking on the policy id would copy it. This is great on small devices but I would show the icon on desktops.

image

rizip1:
@PeterBenc I do not really understand "LGTM. tho my personal preference would be to have the assetName and policyId in the middle between the two lines (similarly to transaction history) (already edited in the picture below)" as I think they are between the two lines :thinking: . The next two suggestions were fixed.

PeterBenc:
> @PeterBenc I do not really understand "LGTM. tho my personal preference would be to have the assetName and policyId in the middle between the two lines (similarly to transaction history) (already edited in the picture below)" as I think they are between the two lines  . The next two suggestions were fixed.

Thnx, and Sorry for not being more clear, 
this is the current look:
image
this I stated as my personal preference
image

My point was only to have the content in the middle between the lines, so som padding/margin would need to be added.

rizip1:
@PeterBenc centering should be fixed, please check the last commit.

PeterBenc (approved):
Thnx great :+1:

PR #880, opened 2021-03-10
Pool owner refactor

PeterBenc:
Changes:
 -  witness pool registration with tokens, with validity interval start and missing ttl
 - refactor pool registration utils
 
 Gotcha:
 - I didn't get to refactor everything that needs to be refactored, especially the part in actions.ts
 - Since adalite currently doesn't support bigint there is a mismatch,


xdzurman:
Attempt with {
    "type": "TxBodyAllegra",
    "description": "",
    "cborHex": "82a50081825820dfdf79a92282e1913cf9bdc2f241987462ebc9e06abee6028b89096c056a2e3d00018182583901746947c4f7346a1e356fe8c8176c99261463621488e806100cf66debb1bc146a5fb0683c4e3836712d115b98619048bc307cc059b6adc76e1a001b7740021a00030d40031905dc04818a03581ce908924a3d410aa9622a23d7775445a3b1e009eb7a93e210144a74ab5820d292b669b3bb3a76c5377e6291800cf24f819be3de66d3d90de9529d33b092ea1b00000006fc23ac001a1443fd00d81e82031864581de11d227aefa4b773149170885aadba30aab3127cc611ddbc4999def61c81581c1d227aefa4b773149170885aadba30aab3127cc611ddbc4999def61c848400190bbb4436e44b9af68400190bb94436e44b9af68400190bbb4422f90b59f68400190bb94422f90b59f682782668747470733a2f2f6164616c6974652e696f2f41444c54392d6d657461646174612e6a736f6e5820559a5abc76e7be62694009acc84d2f46c70d5395b1b626fd13d77aa5e52afaecf6"
}

ends up in 

image

app/frontend/actions.ts
```diff
@@ -1255,89 +1255,75 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
-    }
-
-    const unsignedTxDecoded = decode(file.cborHex)
-    const parsedTx = parseUnsignedTx(unsignedTxDecoded)
-    return parsedTx
-  }
+  /* POOL OWNER */
 
-  const loadPoolCertificateTx = async (state: State, fileObj) => {
+  const loadPoolCertificateTx = async (state: State, file: string) => {
```
xdzurman, line 1260: I'd rename file to fileContent or fileContentStr to make it clearer what it is or what the string actually is (my first guess was file name)

refi93, line 1260: I'd prefer to extract the logic of parsing and translating that to a txPlan to a separate utility function which would have a clear interface and no side-effect on the state. That way we can also unit-test it

app/frontend/actions.ts
```diff
@@ -1347,54 +1333,32 @@ export default ({setState, getState}: {setState: SetStateFn; getState: GetStateF
...
       }
 
+      const {plan, ttl, validityIntervalStart} = state.poolRegTransactionSummary
+
       const txAux = await wallet.getAccount(state.sourceAccountIndex).prepareTxAux(
-        state.poolCertTxVars.plan, // @ts-ignore (Fix byron-shelley formats later)
-        parseInt(state.poolCertTxVars.ttl, 10)
+        plan,
+        // TODO: we should do parsing somewhere else
+        parseCliTtl(ttl),
```
refi93, line 1341: doesn't `state.poolRegTransactionSummary` contain already properly parsed ttl? If not, why?

app/frontend/wallet/shelley/helpers/poolCertificateUtils.ts
```diff
@@ -10,68 +34,64 @@ const enum PoolParamsByteLengths {
...
-  vrfKeyHashHex: string
-  pledgeStr: string
-  costStr: string
-  margin: Margin
-  rewardAccountHex: string
-  poolOwners: Array<PoolOwnerParams>
-  relays: Array<RelayParams>
-  metadata: PoolMetadataParams
+// TODO: fix when migrating to BigInt
+export const checkedNumber = (number: any, variableName: string): number => {
```
refi93, line 78: can the "number" at the input really be anything? It scares me a bit that for example Number(null) is zero, so this function would map a null value to zero without any warning

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -51,13 +81,18 @@ function ShelleyTxAux(
     txBody.set(TxBodyKey.INPUTS, cborizeTxInputs(inputs))
     txBody.set(TxBodyKey.OUTPUTS, cborizeTxOutputs(outputs))
     txBody.set(TxBodyKey.FEE, fee)
-    txBody.set(TxBodyKey.TTL, ttl)
+    if (ttl || ttl === 0) {
```
refi93, line 84: I think this is equivalent to `ttl != null`

app/frontend/actions.ts
refi93, line 1341: I'd expect to do that at the moment of creation of the poolRegTransactionSummary and it can be casted to some branded type like ValidTtl which internal functions then can take for granted to be actually valid

app/frontend/actions.ts
PeterBenc: I separated the logic of parsing the file into a helper function, and converting it to txPlan to another, it seemed strange to me for a function to both parse JSON and also receive a staking address as an argument. Is there any problem with this approach? As I mentioned in the PR description, I didn't get to refactor everything since we don't want to delay release much longer.

app/frontend/wallet/shelley/helpers/poolCertificateUtils.ts
PeterBenc: Good point, I had no idea that Number(null) is 0.. that's pretty strange, anyway, fixed.

app/frontend/wallet/shelley/shelley-transaction.ts
PeterBenc: Fixed.

app/frontend/actions.ts
PeterBenc: Fixed, parsing ttl and validity interval start somewhere else, I didn't create the branded type tho.

app/frontend/actions.ts
refi93: I see, I checked the refactor you did and looks good to me

app/frontend/wallet/shelley/helpers/poolCertificateUtils.ts
```diff
@@ -10,68 +34,64 @@ const enum PoolParamsByteLengths {
...
-  vrfKeyHashHex: string
-  pledgeStr: string
-  costStr: string
-  margin: Margin
-  rewardAccountHex: string
-  poolOwners: Array<PoolOwnerParams>
-  relays: Array<RelayParams>
-  metadata: PoolMetadataParams
+// TODO: fix when migrating to BigInt
+export const validateNumber = (number: number | BigInt, variableName: string): number => {
```
refi93, line 78: sorry for not being clear in the previous comment - the validator is IMHO OK to expect `number: any` (or rather `value: any`) at the input (as the input is unvalidated), but I'd rather add runtime assertions on the input being an instance of number and checking its range.

Also, expecting an instance of BigInt only to reject it if it exceeds MAX_SAFE_INTEGER seems a bit strange and if really needed, I'd rather explicitly acknowledge with a comment that even though some values can be BigInt, the codebase isn't fully migrated to BigInt, that's why we restrict the range even for those

PR #881, opened 2021-03-11
Extend pool recommendation API

xdzurman:
testable with changing
const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/account/poolRecommendation/poolHash/${poolHash}/stake/${stakeAmount}`
to e.g.
const url = `${ADALITE_CONFIG.ADALITE_BLOCKCHAIN_EXPLORER_URL}/api/account/poolRecommendation/poolHash/${poolHash}/stake/25000000000000`

and with staging explorer

For accounts with >=25M, recommend a private pool, handled by BE

### DIFF
--- a/app/frontend/state.ts
+++ b/app/frontend/state.ts
@@ -252,7 +252,9 @@ const initialState: State = {
       stakingHistory: [],
       visibleAddresses: [],
       poolRecommendation: {
-        isInRecommendedPoolSet: true,
+        isInRecommendedPoolSet: false,
+        isInPrivatePoolSet: false,
+        isRecommendationPrivate: false,
         recommendedPoolHash: '',
         status: '',
         shouldShowSaturatedBanner: false,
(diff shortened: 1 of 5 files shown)


app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
```diff
@@ -260,6 +268,7 @@ export default connect(
     rawTransactionOpen: state.rawTransactionOpen,
     stakePool: state.shelleyDelegation.selectedPool,
     txConfirmType: state.txConfirmType,
+    poolRecommendation: getSourceAccountInfo(state).poolRecommendation,
```
PeterBenc, line 271: We should probably remove the connect altogether and not use both connect and useSelector. But we can also refactor this project-wise later.

PeterBenc (approved):
LGTM

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
xdzurman: That's forgotten for sure, I'll remove it

PeterBenc:
I am merging this since it only extends the pool recommendation interface, we should address all needed work in another PR described in #944.

PR #883, opened 2021-03-12
Track crypto provider versions

PeterBenc:
Changes:
 - adds tracking of crypto provider version when submitting transaction
 - for mnemonic it tracks derivationScheme version
 - fixes zero output transactions in refis stats

### DIFF
--- a/app/frontend/types.ts
+++ b/app/frontend/types.ts
@@ -44,6 +44,7 @@ export interface CryptoProvider {
   ensureFeatureIsSupported: (feature: CryptoProviderFeature) => void
   isFeatureSupported: (feature: CryptoProviderFeature) => boolean
   displayAddressForPath: (absDerivationPath: BIP32Path, stakingPath: BIP32Path) => void
+  getVersion: () => string | null
 }
 
 // TODO: remove this and replace with TxCertificateType
--- a/app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
+++ b/app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
@@ -48,6 +48,8 @@ CryptoProviderParams): Promise<CryptoProvider> => {
 
   const getDerivationScheme = () => derivationScheme
 
+  const getVersion = () => null
+
   const deriveXpub = CachedDeriveXpubFactory(
     derivationScheme,
     config.shouldExportPubKeyBulk,
@@ -179,6 +181,7 @@ CryptoProviderParams): Promise<CryptoProvider> => {
     ensureFeatureIsSupported,
     isFeatureSupported,
     displayAddressForPath,
+    getVersion,
   }
 }
 
--- a/server/middlewares/statsGoogleAnalytics.js
+++ b/server/middlewares/statsGoogleAnalytics.js
@@ -69,6 +69,8 @@ const trackTxSubmissions = mung.jsonAsync(async (body, req, res) => {
         : 'otherTxSubmissions'
     const txSubmissionSuccess = body.Right ? 'successful' : 'unsuccessful'
     const txWalletType = req.get('walletType')
+    const txWalletVersion = req.get('walletVersion')
+    const txWalletDerivationScheme = req.get('walletDerivationScheme')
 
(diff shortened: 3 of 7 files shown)


server/middlewares/statsRedis.js
```diff
@@ -49,6 +49,8 @@ const trackTxSubmissions = mung.json((body, req, res) => {
 
     const txWalletType = req.get('walletType')
     incrCountersBy(`${txSubmissionType}:${txWalletType}`, 1)
+    const txCryptoProviderVersion = req.get('cryptoProviderVersion')
+    incrCountersBy(`walletVersions:${txWalletType}-${txCryptoProviderVersion}`, 1)
```
refi93, line 53: if I understand correctly, this way we only track the absolute number of tx submits by wallet version. This isn't really useful as we need to know how the version distribution changes over time to have the insight what's the adoption of newer versions of firmwares as they are rolled out

In general, I think redis is already too polluted and I see it as a place for the most critical business metrics which hw wallet version really isn't. So I'd rather leverage the google analytics stats for that which are not that used but if we need info, it's there and GA also takes care of distributing the data over time without us explicitly caring.

PeterBenc: Actually, this would track the used versions totally, monthly and daily so we would be able to track adoption. Anyway I added the support for google analytics, I ll create a dashboard for it once its merged.

PeterBenc: Btw. It makes sense to drop the middle commit ( redis stats ).

refi93 (approved):

server/middlewares/statsGoogleAnalytics.js
```diff
@@ -84,6 +85,13 @@ const trackTxSubmissions = mung.jsonAsync(async (body, req, res) => {
         value: undefined,
       })
 
+      await trackEvent({
+        ...baseEventData,
+        action: txCryptoProviderVersion,
```
refi93, line 90: maybe txWalletVersion would be more accurate of a name. Our crypto providers as such don't really have a version and this may cause the false impression that they do

refi93 (changes requested):
see comments on naming

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -48,6 +48,8 @@ CryptoProviderParams): Promise<CryptoProvider> => {
 
   const getDerivationScheme = () => derivationScheme
 
+  const getVersion = () => derivationScheme.type
```
refi93, line 51: I think coupling the derivation scheme with "version" is unnecessarily misleading. Technically hw wallets use derivation scheme v2 so I'd decouple that and submit the derivation scheme number and wallet version separately, and it's ok to return explicitly "null" for mnemonic-based wallets - at least it's more faithful to reality

refi93: I'd perhaps even call this "getDeviceVersion" to make it clear this applies to devices

PeterBenc: fixed

server/middlewares/statsGoogleAnalytics.js
PeterBenc: by crypto provider I meant trezor/ledger/mnemonic it self, anyway, renamed to txWalletVersion

app/frontend/types.ts
```diff
@@ -44,6 +44,7 @@ export interface CryptoProvider {
   ensureFeatureIsSupported: (feature: CryptoProviderFeature) => void
   isFeatureSupported: (feature: CryptoProviderFeature) => boolean
   displayAddressForPath: (absDerivationPath: BIP32Path, stakingPath: BIP32Path) => void
+  getVersion: () => string
```
refi93, line 47: this should be string | null now I believe - didn't typescript catch it?

refi93 (approved):
fix that one typing comment pls, otherwise LGTM

PR #884, opened 2021-03-14
Staking key deregister

rizip1:
* Add deregister stake key option
* Refactor modals to be only "centralized" in a single place (due to showing transaction modal from "advanced" tab, though it was finally decided not to be used from there, I still find the refactor useful)
* Texts & visuals of confirmation dialog are inspired by Yoroi
* I tried to divide it into more commits to make it easier to follow

Help wanted in: :pray: 
1. Are there some other dialogs that I might be not aware of and could get broken with this refactor? I tested "transfer/delegate" from accounts screen and those worked fine for me

### DIFF
--- a/app/frontend/components/pages/accounts/accountsDashboard.tsx
+++ b/app/frontend/components/pages/accounts/accountsDashboard.tsx
@@ -5,22 +5,17 @@ import printAda from '../../../helpers/printAda'
 import {State} from '../../../state'
 import {AdaIcon} from '../../common/svg'
 import Alert from '../../common/alert'
-import SendTransactionModal from './sendTransactionModal'
-import DelegationModal from './delegationModal'
-import ConfirmTransactionDialog from '../../../../frontend/components/pages/sendAda/confirmTransactionDialog'
 import AccountTile from './accountTile'
 import {AccountInfo, Lovelace} from '../../../../frontend/types'
 import Conversions from '../../common/conversions'
+import {hasStakingKey} from '../../../selectors'
 
 type DashboardProps = {
   accountsInfo: Array<AccountInfo>
   maxAccountIndex: number
   reloadWalletInfo: any
-  shouldShowSendTransactionModal: boolean
-  shouldShowDelegationModal: boolean
   totalWalletBalance: number
   totalRewardsBalance: number
-  shouldShowConfirmTransactionDialog: boolean
   conversionRates: any
 }
 
@@ -28,11 +23,8 @@ const AccountsDashboard = ({
   accountsInfo,
   maxAccountIndex,
   reloadWalletInfo,
-  shouldShowSendTransactionModal,
-  shouldShowDelegationModal,
   totalWalletBalance,
   totalRewardsBalance,
-  shouldShowConfirmTransactionDialog,
   conversionRates,
 }: DashboardProps) => {
   const InfoAlert = () => (
@@ -94,8 +86,6 @@ const AccountsDashboard = ({
 
   return (
     <Fragment>
-      {shouldShowSendTransactionModal && <SendTransactionModal />}
-      {shouldShowDelegationModal && <DelegationModal />}
       <div className="dashboard-column account">
         <div className="card account-aggregated">
           <div className="balance">
@@ -142,7 +132,9 @@ const AccountsDashboard = ({
                 <AccountTile
                   key={accountInfo.accountIndex}
                   accountIndex={accountInfo.accountIndex}
-                  ticker={accountInfo.shelleyAccountInfo.delegation.ticker}
+                  ticker={
+                    hasStakingKey(accountInfo) && accountInfo.shelleyAccountInfo.delegation.ticker
+                  }
                   availableBalance={accountInfo.balance}
                   rewardsBalance={accountInfo.shelleyBalances.rewardsAccountBalance}
                   shouldShowSaturatedBanner={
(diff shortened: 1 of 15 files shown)


app/frontend/components/pages/accounts/accountsDashboard.tsx
```diff
@@ -142,7 +142,10 @@ const AccountsDashboard = ({
                 <AccountTile
                   key={accountInfo.accountIndex}
                   accountIndex={accountInfo.accountIndex}
-                  ticker={accountInfo.shelleyAccountInfo.delegation.ticker}
+                  ticker={
+                    accountInfo.shelleyAccountInfo.hasStakingKey &&
+                    accountInfo.shelleyAccountInfo.delegation.ticker
```
rizip1, line 147: TODO:

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -313,7 +314,11 @@ const validateTxPlan = (txPlan: TxPlan): void => {
     throw NamedError('OutputTooSmall')
   }
   const totalRewards = txPlan.withdrawals.reduce((acc, {rewards}) => acc + rewards, 0)
-  if (totalRewards > 0 && totalRewards < txPlan.fee) {
+  const deregisteringStakeKey = txPlan.certificates.some(
+    (c) => c.type === CertificateType.STAKING_KEY_DEREGISTRATION
+  )
+  // TODO: confirm this "deregisteringStakeKey" condition
+  if (!deregisteringStakeKey && totalRewards > 0 && totalRewards < txPlan.fee) {
```
rizip1, line 321: Please check if this condition makes sense.
The idea here is that I assume that when deregistering the "stakeKey" the returned "deposit" should be always greater than the "fee" even if the "rewards" are non-zero and less than "fee". Is this assumption correct? Maybe also question on @refi93 ?

app/frontend/selectors.ts
```diff
@@ -0,0 +1,19 @@
...
+import {getActiveAccountInfo} from './state'
+import {AccountInfo} from './types'
+
+/*
+This file contains hooks shared accross multiple components which
+use global state to infer return values.
+Once "actions.ts" is divided into multiple files consider to also divide this file.
+*/
+
+export const useActiveAccount = () => useSelector((state) => getActiveAccountInfo(state))
```
PeterBenc, line 11: Awesome :clap:

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -384,11 +385,34 @@ const prepareTxPlanDraft = (txPlanArgs: TxPlanArgs): TxPlanDraft => {
     }
   }
 
+  const prepareDeregisterStakingKeyTx = (
+    txPlanArgs: DeregisterStakingKeyTxPlanArgs
+  ): TxPlanDraft => {
+    const {withdrawals, outputs} = prepareWithdrawalTx({
```
PeterBenc, line 391: Two notes: 
- we probably dont need to withdraw the rewards if they are equal to 0, just a detail tho
- checkout the `validateTxPlan` function in this file ( https://github.com/vacuumlabs/adalite/blob/af5c06eb105725b939e662ec662b0ae0134d0006/app/frontend/wallet/shelley/shelley-transaction-planner.ts#L314) 
It validates whether the total rewards are 0 or higher than the required fee, (this would cause some of the deregistrations to fail if a user had small rewards) 
We could fix this by passing also the `txType` from `txPlanArgs` and don't check for the total rewards amount in case its deregistration.
I cant say I like this very much but its probably a reasonable fix until we refactor the `selectMinimalTxPlan` function to a reasonable state.

PeterBenc (commented):
Other than the comments LGTM, its maybe worth rebasing this since some of the edited files has been changed meanwhile (including some deregistration logic and fee calculation)

app/frontend/wallet/shelley/shelley-transaction-planner.ts
rizip1: @PeterBenc related to "We could fix this by passing also the txType from txPlanArgs and don't check for the total rewards amount in case its deregistration.". I discussed this with @xdzurman and he claimed that even though the deposit is greater than the fees, the transaction will not pass as only UTXO counts as your "spending power" not your returned deposit. Btw I also changed this condition first, but then if fails on the other checks (e.g. on mainnet) where you do not have any UTXO & are constructing the transaction. Please let me know if you think it should work otherwise, thanks.

rizip1:
@PeterBenc rebased. Please check the last "PR fixes" comment and the above discussion. Thanks

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: Sorry but I am not sure how this relates to deposit or spending power in any way, my only point was that the `validateTxPlan` function throws an error if rewards are lower than the fee. So If a user has rewards `0.01 ADA` and the fee is `0.17 ADA` he basically won't be able to deregister his staking key. 

@rizip1 Probably worth discussing this on a quick call.

PeterBenc (approved):
LGTM

PeterBenc:
Almost forgot, I would discuss with @xdzurman a possibility of showing key deregistration in staking history

PR #895, opened 2021-03-17
Remove byron era code

PeterBenc:
Changes:
 - removes byron wallet logic, 
 - removes byron tests
 - removes some shelley testnet code

Gotcha:
 - open to suggestions, it is possible that some code doesnt relate to byron era specifically but is no longer needed now and I missed it


ppershing:
-2500? Nice :-)

ppershing:
Shouldn't this be merged? I might conflict with lots of refactorings that are underway

refi93 (approved):

app/tests/src/address-manager.js
```diff
@@ -1,154 +0,0 @@
-import assert from 'assert'
```
refi93, line 1: doesn't this remove test coverage for Byron address, which are still used to migrate the funds from Byron-era wallets?

refi93: this code indeed seems to test dead code, but I'd rather introduce fixtures to cover Byron-era addresses so they are not left out - arguably those are the easiest to forget about when doing future modifications

app/tests/src/address-manager.js
```diff
@@ -1,154 +0,0 @@
...
-
-import derivationSchemes from '../../frontend/wallet/helpers/derivation-schemes'
-import CardanoWalletSecretCryptoProvider from '../../frontend/wallet/byron/cardano-wallet-secret-crypto-provider'
-import AddressManager from '../../frontend/wallet/address-manager'
-import mnemonicToWalletSecretDef from '../../frontend/wallet/helpers/mnemonicToWalletSecretDef'
-import {byronAddressManagerSettings} from './common/address-manager-settings'
-import BlockchainExplorer from '../../frontend/wallet/blockchain-explorer'
-
-import mockNetwork from './common/mock'
-import {ByronAddressProvider} from '../../frontend/wallet/byron/byron-address-provider'
```
refi93, line 11: actually, this provider is still being used, so these tests are still relevant

app/tests/src/shelley.js
```diff
@@ -1,106 +0,0 @@
...
-}
-
-// inform mocha global leak detector
-window.wasm = null
-before(loadWasmModule)
-
-describe('shelley address derivation', () => {
-  const mnemonic15Words =
-    'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon address'
-  it('should derive base address from 15-words mnemonic', async () => {
```
refi93, line 31: this test seems to still be testing live code (derivation of a single base address), so I'd keep it

refi93:
@ppershing thanks for notifying, I missed the original e-mail notification with the request for review

refi93 (changes requested):
Seems like some still relevant tests were also removed, see comments

app/tests/src/shelley.js
PeterBenc: It does, but it tests something that is already tested elsewhere so they are redundant.

refi93: I see the full manager being tested, not the individual address providers, but maybe I'm just missing something. If not, this IMHO still serves to test the address provider in isolation which may come in handy for refactors related to the address provider alone, decoupling address-provider-level errors from the address-manager ones

PeterBenc (commented):
@refi93 you made a point, a initially removed all the tests that used some legacy code ( which was being removed) and in that process removed relevant tests for byron address provider and base address provider. Fixed.

app/tests/src/shelley.js
PeterBenc: Fair point, fixed.

app/tests/src/address-manager.js
PeterBenc: You are right, There are no other tests for byron address provider so I ll return these tests. Thnx :+1:

app/tests/src/index.js
```diff
@@ -4,21 +4,15 @@ describe('AdaLite Test Suite', () => {
...
   })
-  // describe('Cardano Wallet', () => {
-  //   require('./cardano-wallet')
-  // })
-  describe('Import/Export Wallet as JSON', () => {
-    require('./keypass-json')
+  describe('Byron Address Manager', () => {
+    require('./address-manager')
   })
+  // describe('Import/Export Wallet as JSON', () => {
```
refi93, line 13: this seems like forgotten commenting-out

refi93 (approved):

PR #902, opened 2021-03-22
Remove balances from global state

rizip1:
**Keys removed from global state**
-> totalWalletBalance
-> totalRewardsBalance

Due to the above refactor those further changes were needed:
-> `isBigDelegator` key was removed and  selector introduced
-> `shouldShowPremiumBanner` was reworked

**shouldShowPremiumBanner**
-> localStorage key renamed to `seenPremiumBanner` as otherwise initial value of `shouldShowPremiumBanner`
is misleading (it could be `true` even though we later override it `false` due to not enough funds)
-> by naming it `seenPremiumBanner` it only focuses on whether the user saw it, and whether to display banner
is decided by the selector

**other notes**
* no "memoization" as for now selectors do not perform any "complex" computations
* what do you think about "selector" postfix? Do you find it redundant or like that it is explicit?
* For easier review, I skipped exchanging `connect` for `useActions | useSelector` for now (I can then push it as a second commit)

Please share your ideas about those changes and whether you have some other tips how to make this refactor more meaningful. I then plan to remove other "redundant" keys in a similar fashion.

### DIFF
--- a/app/frontend/selectors.ts
+++ b/app/frontend/selectors.ts
@@ -1,16 +1,39 @@
+import {State, getActiveAccountInfo} from './state'
+import {BIG_DELEGATOR_THRESHOLD, PREMIUM_MEMBER_BALANCE_TRESHOLD} from './wallet/constants'
 import {useSelector} from './helpers/connect'
-import {getActiveAccountInfo} from './state'
 import {AccountInfo} from './types'
 
 /*
-This file contains hooks shared accross multiple components which
+This file contains hooks/selectors shared accross multiple components which
 use global state to infer return values.
 Once "actions.ts" is divided into multiple files consider to also divide this file.
 */
 
-export const useActiveAccount = () => useSelector((state) => getActiveAccountInfo(state))
+export const totalWalletBalanceSelector = (state: State): number =>
+  state.accountsInfo.reduce((a, {balance}) => balance + a, 0)
 
-export const hasStakingKey = (account: AccountInfo) => account.shelleyAccountInfo.hasStakingKey
+export const totalRewardsBalanceSelector = (state: State): number =>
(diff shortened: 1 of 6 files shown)


app/frontend/selectors.ts
```diff
@@ -0,0 +1,18 @@
+import {State} from './state'
+import {BIG_DELEGATOR_THRESHOLD, PREMIUM_MEMBER_BALANCE_TRESHOLD} from './wallet/constants'
+
+export const totalWalletBalanceSelector = (state: State) =>
```
ppershing, line 4: make return types explicit. This helps with potential refactorings

app/frontend/selectors.ts
```diff
@@ -0,0 +1,18 @@
...
+  state.accountsInfo.reduce((a, {balance}) => balance + a, 0)
+
+export const totalRewardsBalanceSelector = (state: State) =>
+  state.accountsInfo.reduce((a, {shelleyBalances}) => shelleyBalances.rewardsAccountBalance + a, 0)
+
+export const isBigDelegatorSelector = (state: State) => {
+  const totalWalletBalance = totalWalletBalanceSelector(state)
+  return totalWalletBalance > BIG_DELEGATOR_THRESHOLD
```
ppershing, line 12: now, the big question is where should business logic like this reside :thinking: I am not a big fan of hard-coding it in the selectors but I don't see a better place for it right now

ppershing:
Overall seems legit

app/frontend/selectors.ts
rizip1: I think that e.g. this could be exposed by `wallet` module, not exposing the `BIG_DELEGATOR_THRESHOLD` constant (it could be hidden to the outside world), but exposing something like "isBigDelegator" function. Though I believe such refactor is out the scope of this PR, but I believe there will be time for such changes in the near future :)

PeterBenc (approved):
LGTM

PR #904, opened 2021-03-24
Avoid redirect for review apps

rizip1:
* After this is merged, all newly created PRs will have automatically created environments
* If you want to use it for already opened PR, then rebase & reopen them

For testnet CORS were allowed for `https://adalite-*.herokuapp.com`.

### DIFF
--- a/server/helpers/loadConfig.js
+++ b/server/helpers/loadConfig.js
@@ -73,7 +73,8 @@ const {
   PORT,
   REDIS_URL,
   ADALITE_ENABLE_DEBUGGING,
-  ADALITE_SERVER_URL,
+  HEROKU_APP_NAME,
+  HEROKU_IS_REVIEW_APP,
   ADALITE_BLOCKCHAIN_EXPLORER_URL,
   ADALITE_DEFAULT_ADDRESS_COUNT,
   ADALITE_GAP_LIMIT,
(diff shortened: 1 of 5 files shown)


server/middlewares/redirectToBaseUrl.js
```diff
@@ -1,5 +1,9 @@
 module.exports = (req, res, next) => {
-  if (`${req.protocol}://${req.get('host')}` !== process.env.ADALITE_SERVER_URL) {
+  if (process.env.HEROKU_IS_REVIEW_APP === 'true') {
```
refi93, line 2: I'm unable to log into the wallet (tried logging with mnemonic), as it fails on CORS, that can be solved by `ADALITE_ENABLE_SERVER_MOCKING_MODE` being set to true (a flag we use to mock requests to tx submission, with the side effect that all requests are being redirected through the heroku server so no CORS issues)

Another problem is that ADALITE_SERVER_URL is used across the app to know what url to call - that can be set with a post-deploy script I believe: https://stackoverflow.com/questions/38087125/set-host-name-as-an-environment-variable-in-heroku-review-app

rizip1: Thanks, will investigate that.

rizip1:
@refi93 @PeterBenc Calling `POST: https://adalite-disable-redirec-04s0yd.herokuapp.com/api/txs/submit` when sending funds between accounts results in => `{"message":"Cannot read property 'trim' of undefined"}`. Any immediate ideas?

rizip1:
@refi93 please re-check.

server/helpers/loadConfig.js
```diff
@@ -69,7 +69,8 @@ const {
   PORT,
   REDIS_URL,
   ADALITE_ENABLE_DEBUGGING,
-  ADALITE_SERVER_URL,
+  HEROKU_APP_NAME,
+  HEROKU_IS_REVIEW_APP,
```
xdzurman, line 73: just a sanity check - these are injected by heroku? If they're required in local .env, add them to .env.example. If not, ignore this

xdzurman (approved):

server/helpers/loadConfig.js
rizip1: `HEROKU_APP_NAME` is injected by Heroku and `HEROKU_IS_REVIEW_APP` is set by us. I did not add `HEROKU_IS_REVIEW_APP` into `.env.example` as for any local development we just should not set it. As written in the comment in code, `HEROKU_IS_REVIEW_APP` is introduced just because `HEROKU_APP_NAME` is not considered stable, I want to avoid the case when it would suddenly appear for all heroku apps (e.g. staging :) )

PR #908, opened 2021-03-24
Divide actions part 1

rizip1:
Idea for the future is:

-> have multiple "Actions" files
-> later move "actions.ts" into "actions" folder and rename it to "index.ts"
-> if there would be some cyclic-dependencies, extract such "functions/actions" into "actions/common.ts"
-> make all "actions" receive "state" as first object
-> avoid using "setState" outside the "actions"
-> actions can use other actions if not causing cycling deps, e.g. wallet uses "loading/error" actions
-> actions should receive "state" object as first param and optional params object as second param (e.g. error case)

I wanted to avoid more changes in this PR not to grow to big.

**Please pay attention to**
I removed "config/extra" params from loading/error as I believe this is "hacky" pattern and those actions/functions should not have such responsibility.
**Global wallet** object is exposed via `getWallet`.

I ADVISE to review by commits, should be easier to follow.

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -11,24 +11,15 @@ import {
   tokenAmountValidator,
 } from './helpers/validators'
 import debugLog from './helpers/debugLog'
-import getConversionRates from './helpers/getConversionRates'
 import sleep from './helpers/sleep'
-import {NETWORKS, WANTED_DELEGATOR_STAKING_ADDRESSES} from './wallet/constants'
-import {CryptoProviderType} from './wallet/types'
 import NamedError from './helpers/NamedError'
 import {exportWalletSecretDef} from './wallet/keypass-json'
-import mnemonicToWalletSecretDef from './wallet/helpers/mnemonicToWalletSecretDef'
 import sanitizeMnemonic from './helpers/sanitizeMnemonic'
-import {initialState} from './store'
-import captureBySentry from './helpers/captureBySentry'
-import {State, GetStateFn, SetStateFn, getSourceAccountInfo} from './state'
-import ShelleyCryptoProviderFactory from './wallet/shelley/shelley-crypto-provider-factory'
-import {ShelleyWallet} from './wallet/shelley-wallet'
+import {State, getSourceAccountInfo, Store} from './state'
 import {TxPlan, TxPlanResult} from './wallet/shelley/shelley-transaction-planner'
 import getDonationAddress from './helpers/getDonationAddress'
 import {localStorageVars} from './localStorage'
 import {
-  AccountInfo,
   Lovelace,
   CryptoProviderFeature,
   Address,
@@ -45,8 +36,9 @@ import {
 } from './types'
 import {MainTabs} from './constants'
 import {parseCliUnsignedTx} from './wallet/shelley/helpers/stakepoolRegistrationUtils'
-
-let wallet: ReturnType<typeof ShelleyWallet>
+import errorActions from './actions/error'
+import loadingActions from './actions/loading'
+import walletActions, {getWallet} from './actions/wallet'
 
 const debounceEvent = (callback, time) => {
   let interval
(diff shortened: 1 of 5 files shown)


rizip1:
@ppershing In the "draft" PR you asked `"General concern: are we sure the state you are passing here is always up to date?"`
Of course, when one action is calling another action, as is passing state to it, it might not be the case, it can be out of sync. However, this is just the responsibility of the caller (action calling other action) and I do not think we can/should pay some other attention to it.

app/frontend/actions/error.ts
```diff
@@ -0,0 +1,34 @@
+import debugLog from '../helpers/debugLog'
+import captureBySentry from '../helpers/captureBySentry'
+import {Store, State} from '../state'
+
+export default ({setState}: Store) => {
+  return {
+    // TODO: remove "options", as "setError" should not be responsible for setting other
+    // parts of the state
+    setError: (
+      state: State,
```
ppershing: are we sure we want to pass `state` as param if we are not using it? Now that we have typescript, we can afford variability in parameters

rizip1: I was thinking about this too, though I thought that it might be good to be explicit about the fact that all actions should receive "state". You can then avoid error when you add "action params" as first argument and are surprised why it does not work.

app/frontend/actions.ts
```diff
@@ -375,14 +169,14 @@ export default (store: Store) => {
     const newState = getState()
```
ppershing, line 169: :scream:  what? shouldn't `state` be the thing?

app/frontend/actions.ts
```diff
@@ -375,14 +169,14 @@ export default (store: Store) => {
...
     try {
       if (newState.sendTransactionSummary.plan) {
-        txAux = await wallet
+        txAux = await getWallet()
           .getAccount(state.sourceAccountIndex)
```
ppershing, line 173: :scream: why are we mixing state and newState at all? :confused:

rizip1: Maybe let's discuss this on slack call? I am not sure whether your comment is about the refactor / current state or something else.

app/frontend/actions/error.ts
```diff
@@ -0,0 +1,34 @@
...
+      {errorName, error, options}: {errorName: string; error: any; options?: any}
```
PeterBenc: What is the motivation behind wrapping the "action parameters" in object, e.g. (errorName, error, options) =>  {errorName, error, options}? Is this some pattern we want to follow? Just curious since this makes the usage of the `setError` action a bit more cumbersome in some cases. For example instead of 
```
catch(e) {
  setError(state, 'someError', e) 
}
```
we would need to create the object every time since no current error type has the structure {errorName, error}...

app/frontend/actions/loading.ts
```diff
@@ -0,0 +1,32 @@
+import {Store, State} from '../state'
+
+export default ({setState}: Store) => {
+  return {
+    // TODO: remove "optionalArgsObj", as "loadingAction" should not be responsible for setting
+    // unrelated parts of the state
+    loadingAction: (state: State, message: string, optionalArgsObj?: any) => {
```
PeterBenc: Do you like defining the functions in the return statement more than having them defined like this? 
```
const someAction= () => {someFunctionBody}
return {
 someAction,
}
```
I probably find the second approach a bit more readable but maybe there is a advantage to you approach I am not seeing.

rizip1: Good point :+1: will change that

app/frontend/actions/error.ts
rizip1: The motivation comes just from the fact that in redux it is a common pattern that the reducer receives an object of shape:
```
{type: UNIQUE_ACTION_NAME, payload: {})
```
Where payload is the object containing all params.  I guess that in this example it does not add much value (more verbose), but this was the motivation. What do you think @ppershing / @PeterBenc  should I revert this back?

ppershing: having explicit "options" argument is actually a good pattern in TS as it helps to be more explicit and also provides less potential for wrong argument order.

PeterBenc: I totally fine with the payload being an object, I was just curious about the motivation. :+1:

app/frontend/actions/loading.ts
rizip1: @PeterBenc fixed in the last comment & rebased

rizip1: Hmm some parts, were not rebased correctly will fix this

rizip1: Actually, should be fine.

rizip1:
@PeterBenc how do you see this one? I would prefer to merge it ASAP as it would block other refactor. Thanks

PeterBenc (approved):
thnx LGTM

PR #910, opened 2021-03-25
Update ledger lib to 3.0.0

PeterBenc:
Changes: 
- updates ledger lib to 3.0.0
- remove local ledger types


app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -1,6 +1,6 @@
 import LedgerTransportU2F from '@ledgerhq/hw-transport-u2f'
 import LedgerTransportWebusb from '@ledgerhq/hw-transport-webusb'
-import Ledger, {AddressTypeNibbles} from '@cardano-foundation/ledgerjs-hw-app-cardano'
+import Ledger, * as LedgerTypes from '@cardano-foundation/ledgerjs-hw-app-cardano'
```
PeterBenc, line 3: What do you think about importing the types this way?

app/frontend/wallet/helpers/tokenFormater.ts
```diff
@@ -35,9 +35,11 @@ export const aggregateTokenBundles = (tokenBundle: TokenBundle[]): TokenBundle =
 
 export const groupTokenBundleByPolicyId = (
   tokenBundle: TokenBundle
-): {[policyId: string]: TokenBundle} => {
+): {[policyId: string]: Omit<Token, 'policyId'>[]} => {
   return _(tokenBundle)
+    .orderBy(['policyId', 'assetName'], ['asc', 'asc'])
     .groupBy(({policyId}) => policyId)
+    .mapValues((tokens) => tokens.map(({assetName, quantity}) => ({assetName, quantity})))
```
PeterBenc, line 42: This is rather a hotfix solution (I experienced some bugs with different token orders) and I think we should make it more explicit how we sort the assets and address this in a separate PR.

app/frontend/wallet/shelley/helpers/version-check.ts
```diff
@@ -1,10 +1,5 @@
-interface Version {
-  major: string
-  minor: string
-  patch: string
-}
-
-export const hasRequiredVersion = (currentVersionStr: Version, requiredVersionStr) => {
+// TODO: type version
+export const hasRequiredVersion = (currentVersionStr, requiredVersionStr) => {
```
PeterBenc, line 2: Temporarily removed the typing since we should refactor this version checking and I don't want to address it in this PR (potentially making changes to all cryptoProviders and also actions, wallet etc.)

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
rizip1: I personally like it

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -114,7 +106,7 @@ const ShelleyLedgerCryptoProvider = async ({
   const ledger = new Ledger(transport)
   const derivationScheme = derivationSchemes.v2
 
-  const version = await ledger.getVersion()
+  const {version} = await ledger.getVersion()
 
   const getVersion = (): string => `${version.major}.${version.minor}.${version.patch}`
```
rizip1, line 111: Not related to this PR, but why is this a function and not just called `versionString` without being a function?

rizip1 (approved):
Looks good to me, though @refi93 it might be a good idea if you also take a look.

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -211,7 +210,7 @@ const ShelleyLedgerCryptoProvider = async ({
     return Object.entries(tokenObject).map(([policyId, assets]) => {
       const tokens = assets.map(({assetName, quantity}) => ({
         assetNameHex: assetName,
-        amountStr: quantity.toString(),
+        amount: quantity.toString(),
```
ppershing, line 213: what type is quantity? We support numbers (if they are less than max safe integer) and native bigints. Are we using some non-native bigint library? :thinking:

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -220,80 +219,164 @@ const ShelleyLedgerCryptoProvider = async ({
...
+  ): LedgerTypes.Certificate {
+    const {stakingAddress, poolRegistrationParams} = certificate
+    const poolOwners = preparePoolOwners(poolRegistrationParams.poolOwners, stakingAddress, path)
+    const margin = {
+      numerator: poolRegistrationParams.margin.numeratorStr,
+      denominator: poolRegistrationParams.margin.denominatorStr,
+    }
+    const metadata = poolRegistrationParams.metadata
+      ? {
+        metadataUrl: poolRegistrationParams.metadata.metadataUrl,
```
ppershing, line 351: isn't this identity transformation?

PeterBenc: You are right, I am not sure why I did it that way, fixed thnx.

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
PeterBenc: quantity is of type `number` as everything currently in the project. I believe there is already an issue for migrating to native bigint.

refi93: yes, but the point is the current ledgerjs api accepts all three representations of numbers - BigInt, number (up to max safe integer) or stringified number

app/frontend/wallet/helpers/tokenFormater.ts
```diff
@@ -35,9 +35,11 @@ export const aggregateTokenBundles = (tokenBundle: TokenBundle[]): TokenBundle =
 
 export const groupTokenBundleByPolicyId = (
   tokenBundle: TokenBundle
-): {[policyId: string]: TokenBundle} => {
+): {[policyId: string]: Omit<Token, 'policyId'>[]} => {
   return _(tokenBundle)
+    .orderBy(['policyId', 'assetName'], ['asc', 'asc'])
```
refi93, line 40: this may still not work properly for assetNames - the cbor canonical ordering instructs to put shorter keys first:

```
*  If two keys have different lengths, the shorter one sorts earlier;
*  If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier.
```


source: https://tools.ietf.org/html/rfc7049#section-3.9

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
refi93: it's an outward facing function, part of the general crypto provider interface, and other crypto providers also have a getter for the version - trezor/mnemonic crypto provider should be able to migrate to the version as a property as well, but probably not worth it

app/frontend/wallet/helpers/tokenFormater.ts
refi93: I think that an explanatory comment why order matters and bunch of unit tests covering that behavior would actually do in this case

refi93: ah wait, but this is returning an object - this doesn't really give any reasonable guarantees when it comes to keys ordering. It would make sense then to create a separate helper function which transforms the token bundle into a list of tuples - the format that both Ledger and Trezor accept

PeterBenc: Thnx, fixed in another PR.

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
PeterBenc: Yes I am aware I see that probably Pershing point and also yours was that why we are converting it to string if it accepts numbers from now on, the answer is that I ve just chosen the most straightforward way of integrating the ledger app and haven't given this a thought. Either way, I ll remove the conversion to string. Thnx.

app/package.json
```diff
@@ -23,7 +23,8 @@
     "test-chrome-headless": "yarn test && node node_modules/.bin/mocha-headless-chrome -f tests/index.html -a disable-web-security -a user-data-dir=tests/.chrome -a disable-site-isolation-trials"
   },
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "^2.2.1",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "https://github.com/vacuumlabs/ledgerjs-cardano-shelley/releases/download/v3.0.0-rc.2/cardano-foundation-ledgerjs-hw-app-cardano-v3.0.0-rc.2.tgz",
+    "@emurgo/js-chain-libs": "https://github.com/SebastienGllmt/js-chain-libs-pkg",
```
SebastienGllmt, line 27: Do you need this for anything? These are the Jormungandr bindings

PeterBenc: I have no idea how that got back in there, not intentionally for sure.. :D , Thnx for noticing :+1:

refi93 (approved):
Works as expected, however there is a known bug with zero values being rejected by ledgerjs introduced in the new 3.0.0 version: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/issues/101 - this will be fixed in the next release.

For now the impact would be that e.g. testnet pool registrations would fail if a zero pool margin is provided. I can't think of other case where we would be potentially submitting a zero value in an Adalite transaction, so I don't consider that as something that needs to be fixed ASAP

PeterBenc:
> Works as expected - I tested an ordinary transaction, stake delegation, staking key registration/deregistration and pool creation as an owner. However, I just found a bug with zero values being rejected by ledgerjs, introduced in the new 3.0.0 version: [vacuumlabs/ledgerjs-cardano-shelley#101](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/issues/101) - this will be fixed in the next release. The catalyst PR of ledgerjs has that already fixed ([vacuumlabs/ledgerjs-cardano-shelley#99](https://github.com/vacuumlabs/ledgerjs-cardano-shelley/pull/99))
> 
> For now the impact would be that e.g. testnet pool registrations would fail if a zero pool margin is provided. I can't think of other case where we would be potentially dealing with a zero amount in an Adalite transaction, so I don't consider that as something that needs to be fixed ASAP

I actually encountered this as well when trying to withdraw 0 rewards, tho I thought this is a wanted behavior. Thanks a lot for the testing.

Issue #890, opened 2021-03-17
Centralize errors making them to typed errors, get rid of "NamedError"


rizip1:
Currently, errors are given "types" by setting their name with a string identifier, which is not centrally controlled and makes checking its specific type unwieldy. Possibly a good source of inspiration could be how errors are defined in ledgerjs library recently refactored by Ppershing: https://github.com/vacuumlabs/ledgerjs-cardano-shelley/blob/develop/src/errors/deviceStatusError.ts

https://docs.google.com/document/d/1w0nLLw1OuC9LpZLQ_p58W1dkdJglfbyH5GjGL_cuKMU/edit

PR #911, opened 2021-03-25
Refactor-errors-1

PeterBenc:
Changes:

- enum for internal/external/unexpected errors
- translations split into external and internal
- all error related stuff in a separate folder with index.ts

TODO"
- fix the failing integration test
- test it manually a bit more

Gotcha:
- we should dive deeper into which errors are not longer thrown anywhere and remove them from enums and translations
- this is a first step in refactoring the errors, open to suggestions of nexts steps


app/frontend/components/pages/staking/stakingPage.tsx
```diff
@@ -67,7 +66,7 @@ const useEmailSubmitProps = () => {
       if (emailSubmitResult.Left) {
         didSucceed = false
         message = emailSubmitResult.Left
-        throw NamedError('EmailSubmissionRejected')
+        throw Error('EmailSubmissionRejected')
```
ppershing, line 69: Errors should always be classes and should be called with `new`

app/frontend/translations.ts
```diff
@@ -3,168 +3,134 @@ import debugLog from './helpers/debugLog'
...
 
 const ledgerTroubleshootingSuggestion =
   'If you are using Ledger, please make sure Ledger Live app is closed and try connecting your device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger"). For more information please read the section concerning Ledger in our troubleshooting suggestions.'
 
-const translations = {
-  SendAddressInvalidAddress: () => 'Invalid address',
-  SendAmountIsNan: () => 'Invalid format: Amount has to be a number',
-  SendAmountIsNotPositive: () => 'Invalid format: Amount has to be a positive number',
-  SendAmountInsufficientFunds: ({balance}) =>
+const translations: {[key in InternalErrorReason]: (params?: any) => string} = {
```
ppershing, line 13: do we even have translations of something else than errors? If not, it shouldn't be called "translations"

app/frontend/errors.ts
```diff
@@ -0,0 +1,72 @@
+export enum Errors {
```
rizip1: In enum-like semantics, this should use singular (`Error`). Otherwise it is also strange when you use it like `getTranslation(code: Errors)` instead of `getTranslation(code: Error)`. As `code` implies singular and `Errors` plural it can become confusing.

app/frontend/errors/NamedError.ts
```diff
@@ -1,11 +1,17 @@
...
 }
 
-export function NamedError(name: string, optionalParams: OptionalParams = {}) {
-  const e = new Error(optionalParams.message || name)
-  e.name = name
+export function NamedError(
+  name: UnexpectedErrorReason | ExpectedErrorReason,
+  optionalParams: OptionalParams = {}
+) {
+  const e = new Error(optionalParams.message || ((name as unknown) as any))
```
rizip1, line 14: Why does it complain? Are not all the reason keys just strings? :thinking:

app/frontend/translations.ts
refi93: agree, something like `errorReasons` sounds more suitable

refi93:
@PeterBenc is there anything blocking the implementation of changes suggested by @ppershing ?

app/frontend/errors/NamedError.ts
```diff
@@ -0,0 +1,21 @@
+import {InternalErrorReason} from './internalErrorReason'
+import {UnexpectedErrorReason} from './unexpectedErrorReasons'
+
+type OptionalParams = {
+  message?: string
+  causedBy?: Error
+}
+
+export function NamedError(
+  name: UnexpectedErrorReason | InternalErrorReason,
```
rizip1, line 10: Assuming correctly that this gets never fired with `ExternalErrorReason`?

PeterBenc:
> @PeterBenc is there anything blocking the implementation of changes suggested by @ppershing ?

No there isn't, I just didn't get to it, and the changes suggested by Pershing are pretty simple to fix so I was just waiting for other comments.

app/frontend/translations.ts
rizip1: `const errorMessages`?

app/frontend/translations.ts
```diff
@@ -178,12 +144,70 @@ const translations = {
...
+  [ExternalErrorReason.TransportWebUSBGestureRequired]: () =>
+    `TransportWebUSBGestureRequired: ${ledgerTroubleshootingSuggestion}`,
+  [ExternalErrorReason.NotFoundError]: () => `NotFoundError: ${ledgerTroubleshootingSuggestion}`,
+  [ExternalErrorReason.AbortError]: () => `NotFoundError: ${ledgerTroubleshootingSuggestion}`,
+  [ExternalErrorReason.SecurityError]: () => `Access denied: ${ledgerTroubleshootingSuggestion}`,
+  [ExternalErrorReason.RedundantStakePool]: () => 'This stake pool is already chosen.',
+  [ExternalErrorReason.DelegationFeeError]: () => 'Unsuccessful delegation fee calculation.',
+}
+
+function getTranslation(code: InternalErrorReason | ExternalErrorReason, params = {}) {
```
rizip1, line 202: `getErrorMessage`?

app/frontend/translations.ts
refi93: ah sorry, I see "ErrorReason" is already used, so errorMessages perhaps

rizip1:
I do not have any general objections. IMHO it would just need to fix issues reported by @ppershing and a lot of local testing before merge :muscle:

app/frontend/errors/internalErrorReason.ts
```diff
@@ -0,0 +1,53 @@
+export enum InternalErrorReason {
+  SendAddressInvalidAddress,
```
refi93, line 2: these enum values would be integers if I understand correctly, which is not consistent with how ExternalErrorReasons are declared and hence may cause confusion. Isn't this also related to @rizip1 's comment https://github.com/vacuumlabs/adalite/pull/911/files#r607883464 ?

app/frontend/translations.ts
```diff
@@ -178,12 +144,70 @@ const translations = {
...
+  if (InternalErrorReason[code]) {
```
refi93, line 203: I think this maps the errors correctly only thanks to the ExternalErrorReason being declared as a string enum, otherwise, as an integer-based enum it would clash with InternalErrorReason enum

refi93: If we refactor the InternalError and ExternalError to classes inheriting from a NamedError (arguably something like BaseError would be more suitable of a name then though), we can differentiate here based on "typeof" statement (provided the getTranslation API is refactored to receive the error class itself, which should be straightforward)

refi93: I think this points to inherent issues with relying on ErrorReason enums - they just won't mix well as enums don't have an explicit type hence are prone to collisions in logic that deals with multiple enums like this one - we discussed this with @PeterBenc and concluded that it's worth wrapping the InternalErrorReason/UnexpectedErrorReason/etc. in proper classes to be able to properly differentiate them in the logic and make ExternalErrorReason a mapping instead of an enum, as it isn't really an exhaustive set of all external errors, only the known ones and should be apprroached as such to avoid confusion/bugs

PeterBenc (commented):
In the end, I decided to make error enums into string enums. mainly due to compatibility, having a numerical enum would need a refactor on multiple other places so I d leave that for a separate Pr where I ll be refactoring how errors are set to state and all related. I updated the description in #890 adding some todos with which I would like to start the next error handling refactor.

app/frontend/errors/errorMessages.ts
```diff
@@ -180,12 +142,68 @@ const translations = {
...
+  [knownExternalErrors.TransportWebUSBGestureRequired]: () =>
+    `TransportWebUSBGestureRequired: ${ledgerTroubleshootingSuggestion}`,
+  [knownExternalErrors.NotFoundError]: () => `NotFoundError: ${ledgerTroubleshootingSuggestion}`,
+  [knownExternalErrors.AbortError]: () => `NotFoundError: ${ledgerTroubleshootingSuggestion}`,
+  [knownExternalErrors.SecurityError]: () => `Access denied: ${ledgerTroubleshootingSuggestion}`,
+  [knownExternalErrors.RedundantStakePool]: () => 'This stake pool is already chosen.',
+  [knownExternalErrors.DelegationFeeError]: () => 'Unsuccessful delegation fee calculation.',
+}
+
+function getErrorMessage(code: InternalErrorReason | string, params = {}) {
```
rizip1, line 198: Related to part of @refi93 comment:
```
provided the getTranslation API is refactored to receive the error class itself, which should be straightforward as we are currently just extracting the code from errors when invoking getTranslation()
```
So this is planned to be reworked in subsequent PR and are currently all enum keys in those enums distinct so that they do not clash for some values? If not addressed in here and there is no clash, please at least add a comment that it should be done. Thanks

Also I do not see it suggested in here https://github.com/vacuumlabs/adalite/issues/890? :thinking:

PeterBenc: I made sure they do not clash, I added a todo comment and also updated the issue description. :+1: Previously it was not specifically mentioned only because the `getErrorMessage` relates to `how we set errors to state`.

rizip1 (approved):

Issue #888, opened 2021-03-17
Cypress setup


rizip1:
* Create initial cypress setup as a safe haven before another planned refactoring
* Test successful login using mnemonic & transaction creation
* Ask @Siegrift for help with initial setup

xdzurman:
closed by #912 , although some improvements could be added, such as:
- stub login (currently takes around 8s in every test)
- more in-depth testing, currently only the most critical functionality and mostly happy cases are coverred

PR #912, opened 2021-03-25
Introduce initial cypress tests to local and circleCI environments

xdzurman:
- Adds simple tests for Send, Receive, Staking and Advanced tabs
- Adds cypress guides to documentation
- Adds cypress tests to circleCI pipeline
- As best practices instruct, "data-cy=" elements were added to elementes with possibility of renaming 
- Sets up video artifacts for cypress pipeline for easy debugging (can be found in branch > workflow > build > Artifacts tab) [reference](https://app.circleci.com/pipelines/github/vacuumlabs/adalite/1573/workflows/f435fbce-13cb-4dd1-8065-a63959fcf27a/jobs/3377/artifacts)
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,28 +1,17 @@
-# Javascript Node CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-javascript/ for more details
-#
+# Orb 'cypress-io/cypress@1' resolved to 'cypress-io/cypress@1.27.0'
 version: 2
 jobs:
   build:
     docker:
-      # specify the version you desire here
-      - image: circleci/node:11.10-browsers
-
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/mongo:3.4.4
-
+      - image: circleci/node:14.16-browsers
     working_directory: ~/repo
-
     steps:
       - checkout
 
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "yarn.lock" }}-{{ checksum "app/yarn.lock"}}
+            - dependencies-{{ checksum "yarn.lock" }}-{{ checksum "app/yarn.lock"}}
 
       - run: cp .env.example .env
       - run: yarn install
@@ -34,8 +23,31 @@ jobs:
           paths:
             - node_modules
             - app/node_modules
-          key: v1-dependencies-{{ checksum "yarn.lock" }}-{{ checksum "app/yarn.lock"}}
+            - ../.cache/Cypress
+          key: dependencies-{{ checksum "yarn.lock" }}-{{ checksum "app/yarn.lock"}}
 
       # run tests!
       - run: yarn eslint
       - run: yarn test-chrome-headless
+      - run:
+          name: Build the app so that server can be started
+          command: cd app && yarn build && cd ../
+      - run:
+          name: Start server with correct env for cypress tests
+          command: yarn cypress:dev
+          wait-on: https://localhost:3000
+          background: true
+      - run:
+          name: Run Cypress tests
+          store_artifacts: true
+          command: VIDEO=true yarn cypress:run
+      - store_artifacts:
+          path: app/cypress/videos
+      - store_artifacts:
+          path: app/cypress/screenshots
+
+workflows:
+  version: 2
+  workflow:
+    jobs:
+      - build
--- a/README.md
+++ b/README.md
@@ -85,6 +85,27 @@ yarn test
 
 Open `app/tests/index.html` in browser
 
+#### Cypress tests
+
+Cypress requires the app to be running with correct environment variables. Type
+```
+yarn cypress:dev
+```
+to start the app with the correct settings. It calls `yarn dev` itself, but with correct environment variables for testing.
+
+After that, you can either type
+```
+yarn cypress:open
+```
+to start the tests with the interactive Test runner, which allows time travel, logs, pausing, etc.
+
+Or, to run the tests in a headless fashion, type
+```
+yarn cypress:run
+```
+
+If you wish to create a video of the tests, pass in `VIDEO=true yarn cypress:run` and videos of all test suites will be created in /app/cypress/videos.
+
 Check us out at https://www.adalite.io
 
 ---
--- a/app/cypress/.env.cypress
+++ b/app/cypress/.env.cypress
@@ -1,0 +1,63 @@
+# injected by heroku, hence without "ADALITE_" prefix
+PORT=3000
+REDIS_URL=''
+ADALITE_ENABLE_DEBUGGING=false
+ADALITE_SERVER_URL='https://localhost:3000'
+ADALITE_BLOCKCHAIN_EXPLORER_URL='https://explorer-testnet.adalite.io'
+ADALITE_DEFAULT_ADDRESS_COUNT=10
+ADALITE_GAP_LIMIT=20
+ADALITE_STATS_PWD=admin
+
+ADALITE_DEMO_WALLET_MNEMONIC='hill safe victory sun tired fetch police radio sport output buyer deny april fringe stumble'
+
+ADALITE_LOGOUT_AFTER=900
+ADALITE_SUPPORT_EMAIL='support@test.test'
+
+# to tell foreign requests from adalite ones
+ADALITE_BACKEND_TOKEN=""
+
+# to distinct between enviroments
+ADALITE_ENV='local'
+
+# mock server config
+ADALITE_ENABLE_SERVER_MOCKING_MODE=true
+ADALITE_MOCK_TX_SUBMISSION_SUCCESS=true
+ADALITE_MOCK_TX_SUMMARY_SUCCESS=true
+ADALITE_DEVEL_AUTO_LOGIN = false
+
+# for development of trezor firmware
+ADALITE_TREZOR_CONNECT_URL=
+
+# google analytics
+ADALITE_GA_TRACKING_ID=
+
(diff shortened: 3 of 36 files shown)


.circleci/config.yml
```diff
@@ -34,8 +23,37 @@ jobs:
...
+      - store_artifacts:
+          path: app/cypress/videos
+      - store_artifacts:
+          path: app/cypress/screenshots
+
+workflows:
+  version: 2
+  workflow:
+    jobs:
+      - build
```
PeterBenc, line 59: A bit unrelated to the cypress tests, but I thought about splitting this workflow into multiple jobs, 
Problem: 

if yarn audit fails, all the other steps are automatically canceled, in that case ( and I guess it happened to all of us at some point) I go ahead and run yarn audit locally, fix the problem just to find out after another push that now eslint is failing, so I go ahead and fix that, just to find out after another push that but some test are failing. :smile: 
Of course I can always run this locally before every commit tho I would prefer to rely on cicd to tell me, and in that case I would like it to tell me everything that's wrong.

Solution:
I don't really see into ci-ci that much but I guess splitting the workflow into multiple jobs, (build, static checks (like eslint), unit tests, integration tests) would solve the issue. (maybe not I am really not sure)

I guess we can look into in a another PR but since you are already working with and it might be pretty simple.

.circleci/config.yml
```diff
@@ -34,8 +23,37 @@ jobs:
...
       # run tests!
       - run: yarn eslint
       - run: yarn test-chrome-headless
+      - run:
+          name: Build the app so that server can be started
+          command: cd app && yarn build && cd ../
+      - run:
+          name: Start server with correct env for cypress tests
+          command: >
+            yarn generate-cert && ADALITE_ENABLE_HTTPS=true
```
refi93, line 38: I'd extract this into a shell script and reuse it in package.json, so it isn't duplicated and CI and local cypress:run remains aligned

xdzurman: It irked me to put cypress:dev here, since circleCI doesn't need dev environment, webpack watching, etc. I tried out some scripts, but it's not very cross-platform friendly, what would you propose? Stick to unix's "... | xargs"?

.circleci/config.yml
```diff
@@ -34,8 +23,37 @@ jobs:
...
+            ADALITE_SERVER_URL='https://localhost:3000' PORT=3000
+            ADALITE_BLOCKCHAIN_EXPLORER_URL='https://explorer-testnet.adalite.io'
```
refi93, line 40: I'd consider creating a custom .env.test file in the cypress folder and explicitly loading it into yarn dev. For example in my local setup some environment variables were set differently (ADALITE_TREZOR_CONNECT_URL) and cypress was crashing - took me a while to debug

A way to achieve it that I thing should work would be in `loadConfig.js` replacing

```javascript
require('dotenv').config()

```

with

```
require('dotenv').config({ path: process.env.DOTENV_CONFIG_PATH})
```

which then would allow for

`DOTENV_CONFIG_PATH=./my/path/to/.env.anything` where the proper env variables for cypress could be set, without any interference from local .env file

app/cypress/.env.cypress
```diff
@@ -0,0 +1,63 @@
...
+
+# donations
+ADALITE_FIXED_DONATION_VALUE = 40
+ADALITE_MIN_DONATION_VALUE = 1
+
+# IP blacklist as comma separated list
+ADALITE_IP_BLACKLIST = 8.8.8.8, 9.9.9.9
+
+#sentry
+SENTRY_DSN='https://d77d3bf9d9364597badab9c00fa59a31@sentry.io/1501383'
```
refi93, line 46: having the sentry dsn set for tests can mess up our error data

.circleci/config.yml
xdzurman: I'll create a separate issue.

app/cypress/.env.cypress
xdzurman: @PeterBenc it is checked via SENTRY_DSN: check.nonEmptyString, is this necessary?

refi93: I'm pretty sure it isn't - the sentry library itself should be graceful when no DSN is supplied and fall back to not doing anythin instead - you can try it yourself

refi93: https://docs.sentry.io/product/sentry-basics/dsn-explainer/

> If an SDK is not initialized or if it is initialized with an empty DSN, the SDK will not send any data over the network, such as captured exceptions.

README.md
```diff
@@ -85,6 +85,27 @@ yarn test
...
+yarn cypress:open
+```
+to start the tests with the interactive Test runner, which allows time travel, logs, pausing, etc.
+
+Or, to run the tests in a headless fashion, type
+```
+yarn cypress:run
+```
+
+Both commands create a video from the testing process and store it in /app/cypress/videos.
```
rizip1, line 107: Can we disable video creation based on some ENV mainly when running locally?

app/cypress/integration/advancedTab.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    cy.dataCy('Advanced sign button')
+      .should('be.visible')
+      .and('be.disabled')
+    cy.dataCy('Advanced download signature button')
+      .should('be.visible')
+      .and('be.disabled')
+
+    cy.dataCy('Download stake key')
+      .should('be.visible')
+      .and('not.have.attr', 'href', '#undefined')
```
rizip1, line 26: Please add not why is it checking for `#undefined`. Thanks

app/cypress/integration/sendingTab.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    cy.get('@SendButton').click()
+    cy.contains('Transaction review').should('be.visible')
+    cy.contains(sendAddress).should('be.visible')
+    // total is the sum of previously seen fee and input amount
+    cy.dataCy('Send amount total')
+      .should('be.visible')
+      .invoke('text')
+      .then(parseFloat)
+      .should(function($totalAmount) {
+        expect($totalAmount).to.eq(this.FeeAmount + parseFloat(sendAmount))
```
rizip1, line 38: I am just thinking if we can not get strange false-positive errors in some cases with checks like there. As e.g. `0.1 +  0.2 === 0.3` returns `false` in JS, due to JS arithmetics :thinking:

app/cypress/integration/stakingtab.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    cy.contains('Staking and Rewards History').should('be.visible')
+    cy.contains('Delegate Stake')
+      .as('DelegateAccordion')
+      .should('be.visible')
+
+    // expand delegation card if it is hidden
+    cy.dataCy('Delegate button')
+      .as('DelegateButton')
+      .then(($button) => {
+        // @ts-ignore: Is exists for a jquery button
```
rizip1, line 22: what is the role of `jquery` in here? :thinking:

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -236,6 +246,7 @@ const SendAdaPage = ({
       className={`input ${isModal ? '' : 'send-address'} fullwidth`}
       name="send-address"
       placeholder="Receiving address"
+      data-cy="Address text field"
```
rizip1, line 249: Would not it be in general better to avoid white spaces in the `data-cy` values and instead use `AddressTextField` style everywhere? It is easier to introduce two white spaces instead of one and then you are surprised why it does not work.

server/mocking.js
```diff
@@ -99,7 +99,7 @@ async function request(url, method = 'get', body = null, headers = {}) {
     body,
   })
   if (res.status >= 400) {
-    throw new Error(res.status)
+    // throw new Error(res.status) // TODO: mock on cypress level
```
rizip1, line 102: Could you please explain why this was changed?

rizip1 (commented):
Please see few minor notes.
Also from the 3 runs I did, once the test failed for me locally due to `AssertionError: Timed out retrying after 10000ms: expected '<button.button.primary.medium>' to be 'enabled'`. Maybe we can increase the timeout a little?
Anyway, good job having this done, as it will be of a great value, mainly through refactoring times :+1: .
IMHO let's finish/resolve the minor issues and merge ASAP so that we can start using & improving this.

app/cypress/integration/stakingtab.spec.ts
xdzurman: jquery is the only way to conditionally do stuff in cypress

server/mocking.js
xdzurman: Initially, the testnet backend returned 500 on pool recommendation because it did not have any recommended pool set up. This place on the frontend therefore threw, I set up the backend to not throw. I'll revert this

app/cypress/integration/sendingTab.spec.ts
xdzurman: True, the edge case it's testing shouldn't happen too

app/frontend/components/pages/sendAda/sendAdaPage.tsx
xdzurman: done

app/cypress/integration/advancedTab.spec.ts
xdzurman: I actually changed this to strongly check the content of the file

app/cypress/integration/sendingTab.spec.ts
xdzurman: Removed, now checking whether total amount is greater than send amount / min amount

rizip1 (approved):

refi93 (approved):

xdzurman:
closes #888

PR #913, opened 2021-03-25
Fix counting number of distinct assetNames

PeterBenc:
I realized that in case there are same assetNames under a different policyId, the function for calculating min ada for output would consider only one unique assetName, which is true but since they are under different policyId they cause the output to be bigger. For this reason we need to calculate unique number of (policyId, assetName) pairs intead.

### DIFF
--- a/app/frontend/wallet/shelley/shelley-transaction-planner.ts
+++ b/app/frontend/wallet/shelley/shelley-transaction-planner.ts
@@ -33,7 +33,6 @@ import {
   TxWithdrawal,
 } from '../types'
 import {aggregateTokenBundles, getTokenBundlesDifference} from '../helpers/tokenFormater'
-import distinct from '../../helpers/distinct'
 
 type TxPlanDraft = {
   outputs: TxOutput[]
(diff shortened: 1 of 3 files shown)


rizip1 (approved):

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -166,11 +165,14 @@ export const computeMinUTxOLovelaceAmount = (tokenBundle: TokenBundle): Lovelace
   // NOTE: should be 29 but a bug in Haskell set this to 27
   const adaOnlyUtxoSize = utxoEntrySizeWithoutVal + coinSize
 
-  const distinctAssets = distinct(tokenBundle.map(({assetName}) => assetName))
+  // this ensures there are only distinct (policyId, assetName) pairs
```
refi93, line 168: I would cover this with unit tests - it's a function without side-effects so an easy win in terms of testing, yet it has quite a non-trivial logic which is easy to break/introduce regressions as we modify it/implement further possible fixes

app/tests/src/tokenBundle.js
```diff
@@ -0,0 +1,110 @@
+import assert from 'assert'
```
refi93, line 1: This PR seems to be based on the one fixing the assets ordering, but this file is named tokenFormatter.js there: https://github.com/vacuumlabs/adalite/pull/922/files#diff-455d601662bff42e0cc51f668eab06af1897ae39c9330ba6ec1f3928119d231f - so make sure to not accdentally introduce a duplicate test when merging/rebasing. I like the tokenBundle.js name more if I had to pick one

refi93 (approved):

PR #915, opened 2021-03-28
Send and delegate connect refactor

rizip1:
* Based on the initial 2 commits this was expected to be easy refactor but:
* After the second commit, I got myself into the issue shown on the image below (after submit delegation). I only exchanged "connect" for "useActions", but when I started to investigate why the error happens, I just felt that this part strongly needs refactor, as it was very difficult for me to figure out the flow of information
issue

**Incompatibility changes**
* if the pool hash is invalid, the fee is not calculated, but I consider this to be expected.

=> divide actions part 1 PR should be merged first

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -653,7 +653,7 @@ export default (store: Store) => {
       setState({
         shelleyDelegation: {
           ...newState.shelleyDelegation,
-          delegationFee: txPlanResult.txPlan.fee + txPlanResult.txPlan.deposit,
+          delegationFee: (txPlanResult.txPlan.fee + txPlanResult.txPlan.deposit) as Lovelace,
         },
       })
       const delegationTransactionSummary: DelegateTransactionSummary = {
@@ -667,6 +667,7 @@ export default (store: Store) => {
         txSuccessTab: newState.txSuccessTab === 'send' ? newState.txSuccessTab : '',
       })
     } else {
+      // REFACTOR: (Untyped errors)
       const validationError =
         delegationPlanValidator(balance, 0 as Lovelace, txPlanResult.estimatedFee) ||
         txPlanResult.error
@@ -680,53 +681,17 @@ export default (store: Store) => {
 
   const debouncedCalculateDelegationFee = debounceEvent(calculateDelegationFee, 500)
 
-  const validateDelegationAndCalculateFee = () => {
-    const state = getState()
-    const selectedPool = state.shelleyDelegation.selectedPool
-    const delegationValidationError = selectedPool.validationError
-
-    setError(state, {
-      errorName: 'delegationValidationError',
-      error: delegationValidationError,
-    })
-    setState({
-      delegationValidationError,
-    })
-    if (!delegationValidationError) {
-      setState({
-        calculatingDelegationFee: true,
-        gettingPoolInfo: true,
-      })
-      debouncedCalculateDelegationFee(state)
-    } else {
-      setState({
-        shelleyDelegation: {
-          ...state.shelleyDelegation,
-          delegationFee: 0 as Lovelace,
-        },
-      })
-    }
-  }
-
-  const updateStakePoolIdentifier = (
-    state: State,
-    poolHash: string,
-    validationError: any = null
-  ): void => {
+  const updateStakePoolIdentifier = (state: State, poolHash: string): void => {
     const newPool = poolHash && state.validStakepoolDataProvider.getPoolInfoByPoolHash(poolHash)
-    const oldPool = state.shelleyDelegation.selectedPool
-    if (newPool && newPool?.poolHash === oldPool?.poolHash) return
     setState({
       shelleyDelegation: {
         ...state.shelleyDelegation,
-        selectedPool: {
-          ...newPool,
-          validationError,
-        },
+        selectedPool: newPool,
       },
+      calculatingDelegationFee: true,
+      gettingPoolInfo: true,
     })
-    if (validationError) return
-    validateDelegationAndCalculateFee()
+    debouncedCalculateDelegationFee(state)
   }
 
   const resetStakePoolIndentifier = (): void => {
@@ -734,14 +699,6 @@ export default (store: Store) => {
     setState({
       delegationValidationError: null,
     })
-  }
-
-  const selectAdaliteStakepool = (state: State): void => {
-    const newState = getState()
-    updateStakePoolIdentifier(
-      newState,
-      getSourceAccountInfo(newState).poolRecommendation.recommendedPoolHash
-    )
   }
 
   /* MULTIPLE ACCOUNTS */
@@ -859,7 +816,6 @@ export default (store: Store) => {
       shouldShowDelegationModal: true,
       txSuccessTab: '',
     })
-    selectAdaliteStakepool(state)
   }
 
   const closeDelegationModal = (state: State) => {
@@ -966,8 +922,6 @@ export default (store: Store) => {
         .getAccount(state.sourceAccountIndex)
         .generateNewSeeds()
       resetAccountIndexes(state)
-      resetDelegation()
-      selectAdaliteStakepool(state)
       setState({
         waitingForHwWallet: false,
         // TODO: refactor txSuccesTab
@@ -1059,7 +1013,6 @@ export default (store: Store) => {
     resetTransactionSummary(state)
     resetSendFormFields(state)
     resetDelegation()
-    if (mainTab === MainTabs.STAKING) selectAdaliteStakepool(state)
   }
 
   const closeUnexpectedErrorModal = (state) => {
@@ -1264,7 +1217,6 @@ export default (store: Store) => {
     updateStakePoolIdentifier,
     resetStakePoolIndentifier,
     setActiveMainTab,
-    selectAdaliteStakepool,
     convertNonStakingUtxos,
     loadErrorBannerContent,
     withdrawRewards,
--- a/app/frontend/components/pages/delegations/delegateInput.tsx
+++ b/app/frontend/components/pages/delegations/delegateInput.tsx
@@ -1,10 +1,5 @@
-import {Stakepool, PoolRecommendation} from '../../../types'
 import {Fragment, h} from 'preact'
-import {useEffect, useState} from 'preact/hooks'
-import {connect} from '../../../libs/unistore/preact'
-import actions from '../../../actions'
-import {getSourceAccountInfo, State} from '../../../state'
-import {StakepoolDataProvider} from '../../../../frontend/helpers/dataProviders/types'
+import {useSelector} from '../../../helpers/connect'
 import {ADALITE_CONFIG} from '../../../../frontend/config'
 
 type StakePoolLabelProps = {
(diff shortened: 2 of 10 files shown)


app/frontend/actions.ts
```diff
@@ -673,53 +674,19 @@ export default (store: Store) => {
...
-      })
-    }
-  }
-
-  const updateStakePoolIdentifier = (
-    state: State,
-    poolHash: string,
-    validationError: any = null
-  ): void => {
+  const updateStakePoolIdentifier = (state: State, poolHash: string): void => {
```
rizip1, line 677: I was just totally confused by this function & decided to remove it. @PeterBenc please double-check that it is handled correctly. :pray:

app/frontend/actions.ts
```diff
@@ -729,14 +696,6 @@ export default (store: Store) => {
     })
   }
 
-  const selectAdaliteStakepool = (state: State): void => {
```
rizip1, line 732: Was there any reason behind this? It seems to me that it is just fine to remove this.

app/frontend/components/pages/delegations/delegateInput.tsx
```diff
@@ -29,86 +24,18 @@ const StakePoolLabel = ({
...
-
-const DelegateInput = ({
-  poolRecommendation,
-  pool,
-  validStakepoolDataProvider,
-  updateStakePoolIdentifier,
-  resetStakePoolIndentifier,
-}: Props): h.JSX.Element => {
-  const [fieldValue, setFieldValue] = useState('')
-  const [isTicker, setIsTicker] = useState(false)
```
rizip1, line 84: Those states were redundant as the values can be inferred.

app/frontend/components/pages/delegations/delegateInput.tsx
```diff
@@ -29,86 +24,18 @@ const StakePoolLabel = ({
...
-}
-
-const DelegateInput = ({
-  poolRecommendation,
-  pool,
-  validStakepoolDataProvider,
-  updateStakePoolIdentifier,
-  resetStakePoolIndentifier,
-}: Props): h.JSX.Element => {
-  const [fieldValue, setFieldValue] = useState('')
```
rizip1, line 83: General recommendation. Do not get into flow like:
"child" -> "has local state and sync it with global" -> "parent" -> "read from global"

...just move the state to parent and avoid global state if not really necessary.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -54,48 +97,100 @@ const Delegate = ({withAccordion, title}: Props): h.JSX.Element => {
...
+  const handleOnInput = (event: any): void => {
+    const fieldValue: string = event?.target?.value
+    setFieldValue(fieldValue)
+
+    // Note: this could be debounced
+    if (fieldValue) {
+      const {poolHash, error} = validateInput(fieldValue, validStakepoolDataProvider)
+      if (!error) {
+        /*
+        Relates to REFACTOR (calculateFee):
```
rizip1, line 141: I think that it would be ideal if this component just calculated plan, rendered modal, pass data, ... and no need for global state.

rizip1:
@ppershing @PeterBenc If you are not sure about 3rd commit, I can try to postpone it, and just hack the infinite cycle using `useRef` or `eslint-disable-line` though I am curious what you think about the suggested changes.

app/frontend/components/pages/delegations/delegateInput.tsx
rizip1: In general "parent" (usualy the form screen) should be responsible for storing form state, not the form fields itself.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -46,58 +49,148 @@ const BigDelegatorOffer = (): h.JSX.Element => (
   </div>
 )
 
+// REFACTOR: (Untyped errors): move to types
+// is "hasTickerMapping" something specific or general?
```
rizip1, line 53: ?

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -46,58 +49,148 @@ const BigDelegatorOffer = (): h.JSX.Element => (
...
+    calculatingDelegationFee: state.calculatingDelegationFee,
+    delegationFee: state.shelleyDelegation.delegationFee,
+    txSuccessTab: state.txSuccessTab,
+    gettingPoolInfo: state.gettingPoolInfo,
+    isShelleyCompatible: state.isShelleyCompatible,
+    poolRecommendation: getSourceAccountInfo(state).poolRecommendation,
+    isBigDelegator: isBigDelegatorSelector(state),
+    validStakepoolDataProvider: state.validStakepoolDataProvider,
+  }))
+  const {confirmTransaction, updateStakePoolIdentifier} = useActions(actions)
```
rizip1, line 123: Seems like `useActions` returns new reference of `updateStakePoolIdentifier` :thinking:

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -46,58 +49,148 @@ const BigDelegatorOffer = (): h.JSX.Element => (
...
+
+  // TODO: this reference always changes inside "useEffect" which causes infite loop.
+  // Till this is investigated, this workaround is used instead of "eslint-disable-..."
+  const _updateStakePoolIdentifier = useRef(updateStakePoolIdentifier)
+
+  const [fieldValue, setFieldValue] = useState('')
+  const [error, setError] = useState<Error>(null)
```
rizip1, line 130: I suggest only keep this error in the local state and not somehow propagate it to the global state as no other component really needs it. Or does it?

app/frontend/types.ts
```diff
@@ -208,7 +208,6 @@ export type Stakepool = {
   ticker: string
   homepage: string
   poolHash: string
-  validationError?: any
```
rizip1, line 211: Such data should not be stored together with pool data, validation (if not stored just locally), should be under a separate key, not part of the "stakePool" data itself

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -54,48 +97,100 @@ const Delegate = ({withAccordion, title}: Props): h.JSX.Element => {
...
+    poolRecommendation,
+    validStakepoolDataProvider,
+  ])
 
   const delegationHandler = async (): Promise<void> => {
     await confirmTransaction('delegate')
   }
 
-  const validationError = !!delegationValidationError || !stakePool || !!stakePool.validationError
+  const validationError = !!delegationValidationError || !!error || !stakePool
```
ppershing, line 179: why is !stakePool only single negation? :thinking:

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -54,48 +97,100 @@ const Delegate = ({withAccordion, title}: Props): h.JSX.Element => {
...
+      poolRecommendation?.recommendedPoolHash || currentDelegation?.poolHash
 
-  /*
-  REFACTOR (calculateFee):
-  Calculation of "fee" should be fired from there explicitely using "useEffect"
-  when component mounts.
-  */
+    if (recommendedPoolHash) {
+      const {poolHash, error} = validateInput(recommendedPoolHash, validStakepoolDataProvider)
+      if (error) {
```
ppershing, line 161: shouldn't this just reuse action above?

app/frontend/components/pages/delegations/delegatePage.tsx
rizip1: I assume it should read as `if stake pool is evaluated as false` => `we consider it error situation`. If it was `!!stakePool` we would consider it an error if stake pool was defined. Though maybe we do not need such condition, but I did not want to change logic much in this refactor.

rizip1:
> @ppershing @PeterBenc If you are not sure about 3rd commit, I can try to postpone it, and just hack the infinite cycle using `useRef` or `eslint-disable-line` though I am curious what you think about the suggested changes.

... hmm but I guess I can not as it then results in the error screenshotted above.

app/frontend/components/pages/delegations/delegatePage.tsx
rizip1: 15 points for gryffindor :+1: .

rizip1: Please re-check in https://github.com/vacuumlabs/adalite/pull/915/commits/39f64ff8e16a53ba25e0c95567f94c0984ea3ca4

app/frontend/actions.ts
PeterBenc: I thinks its fine as well.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -46,58 +49,145 @@ const BigDelegatorOffer = (): h.JSX.Element => (
...
+    // Note: this could be debounced
+    if (newValue) {
+      handleInputValidation(newValue)
+    }
+  }
+
+  // init "stake pool input" and refresh it when "currentDelegation" changes
+  useEffect(() => {
+    const recommendedPoolHash =
+      poolRecommendation?.recommendedPoolHash || currentDelegation?.poolHash
```
PeterBenc, line 164: Do we need the currentDelegation here just so the effect is triggered when its changed? Because otherwise it doesn't make a lot of sense to me to set `recommnededPoolHash` to currentDelegation.poolHash

PeterBenc (commented):
First of all thnx for refactoring this, I had no idea otherwise I would warn you :smile: 
I think the case for input being empty is not handled correctly (poolInfo should be empty) also fee should be set to zero in case of an error or empty field (also probably transaction summary should be reset). I understand that this is just a refactor so we probably don't need to address these in this PR ( but we shouldn't postpone this much either) 
image

app/frontend/components/pages/delegations/delegatePage.tsx
PeterBenc: I dont think it does. :+1:

app/frontend/components/pages/delegations/delegatePage.tsx
PeterBenc: Something specific, we implemented a "search by ticker" a couple of months ago. We receive pool info from backend but in some cases we don't receive the tickers. (not much we can do about that). In that case, if a user entered a ticker, we would like to show him a specific error that `search by ticker is temporarily disabled` rather than something general (and misleading) like `Invalid stakepool identifier)`

app/frontend/actions.ts
PeterBenc: I am not suprised, Previously there were multiple stakepool inputs for multiple pools (also reason for `validationError` in being in poolInfo object) and this is just a left over.

app/frontend/components/pages/delegations/delegatePage.tsx
rizip1: Yes I need it there because of that and this `poolRecommendation?.recommendedPoolHash || currentDelegation?.poolHash` condition was there before so I did not want to change it. Should I change it just to `poolRecommendation?.recommendedPoolHash`? I can then still keep `currentDelegation` in the dependencies of `useEffect`.

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -21,6 +23,7 @@ type DelegationValidationProps = {
   txSuccessTab: string
 }
 
+// REFACTOR: "txSuccessTab", what about using "success notification" instead?
```
xdzurman, line 26: First, it should be retyped into an enum, currently it's a freestyle string. The name sounds weird but after I checked what its responsibility is, the name quite fits. However, I think it's fundamentally flawed to have such a state variable and it should be changed (it just says what tab most recently sent a successful tx). Currently, if you transfer the funds between two accounts in the account tab, a success message is shown in the Sending tab, which is kinda weird

app/frontend/components/pages/delegations/delegatePage.tsx
```diff
@@ -46,58 +49,145 @@ const BigDelegatorOffer = (): h.JSX.Element => (
...
+  validStakepoolDataProvider: StakepoolDataProvider
+): ValidatedInput => {
+  if (ADALITE_CONFIG.ADALITE_ENABLE_SEARCH_BY_TICKER) {
+    const pool =
+      validStakepoolDataProvider.getPoolInfoByPoolHash(fieldValue) ||
+      validStakepoolDataProvider.getPoolInfoByTicker(fieldValue)
+    if (pool) return {poolHash: pool.poolHash, error: null}
+
+    const hasTickerMapping = validStakepoolDataProvider.hasTickerMapping
+    const isTickerString = fieldValue.length <= 5 && fieldValue.toUpperCase() === fieldValue
```
xdzurman, line 75: I don't think `fieldValue.toUpperCase() === fieldValue` is necessary, there can be [lower case tickers](https://pooltool.io/pool/76619d4c31c0570824e369d13a95542b65ee29546f2527f96042e602/epochs) as well, cc @PeterBenc

rizip1:
@PeterBenc please check last commit. Also please check this `I don't think fieldValue.toUpperCase() === fieldValue is necessary, there can be lower case tickers as well, cc` comment, whether I should change that or leave as it is. Thanks

app/frontend/components/pages/delegations/delegatePage.tsx
PeterBenc: Agreed we should remove the `field.toUpporCase()` condition :+1:

app/frontend/components/pages/delegations/stakePoolInfo.tsx
```diff
@@ -4,12 +4,26 @@ import {Stakepool, Lovelace} from '../../../types'
...
+  validationError: Error
+  empty: boolean
 }
 
-export const StakePoolInfo = ({pool, gettingPoolInfo}: StakePoolInfoProps): h.JSX.Element => {
+export const StakePoolInfo = ({
+  pool,
+  gettingPoolInfo,
+  validationError,
+  empty,
```
PeterBenc, line 25: It seems to me we don't need this parameter, passing pool as null should imho solve the issue and also keep the current behaviour.

PeterBenc (commented):
Couple of problems I encoutered

This happens after deregistering staking key.
image



This happens after delegation
image

Also I noticed that after copy-pasting valid stakepool Id, Error appears for a brief moment and then `getting pool info` triggers.

app/frontend/components/pages/delegations/delegatePage.tsx
PeterBenc: Lets keep it as it is, we should address the pool recommendation in another PR.

rizip1:
@PeterBenc I could not reproduce any of the issues you mentioned :thinking: did you tested it locally using the fresh and up to date commits?

rizip1:
> @PeterBenc I could not reproduce any of the issues you mentioned  did you tested it locally using the fresh and up to date commits?

Could replicate, will check that.

PeterBenc (approved):

PR #917, opened 2021-03-29
Transactions refactor part 1

rizip1:
This one is aimed to fix race-conditions when calculating multiple TxPlans at once. Now we should have one TxPlan per TxType and it should be determined by `txConfirmType` when confirming the transaction.
We discussed that we do not expect multiple TxPlans for on TxType so I did not consider this scenario.

Among that, I tried to make chosen action parameters more explicit. Also, I postfixed parts that I did not refactor with **"Old"**, what do you think about that?

The refactor of "connect" related to "sendAdaPage" and "delegate" is handled in a https://github.com/vacuumlabs/adalite/pull/915 so I did not incorporate it here, though it would help to ensure better types.

Ideally I would prefer to drop this "txPlan" info from global state altogether, but this would require a lot extra changes, so maybe this could be changed in the future. In ideal scenarios I would prefer style like this:

// e.g. sendAdaPage (very simplified)
// but idea is that we do not really need global state, that is more difficult to reason about (race conditions, reset in right time, etc) for that
```
const [txPlan, setTxPlan] = useState(...)
const [formData, setFormData] = useState(...)
// good router would help here
const [showSubmit, setShowSubmit] = useState(...)

useEffect(() => {
   setTxPlan(calculateTxPlan(formData))  
}, [reasonable deps])

const onSubmit = () => {
  showSubmit(true)
}

// and "Confirm" handled similarly

return (
  {!showSubmit && (
    <Form {...formData} />
    <button label="Send ADA" />
  )}
  {showSubmit && <SubmitTransactionModal {...{txPlan, formData}} />}
)

```

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -33,6 +33,7 @@ import {
   WithdrawTransactionSummary,
   DelegateTransactionSummary,
   DeregisterStakingKeyTransactionSummary,
+  TransactionSummary,
 } from './types'
 import {MainTabs} from './constants'
 import {parseCliUnsignedTx} from './wallet/shelley/helpers/stakepoolRegistrationUtils'
@@ -164,20 +165,26 @@ export default (store: Store) => {
 
   /* TRANSACTION */
 
-  const confirmTransaction = async (state: State, txConfirmType): Promise<void> => {
+  const _confirmTransaction = async (
+    state: State,
+    {
+      txConfirmType,
+      txPlan,
+      sourceAccountIndex,
+    }: {txConfirmType: TxType; txPlan: TxPlan; sourceAccountIndex: number}
+  ): Promise<void> => {
     let txAux
-    const newState = getState()
     try {
-      if (newState.sendTransactionSummary.plan) {
+      if (txPlan) {
         txAux = await getWallet()
-          .getAccount(state.sourceAccountIndex)
-          .prepareTxAux(newState.sendTransactionSummary.plan)
+          .getAccount(sourceAccountIndex)
+          .prepareTxAux(txPlan)
       } else {
         loadingAction(state, 'Preparing transaction plan...')
         await sleep(1000) // wait for plan to be set in case of unfortunate timing
         const retriedState = getState()
         txAux = await getWallet()
-          .getAccount(state.sourceAccountIndex)
+          .getAccount(sourceAccountIndex)
           .prepareTxAux(retriedState.sendTransactionSummary.plan)
       }
     } catch (e) {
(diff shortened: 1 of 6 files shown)


app/frontend/actions.ts
```diff
@@ -164,20 +165,26 @@ export default (store: Store) => {
...
+      txConfirmType,
+      txPlan,
+      sourceAccountIndex,
+    }: {txConfirmType: TxType; txPlan: TxPlan; sourceAccountIndex: number}
+  ): Promise<void> => {
     let txAux
-    const newState = getState()
     try {
-      if (newState.sendTransactionSummary.plan) {
+      if (txPlan) {
```
rizip1, line 178: FYI: basically what was changed is that the `state.something` was exchanged for explicit params

app/frontend/actions.ts
```diff
@@ -186,26 +193,50 @@ export default (store: Store) => {
...
-      txConfirmType === 'convert' ||
-      txConfirmType === 'withdraw' ||
-      txConfirmType === 'deregisterStakeKey'
+      txConfirmType === TxType.CONVERT_LEGACY ||
+      txConfirmType === TxType.WITHDRAW ||
+      txConfirmType === TxType.DEREGISTER_STAKE_KEY
 
     setState({
       shouldShowConfirmTransactionDialog: true,
-      txConfirmType: isTxBetweenAccounts ? 'crossAccount' : txConfirmType,
```
rizip1, line 201: I believe that "crossAccount" should not be `txConfirmType` so I moved it to a separate property that I would also ideally remove later on, and only make it computed

app/frontend/actions.ts
```diff
@@ -894,34 +955,43 @@ export default (store: Store) => {
...
+    }: {txSummary: TransactionSummary; sourceAccountIndex: number; sendAddress: string}
+  ) => {
+    const {usingHwWallet, keepConfirmationDialogOpen, hwWalletName} = state
+
     setState({
       shouldShowSendTransactionModal: false,
       shouldShowDelegationModal: false,
     })
-    if (!state.keepConfirmationDialogOpen) {
+    if (!keepConfirmationDialogOpen) {
```
rizip1, line 972: Also here the `state.something` was just moved to params

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
```diff
@@ -254,65 +254,93 @@ const ConvertFundsReview = ({
...
+        sourceAccountIndex,
+        txSummary: cachedTransactionSummaries[TxType.DEREGISTER_STAKE_KEY],
+      })
+    }
+    // to be refactored cases
+    return submitTransaction({sendAddress, sourceAccountIndex, txSummary: transactionSummary})
+  }
+
+  // Refactor: tmp util till all transactions types use "cachedTransactionSummaries"
+  const getModalBody = () => {
```
rizip1, line 301: I did not want to "refactor" all, as it would be a nightmare to review to I decided to have it split like this.

app/frontend/state.ts
```diff
@@ -80,11 +84,24 @@ export interface State {
 
   // transaction
   sendTransactionSummary: TransactionSummary
+
+  // Note: This allows for multiple transaction summaries by TxType, to avoid race-conditions
+  // when calculating transaction summaries for more TxTypes at the same time.
+  // Note that this still does not allow for multiple cached transaction summaries from
+  // the same TxType, however, this should not be use-case for us in the near future.
+  cachedTransactionSummaries: {
+    [TxType.CONVERT_LEGACY]?: TransactionSummary & SendTransactionSummary
```
rizip1, line 93: Any ideas on how to make this maybe "better" but not overcomplicate it? Or do you consider this fine?

app/frontend/components/pages/delegations/shelleyBalances.tsx
```diff
@@ -118,11 +112,7 @@ const shelleyBalances = ({
                 : `${printAda(nonStakingBalance)}`}
               <AdaIcon />
             </div>
-            <button
-              disabled={calculatingDelegationFee}
-              className="button secondary convert"
-              onClick={convertNonStakingUtxos}
-            />
+            <button className="button secondary convert" onClick={convertNonStakingUtxos} />
```
rizip1, line 115: IMPORTANT: I will revert this part as "delegate" is not yet using "cachedTxPlans" but the old approach, so those checks must stay here. They can only be removed for "DeregisterStakeKey"

rizip1: But it should be fine to review the rest and the concept in general.

app/frontend/components/pages/delegations/deregisterStakeKey.tsx
```diff
@@ -21,11 +15,7 @@ const DeregisterStakeKeyPage = () => {
       <p className="deregister-stake-key-card-disclaimer">
         ...if you do not want to use this wallet anymore
       </p>
-      <button
-        className="button secondary cancel-delegation"
-        disabled={calculatingDelegationFee}
-        onClick={() => deregisterStakingKey()}
-      >
+      <button className="button secondary cancel-delegation" onClick={() => deregisterStakingKey()}>
```
rizip1, line 18: If we use the new approach for multiple transactions we can remove such checks from more places.

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
xdzurman: Now it doesn't make sense, but I'd later opt for handling all txs types here (now, TxType.POOL_REG_OWNER is handled separately) and when that's done, I would take advantage of typescript's exhaustive switches and our "assertUnreachable". If you agree, you can comment it down or I'll create an issue. 

A switch would look nicer than "ifs" even now imho

rizip1:
Lets postpone till https://github.com/vacuumlabs/adalite/pull/936 is merged.

xdzurman (changes requested):
Can you please rebase so that new changes and cypress are in place?

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
rizip1: Actually, we would probably opt for having multiple smaller and specific dialogs as discussed so I am not sure whether comment it out inside this PR. But I can at least change it to `switch`.

app/frontend/state.ts
rizip1: I would prefer another PR, to close this quicker. I will create an issue for that.

rizip1:
@xdzurman @PeterBenc please check https://github.com/vacuumlabs/adalite/pull/953 it is duplicate of this one, but due to changes in `actions` it was easier to create new branch and duplicate this one, than rebase.

PR #925, opened 2021-03-31
Asset fingerprint

PeterBenc:
Adds asset fingerprint
As we discussed, for some time we want to show also the policy id, we ll remove it when trezor releases.

### DIFF
--- a/app/frontend/components/pages/txHistory/transactionHistory.tsx
+++ b/app/frontend/components/pages/txHistory/transactionHistory.tsx
@@ -1,5 +1,8 @@
 import {h, Fragment} from 'preact'
-import {assetNameHex2Readable} from '../../../wallet/shelley/helpers/addresses'
+import {
+  assetNameHex2Readable,
+  encodeAssetFingerprint,
+} from '../../../wallet/shelley/helpers/addresses'
 import printAda from '../../../helpers/printAda'
 import {getActiveAccountInfo, State} from '../../../state'
 import actions from '../../../actions'
@@ -17,6 +20,7 @@ import {
 } from '../../../types'
 import {AdaIcon, StarIcon} from '../../common/svg'
 import {LinkToAsset} from '../../common/asset'
+import {StringEllipsis} from '../stringEllipsis'
 import moment = require('moment')
 
 const FormattedAmount = ({amount}: {amount: Lovelace}): h.JSX.Element => {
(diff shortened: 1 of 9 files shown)


app/frontend/components/pages/txHistory/transactionHistory.tsx
```diff
@@ -102,6 +106,10 @@ const MultiAsset = ({star, name, hash, amount}: MultiAssetProps) => (
       <span className="ellipsis">{hash.slice(0, -6)}</span>
       {hash.slice(-6)}
     </div>
+    <div className="multi-asset-hash">
+      <span className="ellipsis">{fingerprint.slice(0, -6)}</span>
+      {fingerprint.slice(-6)}
+    </div>
```
xdzurman, line 112: These ellipses and slicing are getting repetitive, what would you say to a helper returning 
```
<span className="ellipsis">{whatever.slice(0, -6)}</span>{whatever.slice(-6)}
```
?

app/tests/src/tokenBundle.js
```diff
@@ -86,25 +87,103 @@ describe('Min ada calculation', () => {
...
+describe('Asset fingerprint encoding', () => {
+  it('should encode correct fingerprints for various policies and assetNames', () => {
+    const testFingerprints = assetFingerprintTestVectors.map(
+      ({assetfingerprint}) => assetfingerprint
+    )
+    const resultFingerprints = assetFingerprintTestVectors.map(({policyid, assetname}) =>
+      encodeAssetFingerprint(policyid, assetname)
+    )
+    assert.deepEqual(testFingerprints, resultFingerprints)
+  })
```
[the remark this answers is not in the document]
xdzurman: I think it's OK to do it at once, you'll still see which ones are not deep equal in the array.

app/frontend/components/pages/txHistory/transactionHistory.tsx
rizip1: Agree :+1:

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -29,6 +30,15 @@ export const encodeAddress = (address: Buffer): Address => {
   return bech32.encode(addressPrefix, address)
 }
 
+// TODO: we might want to add this to cardano-crypto.js
+export const encodeAssetFingerprint = (policyIdHex: HexString, assetNameHex: HexString): string => {
```
rizip1, line 34: Why is it located in the `address` folder? Is fingerprint some kind of address for tokens or what is actually its purpose? :thinking:

rizip1 (approved):

app/frontend/wallet/shelley/helpers/addresses.ts
PeterBenc: Good point, "addresses" is pretty outdated name, you are right to we should divide the file into multiple helper files since it already contains non-address related stuff such as `xpub2ChainCode`.  I ll create an issue for it.

PR #928, opened 2021-04-01
Remove shouldShowExportOption from global state

xdzurman:
A mini PR which removes shouldShowExportOption. I'd continue with other keys but then I remembered I was meant to refactor just the "form" global state

### DIFF
--- a/app/frontend/actions.ts
+++ b/app/frontend/actions.ts
@@ -64,12 +64,8 @@ export default (store: Store) => {
   } = walletActions(store)
   const {setState, getState} = store
 
-  const setAuthMethod = (state: State, option: AuthMethodType): void => {
-    setState({
-      authMethod: option,
-      shouldShowExportOption:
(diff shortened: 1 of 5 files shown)


app/frontend/actions.ts
```diff
@@ -58,11 +58,7 @@ export default (store: Store) => {
   const {setState, getState} = store
 
   const setAuthMethod = (state: State, option: AuthMethodType): void => {
-    setState({
-      authMethod: option,
-      shouldShowExportOption:
-        option === AuthMethodType.MNEMONIC || option === AuthMethodType.KEY_FILE,
-    })
+    setState({authMethod: option})
```
PeterBenc, line 61: I would rename the `option` argument to `authMethod` but I guess that's a bit unrelated to this refactor....

PeterBenc (approved):
LGTM thnx

rizip1 (approved):

PR #936, opened 2021-04-07
Split actions into multiple files

DavidTranDucVL:
- Moved all functions from actions.ts to separate files.
- I noticed some actions that could have been refactored, but in the end I chose to only move existing code and don't change or refactor any logic. That will make easier to resolve conflicts and merge this PR with other PRs that are currently open.


app/frontend/actions/send.ts
```diff
@@ -0,0 +1,307 @@
...
+    } catch (e) {
+      setState({
+        calculatingFee: false,
+      })
+      setError(state, {errorName: 'sendAmountValidationError', error: {code: e.name}})
+      return
+    }
+  }
+
+  const convertNonStakingUtxos = async (state: State): Promise<void> => {
```
rizip1, line 216: I am not sure whether this is the best place for this and `witdrawRewards` action, but I do not better alternatives right now. Someone else any ideas?

app/frontend/actions/common.ts
```diff
@@ -0,0 +1,108 @@
...
+        coins: 0 as Lovelace,
+        token: null,
+        minimalLovelaceAmount: 0 as Lovelace,
+        fee: 0 as Lovelace,
+        plan: null,
+      },
+    })
+  }
+
+  const resetSendFormState = (state: State) => {
```
rizip1, line 33: If I see correctly, this is only used in `submitTransaction` so it could just be part of `action/transactions` and would not have to be exported.

app/frontend/actions.ts
```diff
@@ -58,6 +57,7 @@ export default (store: Store) => {
     resetPoolRegTransactionSummary,
     deregisterStakingKey,
   } = poolOwnerActions(store)
+  const {exportJsonWallet} = exportWalletActions(store)
```
rizip1: I think that we can move it into `wallet` file, it seems odd to me to have just single action in `exportWallet` when we already have `wallet` file.

DavidTranDucVL: Fixed

app/frontend/actions/common.ts
DavidTranDucVL: Fixed

app/frontend/actions/send.ts
DavidTranDucVL: I moved it to `actions/transaction.ts` for now.

rizip1 (approved):

PR #948, opened 2021-04-09
Refactor state 3

xdzurman:
- Refactor wallet names to an enum
 - Refactor usingHwWallet state var into a selector
 - Remove useless keys from State (isDelegationValid, currentDelegation)

### DIFF
--- a/app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
+++ b/app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
@@ -22,7 +22,7 @@ import {
   HexString,
   Address,
 } from '../../types'
-import {Network, TxByronWitness, TxShelleyWitness} from '../types'
+import {Network, TxByronWitness, TxShelleyWitness, WalletName} from '../types'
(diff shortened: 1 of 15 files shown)


app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -42,7 +42,7 @@ CryptoProviderParams): Promise<CryptoProvider> => {
 
   const isHwWallet = () => false
 
-  const getWalletName = () => 'Mnemonic'
+  const getWalletName = (): WalletName => WalletName.MNEMONIC
```
rizip1, line 45: Btw is not the return type `WalletName.MNEMONIC`?
The same question for the usage below.
I am just curious if it is not good to be explicit it the return type itself is explicit.
What do you think?

rizip1 (approved):
See question, but otherwise looks good to me.

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
xdzurman: Fixed. Forced it into a commit where it belonged.

PR #949, opened 2021-04-09
Cache backend for cypress and lower sleep time for tx submit mock

PeterBenc:
- Implements rather a very simple caching of request for cypress,
(open to suggestions on the hash function or the approach)
the alternative would be to create create a separate command that runs the tests and creates fixtures for the tests, tho this would require a bit more work and I doubt it would speed the testing process up.
- Lowers the sleep time when mocking requests. 

Improved integration tests duration, previously ~4:30 now ~3:00 

Gotcha: 
- let me know if the previous sleep durations had some special purpose other than mimicking the real response time, if we still consider this useful I can create a .env variable on which would the sleep time depend, and set it to 0 for cypres. IMO we dont need the mimicking of real response time anywhere.

### DIFF
--- a/app/cypress/support/index.ts
+++ b/app/cypress/support/index.ts
@@ -17,11 +17,37 @@
 // Import commands.js using ES2015 syntax:
 import './commands'
 
+
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
 
+/*
+ Source: https://gist.github.com/iperelivskiy/4110988
+*/
+const getHash = function(s: string) {
+  let h = 0xdeadbeef
+  for (let i = 0; i < s.length; i++) {
+    h = Math.imul(h ^ s.charCodeAt(i), 2654435761)
+  }
+  return ((h ^ (h >>> 16)) >>> 0).toString(16)
+}
+
+const cachedBackendResponses = {}
+
 // run before each test in each file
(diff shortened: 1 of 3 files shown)


app/cypress/support/index.ts
```diff
@@ -17,11 +17,27 @@
 // Import commands.js using ES2015 syntax:
 import './commands'
 
+
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
 
+const cachedBackendResponses = {}
+const hashCode = (s) => s.split('').reduce((a, b) => {a = ((a << 5) - a) + b.charCodeAt(0); return a & a}, 0)
```
xdzurman, line 25: a source of this would be nice

app/cypress/support/index.ts
```diff
@@ -17,11 +17,27 @@
...
+
 // run before each test in each file
 beforeEach('Logs in via UI', () => {
+  cy.intercept({url: /^https:\/\/localhost:3000\/api/}, (req) => {
```
xdzurman, line 29: this url seems kinda restrictive if we wanted to change anything. I think a substring would be sufficient

xdzurman (approved):
Perhaps test a regular dev environment with mocking modes on, whether it created any race conditions or something, otherwise LGTM

PR #950, opened 2021-04-09
Connect to hooks 1

xdzurman:
This PR converts connect - function/class components to function components with useSelector + useActions. The main focus is on components which are not that likely to be touched by anyone during other refactors.


rizip1 (approved):
LGTM. Make sure that the screens work as before & IMO good to merge.

PR #951, opened 2021-04-09
Refactor tx plan computation into recursive

PeterBenc:
Previously:
- tx plan computation was in a really ugly for cycle
- in case wallet has for example a two utxos 2 ADA and 5 ADA, there was a 50% chance that the 2 ADA utxo gets to the tx plan computation first and if user wanted to send 1.1 ADA, 0.9 would be added to the fee (since its too small for change output) 
- if a wallet didn't have registered staking key, and had less then amount 2ADA balance, no error was shown

Changes: 
 - tx plan computation refactored into recursion
 - we try to add the change to the fee only if the plan computation with all utxos failed
 - error is shown didn't have registered staking key, and had less then amount 2ADA balance

### DIFF
--- a/app/frontend/wallet/shelley/shelley-transaction-planner.ts
+++ b/app/frontend/wallet/shelley/shelley-transaction-planner.ts
@@ -54,8 +54,10 @@ export type TxPlanResult =
       success: false
       error: any
       estimatedFee: Lovelace
+      deposit: Lovelace
       minimalLovelaceAmount: Lovelace
     }
+
 export interface TxPlan {
   inputs: Array<TxInput>
   outputs: Array<TxOutput>
@@ -64,6 +66,7 @@ export interface TxPlan {
   deposit: Lovelace
   additionalLovelaceAmount: Lovelace
   fee: Lovelace
+  baseFee: Lovelace
   withdrawals: Array<TxWithdrawal>
 }
 
@@ -205,12 +208,12 @@ export function computeTxPlan(
   possibleChange: TxOutput,
   certificates: Array<TxCertificate>,
   withdrawals: Array<TxWithdrawal>
-): TxPlan {
+): TxPlanResult {
   const totalRewards = withdrawals.reduce((acc, {rewards}) => acc + rewards, 0)
   const totalInput = inputs.reduce((acc, input) => acc + input.coins, 0) + totalRewards
   const totalInputTokens = aggregateTokenBundles(inputs.map(({tokenBundle}) => tokenBundle))
   const deposit = computeRequiredDeposit(certificates)
-  const totalOutput = outputs.reduce((acc, {coins}) => acc + coins, 0) + deposit + totalRewards
+  const totalOutput = outputs.reduce((acc, {coins}) => acc + coins, 0) + deposit
   const totalOutputTokens = aggregateTokenBundles(outputs.map(({tokenBundle}) => tokenBundle))
 
   // total amount of lovelace that had to be added to token-containing outputs
@@ -222,34 +225,43 @@ export function computeTxPlan(
   const tokenDifference = getTokenBundlesDifference(totalInputTokens, totalOutputTokens)
   // Cannot construct transaction plan, not enought tokens
   if (tokenDifference.some(({quantity}) => quantity < 0)) {
-    throw new UnexpectedError(UnexpectedErrorReason.CannotConstructTxPlan)
-  }
-
-  if (
-    totalOutput > Number.MAX_SAFE_INTEGER ||
-    totalOutputTokens.some(({quantity}) => quantity > Number.MAX_SAFE_INTEGER)
-  ) {
-    throw new InternalError(InternalErrorReason.CoinAmountError)
+    return {
+      success: false,
+      minimalLovelaceAmount: additionalLovelaceAmount,
+      estimatedFee: 0 as Lovelace,
+      deposit,
+      error: {code: UnexpectedErrorReason.CannotConstructTxPlan},
+    }
   }
 
   const feeWithoutChange = computeRequiredTxFee(inputs, outputs, certificates, withdrawals)
 
-  // Cannot construct transaction plantotalOutput
-  if (totalOutput + feeWithoutChange > totalInput) {
-    throw new UnexpectedError(UnexpectedErrorReason.CannotConstructTxPlan)
+  // Cannot construct transaction plan
+  if (inputs.length === 0 || totalOutput + feeWithoutChange > totalInput) {
+    return {
+      success: false,
+      minimalLovelaceAmount: additionalLovelaceAmount,
+      estimatedFee: feeWithoutChange,
+      deposit,
+      error: {code: UnexpectedErrorReason.CannotConstructTxPlan},
+    }
   }
 
   // No change necessary, perfect fit
   if (totalOutput + feeWithoutChange === totalInput && tokenDifference.length === 0) {
     return {
-      inputs,
-      outputs,
-      change: null,
-      certificates,
-      deposit,
-      additionalLovelaceAmount,
-      fee: feeWithoutChange,
-      withdrawals,
+      success: true,
+      txPlan: {
+        inputs,
+        outputs,
+        change: null,
+        certificates,
+        deposit,
+        additionalLovelaceAmount,
+        fee: feeWithoutChange,
+        baseFee: feeWithoutChange,
+        withdrawals,
+      },
     }
   }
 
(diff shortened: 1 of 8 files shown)


app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -284,59 +305,103 @@ export function computeTxPlan(
...
+    },
   }
+}
 
-  // we cant build the transaction with big enough change lovelace
-  if (change.tokenBundle.length > 0 && change.coins < minimalChangeLovelace) {
-    throw new InternalError(InternalErrorReason.ChangeOutputTooSmall)
-  }
+// TODO: remove this altogether, since shelley all utxos are profitable
+export const isUtxoProfitable = (utxo: UTxO): boolean => {
```
refi93, line 344: can't we simply remove this?

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -452,41 +516,24 @@ export const selectMinimalTxPlan = (
...
-            (acc, output) =>
-              output.tokenBundle.length > 0
-                ? acc + computeMinUTxOLovelaceAmount(output.tokenBundle)
-                : 0,
-            0
-          ) as Lovelace,
-        }
-      }
+  const _selectMinimalTxPlan = (
+    numOfInputs: number,
```
refi93, line 520: I'd name this just `numInputs`

app/frontend/wallet/shelley/shelley-transaction-planner.ts
refi93: although technically, there can still be legacy inputs that are not profitable and it would be sensible to filter them out

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -284,59 +305,103 @@ export function computeTxPlan(
...
     return {
+      success: true,
+      txPlan: {
+        inputs,
+        outputs,
+        change: null,
+        certificates,
+        deposit,
+        additionalLovelaceAmount,
+        fee: (feeWithChange + change.coins) as Lovelace,
```
refi93, line 322: shouldn't this be "fee without change"?

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -452,41 +516,24 @@ export const selectMinimalTxPlan = (
...
+    allowAddingChangeToFee: boolean
+  ): TxPlanResult => {
+    const inputs: TxInput[] = utxos.slice(0, numOfInputs)
+    const txPlanResult = validateTxPlan(
+      computeTxPlan(inputs, outputs, change, certificates, withdrawals, allowAddingChangeToFee)
+    )
+    if (txPlanResult.success === true) {
+      return txPlanResult
     }
+    return numOfInputs < utxos.length
```
refi93, line 530: do we really need recursion? wouldn't it work as well with an ordinary while loop which is easier to follow?

Regarding the `allowAddingChangeToFee` I'd rather enhance the output of `computeTxPlan` to be able to tell how the fee was obtained, whether any change output has been "burned" - for that I'd consider just adding a `baseFee` property telling the fee attributed to the transaction as such (i.e. based on its base size) and that could be used to infer whether an output has been burned and hence whether it's worth trying to add additional inputs or not

refi93 (changes requested):
see comments - I'm not completely bought into why we need recursion and I'd rather avoid it for the sake of readability/maintainability

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: It may seem from the diff that I somehow touched this code, but I haven't.. :smile: but yes lets remove it. Regarding the legacy utxos, yes there still can be legacy utxos that are not profitable but we agreed back in August last year that we are gonna use these utxos anyway since we are converting everything to staking balance. I think a user don't really want having a `Non-staking Balance` of 0.00001 ada.  Anyway Im gonna remove it if you don't have any objections.

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: I agree, I'm not completely bought into using recursion as well, I can't say I like it very much :smile: but this is rather a simple one (basically just passing an incrementing counter) so I considered it fine. 

I m gonna try the approach you proposed and if It works I am all for it. :smile:

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: Thnx, it seems it should :+1:

PeterBenc: Interesting, I couldnt wrap my head around how it previously worked if it is wrong, but it turned out to be correct.
`(feeWithoutChange + (change.coins + (feeWithChange - feeWithoutChange)))` which is basically same as `feeWithChange + change.coins`
We can discuss this in slack but maybe you are already getting that the change amount is firstly calculated like
`totalInput - totalOutput - feeWithChange` so for this special condition we need to add the difference between `feeWithoutChange` and `feeWithChange` to the change.

PeterBenc (commented):
I fixed what you suggested, tho I need to look how is calculated fee in case there are rewards and we add change output to fee. I putting this on hold for now as I have to concentrate on other project for couple of days. I ll ping you when ist ready

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -286,57 +309,94 @@ export function computeTxPlan(
   // if we cannot create a change output with minimal ada we add it to the fee
   if (change.tokenBundle.length === 0 && change.coins < minimalChangeLovelace) {
```
refi93, line 310: will this work for multiasset though? You can "burn" the ADA as fees, but that's not the case with tokens :thinking:

refi93, line 310: I see no logic for failure in case the change output contains tokens and the ADA is not enough to cover the minimum ADA per output - how is that covered?

app/frontend/wallet/shelley/shelley-transaction-planner.ts
refi93: I see, basically you need to discard the change output ADA towards the fee that was computed for the transaction containing that output - rather than putting a long expressions that is obviously simplifiable I'd put a comment explaining that - that's more valuable for somebody trying to understand what's going on

refi93:
> I fixed what you suggested, tho I need to look how is calculated fee in case there are rewards and we add change output to fee. I putting this on hold for now as I have to concentrate on other project for couple of days. I ll ping you when ist ready

I see - that actually crossed my mind as well and conceptually, any withdrawals should count towards "totalInputs" when computing the fee

app/frontend/wallet/shelley/shelley-transaction-planner.ts
PeterBenc: Check out the `validateTxPlan` function, it checks every output for its minimal Lovelace amount. Note that I intentionally moved all validations to the `validateTxPlan` function and left the `computeTxPlan` only the responsibility of computing a balanced plan.

refi93:
If possible, I'd add a test case covering the edge-case outlined in the PR description, other than that, LGTM

refi93 (changes requested):
If possible, I'd add a unit test covering the edge-case outlined in the PR description (as an assurance that it works correctly, otherwise hard to test manually, and to prevent possible regressions), other than that, LGTM

app/tests/src/transaction-planner.js
```diff
@@ -0,0 +1,65 @@
...
+    txHash: 'deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef',
+    address:
+      'addr1q8eakg39wqlye7lzyfmh900s2luc99zf7x9vs839pn4srjs2s3ps2plp2rc2qcgfmsa8kx2kk7s9s6hfq799tmcwpvpsjv0zk3',
+    coins: 2000000,
+    outputIndex: 0,
+    tokenBundle: [],
+  },
+}
+
+const successPlanSettings = {
```
refi93, line 24: I'd name this successPlanFixtures, which is a more standard terminology for tests

refi93 (approved):
LGTM, one minor naming remark

PR #953, opened 2021-04-11
Refactor caching of transaction plans

rizip1:
* cache txPlan per TxType to avoid race-conditions
* require explicit params for "confirm" and "submitTransaction"
* keep backward functionality ("old" prefixes) for parts that were
not updated yet
* remove unnecessary race-condition check from deregister staking screen

This one is copy of https://github.com/vacuumlabs/adalite/pull/917, but due to changes in `actions.ts` it was easier to create new PR. I will close original PR and delete its branch once this is merged.


xdzurman (approved):
LGTM

PR #958, opened 2021-04-18
Webpack hot reloading

rizip1:
Resolves https://github.com/vacuumlabs/adalite/issues/941

Hot reloading for JS and CSS files to speed up the development process.

The setup is **different** from the setup we have on Adalite-2.0. Here we:
* keep express for serving the app and use `webpack-hot-middleware` && `webpack-dev-middleware` to achieve hot reloading
* keep `ts-loader` as` babel-loader` did not work with Preact in any setup I have tried and also has issues with `const enum`

Changes:
* `webpack.build.config.js` was moved one level up because the server needs it and otherwise the paths to webpack config from server and eslint (import/resolver) would conflict
* I removed a lot of "yarn install" from `package.json` scripts as I found it unnecessary
* Dev & Build scripts were removed from `app` because now the webpack config is anyway one level app and the server is responsible for development build anyway
* The webpack plugins were added due to hot-reloading css files as those css files are importing SVGs etc

_Note that this is not the FastRefresh that we have on Adalite-2.0 but standard "hot-reload"_

Future TODOs:
* update webpack and its dependencies
* setup css modules
* exchange **connect** for **useSelector** as connected components are broken after hot reload. Basically you need to reload the app in such a case, but no you need to do it anyways


rizip1:
Please also try locally if it all works for you.

app/package.json
```diff
@@ -48,7 +42,8 @@
     "fetch-readablestream": "^0.2.0",
     "mocha": "^5.0.1",
     "mocha-headless-chrome": "^2.0.3",
-    "sinon": "^7.3.2"
+    "sinon": "^7.3.2",
+    "webpack": "^4.1.1"
```
refi93, line 46: do we need webpack in both package.json files?

refi93: ah sorry I just noticed the TODO comment - still curious why do we need the two versions now

rizip1: It's because the `test` script of `app/package.json` uses it.
I could also move the webpack config that the `test` script refers to into the root folder, though then I would also need to change some mocha dependencies. Maybe it is the right thing to do, though I thought that it may be a little out of the scope of this PR. What do you think? Should I move it to the root and also move some mocha dependencies?

refi93:
I noticed that server dependencies leak into the frontend app, however, that was the case even before this PR. I found it out by importing a random backend package in the frontend and printing it, namely `express-ipfilter`. However, this is apparently a problem even on the current develop (awkward silence), not sure when it broke and given that node apparently searches in the parent folder for unresolved modules (https://stackoverflow.com/questions/32455431/prevent-require-from-looking-up-modules-in-the-parent-directory) I wouldn't be surprised if it ever worked as intended.

Fortunately, I found a way to ensure that separation from top-level dependencies by eslint, just add into `.eslintrc` the following and it will warn us if some import in the app folder ever resolves to a package in top-level node_modules:

```
"import/no-restricted-paths": [
      "error",
      {
        "zones": [
          { "target": "./app", "from": "./node_modules" }
        ]
      }
    ]
```

source: https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-restricted-paths.md

app/package.json
refi93: I'd do it now if it isn't too much effort, better have it in a consistent state to not bloat our backlog with further "tech debt", especially if it is little effort now, as doing it later would just incur the cost in form of time spent onboarding/sorting out webpack inconsitencies that possibly emerged after this PR

package.json
```diff
@@ -84,6 +84,8 @@
     "file-loader": "^6.2.0",
     "husky": "^3.0.0",
     "lint-staged": "^9.2.0",
+    "mocha": "^8.3.2",
```
refi93, line 87: better keep original versions, that's the likeliest reason behind broken tests

app/package.json
```diff
@@ -40,10 +37,7 @@
...
     "fetch-mock": "^6.5.1",
     "fetch-readablestream": "^0.2.0",
-    "mocha": "^5.0.1",
-    "mocha-headless-chrome": "^2.0.3",
-    "sinon": "^7.3.2",
-    "webpack": "^4.1.1"
+    "sinon": "^7.3.2"
   },
   "resolutions": {
     "ssri": ">=8.0.1",
```
refi93: also check if these resolutions need porting - if I remember correctly, some of them are there because of mocha

package.json
rizip1: Yeah, actually it was, reverted

refi93 (approved):
LGTM, just retain the ssri dependency resolution which yarn audit unfortunately doesn't catch, but dependabot would report it once we merge it into develop (as it did when I fixed it in https://github.com/vacuumlabs/adalite/pull/939)

PR #963, opened 2021-04-20
Remove connect from chosen components

rizip1:
* removed from:
        loading
        accountDashboard
        dashboardPage
        accountDropdown
        accountTile
        
* improves: TS inference & hot-reloading experience


refi93 (approved):
LGTM, as discussed offline I'd recommend updating the webpack readme with relevant resources for future reference should anybody face problems with hot reloading

PR #964, opened 2021-04-20
Fix incorrect styles order in development

rizip1:
The problem is that you can not apply CSS styles in any order otherwise it resolves in a very strange behavior during development.
I decided to remove the "glob" as the sorted order is not good enough for us (it needs to be specific) and we do not change screen resolutions that often. If we start using CSS-modules we would eventually drop those files completely.

The problem occurs only during development, in the production build the styles are applied correctly.

### DIFF
--- a/webpack.build.config.js
+++ b/webpack.build.config.js
@@ -12,9 +12,22 @@ if (!isProd) {
   /*
   Till we do not import css from react components, we need to have
   this workaround for Hot reload to work with global css.
+  We are not using "glob" to get all files in "css" folder as order of files matters.
   */
-  const globalCssPathnames = glob.sync('./app/public/css/**/*.css') || []
-  entry = entry.concat(globalCssPathnames)
+  const cssPathnames = [
+    './app/public/css/styles.css',
+    './app/public/css/0-767px.css',
+    './app/public/css/0-1366px.css',
+    './app/public/css/767-1366px.css',
+    './app/public/css/768-1024px.css',
+    './app/public/css/1024-1112px.css',
+  ]
+  // Check if "cssPathnames" are up-to-date with css files stored in "app/public/css" folder
+  if (JSON.stringify([...cssPathnames].sort()) !== JSON.stringify(glob.sync('./app/public/css/**/*.css').sort())) {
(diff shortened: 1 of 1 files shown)


webpack.build.config.js
```diff
@@ -12,8 +11,16 @@ if (!isProd) {
   /*
   Till we do not import css from react components, we need to have
   this workaround for Hot reload to work with global css.
+  We are not using "glob" to get all files in "css" folder as order of files matters.
   */
-  const globalCssPathnames = glob.sync('./app/public/css/**/*.css') || []
+  const globalCssPathnames = [
```
refi93, line 16: I find this name a bit misleading as the paths are obviously relative. I know this is a preexisting name but I don't really see why we couldn't just name it cssPaths or something like that.

Also, I'd consider retaining the original fetching of all css files and asserting that the lists are equal (in terms of elements contained). Otherwise I can see future ourselves scratching our heads why some css file is not being picked up, and such assertion seems like a good fool-proof measure to prevent any such surprises

refi93 (changes requested):
LGTM, just left two comments to consider

webpack.build.config.js
```diff
@@ -12,9 +12,22 @@ if (!isProd) {
...
+  const cssPathnames = [
+    './app/public/css/styles.css',
+    './app/public/css/0-767px.css',
+    './app/public/css/0-1366px.css',
+    './app/public/css/767-1366px.css',
+    './app/public/css/768-1024px.css',
+    './app/public/css/1024-1112px.css',
+  ]
+  // Check if "cssPathnames" are up-to-date with css files stored in "app/public/css" folder
+  if (JSON.stringify([...cssPathnames].sort()) !== JSON.stringify(glob.sync('./app/public/css/**/*.css').sort())) {
```
rizip1, line 26: @refi93 I did not find it necessary to write some array equals function for this. Do you think it should be there?

refi93: not really, that was just my first thought about approaching it, this is does the job as well and is simpler, so looks good to me

refi93 (approved):

PR #965, opened 2021-04-20
CSS modules

rizip1:
1st commit: has standalone PR (https://github.com/vacuumlabs/adalite/pull/964) I will rebase once it is merged

Add CSS modules and example usage for the "balance" component.
Note that not all "balance" classes are replaced as many are also used in other components.

I also decided to add **sass** support, so that we can define some common variables e.g. for screen resolution. (I can remove that commit if we do not want it)
I can investigate whether we can do the same with "custom-css-properties" (maybe we can :thinking: ) though it might be nice to also use other sass features as nested selectors etc, so this is why I would suggest keeping it.

**Note for the future:**
We should be replacing constructs like:
```
 <button
          className={`button secondary balance ${styles.refreshButton}`}
          onClick={reloadWalletInfo}
 >
          Refresh
</button>
```
With:
```
<Button
          className={styles.refreshButton}
          onClick={reloadWalletInfo}
          type="secondary"
 >
          Refresh
</Button>
```
In other words, encapsulate styles into components, otherwise, it would be difficult to remove global styles as e.g. `secondary` or `button` classnames are part of many global selectors. (This is the reason why all classes from the "balance" component were not rewritten to CSS modules)


xdzurman (commented):
I agree with introducing Sass. I looked around whether we could further shorten the `@media screen and` commands, since we're gonna see a lot of it from now on. Do you think solving this using mixins or something else would bring some value?

app/frontend/components/common/balance.module.scss
```diff
@@ -0,0 +1,55 @@
...
+    width: 20px;
+    height: 20px;
+    background-image: url('../../../public/assets/refresh_button_symbol.svg');
+  }
+
+  &:hover::before {
+    background-image: url('../../../public/assets/refresh_button_symbol_active.svg');
+  }
+
+  @media screen and (min-width: $breakpoint-tablet) and (max-width: $breakpoint-desktop-small) {
```
xdzurman, line 36: I agree with introducing Sass. I looked around whether we could further shorten the `@media screen and` commands, since we're gonna see a lot of it from now on. Do you think solving this using mixins or something else would bring some value?

rizip1: I like the mixin idea :+1: Though I would go for it once we have this usage at few more places.

server/index.js
```diff
@@ -121,6 +121,8 @@ app.get('*', (req, res) => {
           <link rel="stylesheet" type="text/css" href="css/768-1024px.css">
           <link rel="stylesheet" type="text/css" href="css/1024-1112px.css">
           <link rel="icon" type="image/ico" href="assets/favicon.ico">
+          ${''}
```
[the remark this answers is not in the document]
rizip1: Will remove before merging, thanks.

xdzurman (approved):

PR #967, opened 2021-04-26
Delegate transaction refactor

rizip1:
* Remove actions with "Old" postfix from delegation flow
* Remove race-condition check which is no longer necessary
* Move "deregisterStakeKey" to "delegate" file, instead of "poolOwner" file

For easier understanding, I suggest to review commit by commit.

### DIFF
--- a/app/frontend/actions/delegate.ts
+++ b/app/frontend/actions/delegate.ts
@@ -1,15 +1,25 @@
 import {Store, State, getSourceAccountInfo} from '../state'
+import {withdrawalPlanValidator, delegationPlanValidator} from '../helpers/validators'
 import {getWallet} from './wallet'
 import errorActions from './error'
+import loadingActions from './loading'
 import commonActions from './common'
-import {DelegateTransactionSummary, Lovelace, TxType} from '../types'
+import transactionActions from './transaction'
+import {
+  DelegateTransactionSummary,
+  CryptoProviderFeature,
+  Lovelace,
+  TxType,
+  DeregisterStakingKeyTransactionSummary,
+} from '../types'
 import debounceEvent from '../helpers/debounceEvent'
-import {delegationPlanValidator} from '../helpers/validators'
 
 export default (store: Store) => {
   const {setState, getState} = store
   const {setError} = errorActions(store)
-  const {prepareTxPlan, setTransactionSummaryOld} = commonActions(store)
+  const {loadingAction, stopLoadingAction} = loadingActions(store)
+  const {prepareTxPlan, setTransactionSummary} = commonActions(store)
+  const {confirmTransaction} = transactionActions(store)
 
   const hasPoolIdentifiersChanged = (state: State) => {
     const {shelleyDelegation: newShelleyDelegation} = getState()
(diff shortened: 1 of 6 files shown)


xdzurman (approved):
LGTM

app/frontend/actions/delegate.ts
```diff
@@ -131,10 +144,71 @@ export default (store: Store) => {
...
 
+  const delegate = async (state: State): Promise<void> => {
+    return await confirmTransaction(state, {
+      txConfirmType: TxType.DELEGATE,
+      txPlan: state.cachedTransactionSummaries[TxType.DELEGATE].plan,
+      sourceAccountIndex: state.sourceAccountIndex,
+    })
+  }
+
+  const deregisterStakingKey = async (state: State): Promise<void> => {
```
refi93, line 155: I find it a bit odd that the file grouping these actions is called delegate - if it tries to group actions related to staking, perhaps that would be a more descriptive name, but I understand that would trigger quite a lot of changes, so this is rather a note for a later iteration of refactoring

app/frontend/actions/delegate.ts
```diff
@@ -131,10 +144,71 @@ export default (store: Store) => {
...
+    const supportError = getWallet().ensureFeatureIsSupported(CryptoProviderFeature.WITHDRAWAL)
```
refi93, line 156: ensureFeatureIsSupported seems to not return anything as per the typing: https://github.com/vacuumlabs/adalite/blob/e4cde52738dbb2e224f1d6df68a13a3113c12446/app/frontend/types.ts#L44

refi93: ah I see, I should be looking at the function with the same name, but declared in shelley-wallet.ts :facepalm: - this is madness :D If it returns a boolean, something like isFeatureSupported would imho be more spot-on as a name

app/frontend/components/pages/delegations/shelleyBalances.tsx
```diff
@@ -66,11 +65,7 @@ const shelleyBalances = ({
         )}
       </div>
       {!!rewardsAccountBalance && (
-        <button
-          disabled={calculatingDelegationFee}
-          className="button secondary balance withdraw"
-          onClick={withdrawRewards}
-        >
+        <button className="button secondary balance withdraw" onClick={withdrawRewards}>
```
refi93, line 68: didn't the disabling of the button while delegation fees are being computed serve as a protection against repeated triggering of the logic? Shouldn't we debounce it then?

rizip1: Well, the `calculatingDelegationFee` is not related to `withdrawRewards`, but to calculation of delegation fees trigers immediately when user lands on this screen. Now they are not sharing transaction plans, so this is why it was removed. Do you see there some other potential problems?

rizip1: And the repeated protection is not there, but also was not there, as `calculatingDelegationFee` is not set when `withdrawRewards` is called.

app/frontend/actions/delegate.ts
rizip1: Thanks, I could create an issue for that, so it can be changed in standalone PR.

refi93 (approved):

PR #968, opened 2021-04-27
Add support for catalyst voting

xdzurman:
- Adds a registration card to Advanced tab (for now, we may find a better place), which is used to open a 3 step registration modal
- The requirement for voting is that the round is active, the user has more than 500 ADA and their staking key is registered
- Introduces ability to send voting metadata in transactions
- Adds unit and integration tests for voting

TODOs:
- Update trezor support version for catalyst (ensureFeatureIsSupported) once it's known
- Rewrite "encryptWithPassword" function to use only libraries adalite used until now, and remove these new packages (chacha, pbkdf2). This encryption part is currently heavily inspired by yoroi

Questions:
- Will it be possible to extract types from trezor?
- Is the current flow OK, or would we want to create the tx plan already in the Voting Dialog (and not later in Voting Summary), so that the user can see the fee, etc.?

PS:
- Although there are still some TODOs, I created this PR for overall transparency and the ability for you to pull an "early break" if I'm heading in the wrong direction


.env.example
```diff
@@ -62,3 +62,6 @@ ADALITE_ENABLE_LEDGER="true"
 
 # delegation by ticker
 ADALITE_ENABLE_SEARCH_BY_TICKER=true 
+
+# catalyst next round start and end dates in UNIX format (example: 2021-2030)
+ADALITE_NEXT_VOTING_UNIX_PAIR=1609459200000, 1893456000000
```
refi93, line 67: I'd rather use ISO timestamps which are easier to sanity-check on first sight, and split the start and end into separate env variables

app/public/css/styles.css
```diff
@@ -4385,3 +4385,144 @@ label.button.primary.disabled {
 }
 
 /* STAKE KEY DEREGISTER: (END) */
```
refi93, line 4387: I'd consider switching this to css modules which should be the default way of doing designs going forward. Here's a PR introducing them: https://github.com/vacuumlabs/adalite/pull/965

.env.example
refi93: also, don't forget to add this environment variable to the preview app env variables in heroku, otherwise the deployment fails. I added it, but another commit to trigger the deploy is probably needed

app/frontend/helpers/catalyst.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// hotfix alternative to crypto-random-string
+const cryptoRandomString = ({length}: {length: number}): string => {
+  const randBytes = Buffer.alloc(length)
+  window.crypto.getRandomValues(randBytes)
+  return randBytes.toString('hex')
+}
+
+export async function encryptWithPassword(
```
refi93, line 36: seems worth covering with unit tests so we are sure that the implementation doesn't accidentally change which would otherwise be hard to spot

app/public/css/styles.css
xdzurman: Thanks, we already discussed this with Riso previously and that's exactly the plan

.env.example
xdzurman: I agree with the ISO format. However, the two timestamps don't make much sense without each other for me. This way, the person who sets them will not accidentally forget to set the other one, because they're in pair. When they're in this natural order, the mistake of setting them contrariwise is mitigated to some extent as well, we can even .sort() them in the worst case when they're like this

refi93: the presence of both timestamps (or neither) and order can be enforced when validating the config. I would rather not be too graceful with their handling - anything odd would be intercepted when deploying as that's when we validate the config as well (if I'm not mistaken)

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -121,7 +141,11 @@ export function estimateTxSize(
 
   const txWitnessesSize = shelleyWitnessesSize + byronWitnessesSize
 
-  const txMetaSize = 1 // currently null
+  const placeholderMetaSignature = 'x'.repeat(CATALYST_SIGNATURE_BYTE_LENGTH * 2)
+  const txAuxiliaraDataSize = auxiliaryData
```
refi93, line 145: I'd expect here some switch statement which would apply only if the auxiliary data passed is a voting registration, so typescript warns us if we happen to serialize other kind of metadata, where this formula wouldn't work

refi93:
> Should I prepare the code for a more universal of handling of tx auxiliary data? Right now it's kind of fixed towards voting only
I wouldn't overgeneralize it, but at the same time I would make sure that whatever logic relies on auxiliary data being a voting registration that there would be branching (switch/if-else/...) ensuring that typescript would warn about unhandled cases should we introduce other kind of auxiliary data, so we know where to update the logic

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -136,15 +148,36 @@ CryptoProviderParams): Promise<CryptoProvider> => {
...
+    const txAuxiliaryData = cborizeTxAuxiliaryVotingData(auxiliaryData, registrationDataSignature)
+    return txAuxiliaryData
+  }
+
   async function signTxGetStructured(
     txAux: TxAux,
     addressToPathMapper: AddressToPathMapper
   ): Promise<CborizedTxSignedStructured> {
-    const {shelleyWitnesses, byronWitnesses} = await prepareWitnesses(txAux, addressToPathMapper)
+    let _txAux = txAux
```
refi93, line 167: is the 'txAux` value used at all? can't (and shouldn't) this be just `null`?

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -472,21 +498,34 @@ const ShelleyLedgerCryptoProvider = async ({
...
+    let _txAux = txAux
+    let txMeta = null
+    if (txAux.auxiliaryData) {
+      txMeta = cborizeTxAuxiliaryVotingData(
+        txAux.auxiliaryData,
+        response.auxiliaryDataSupplement.catalystRegistrationSignatureHex
+      )
+      _txAux = ShelleyTxAux({
+        ...txAux,
+        metadataHash: response.auxiliaryDataSupplement.auxiliaryDataHashHex,
```
refi93, line 515: this should rather be auxiliaryDataHash

.env.example
xdzurman: fixed

app/frontend/helpers/catalyst.ts
xdzurman: added in #8edba1a1a0e31a288a163ff247218ab360d45a06

refi93:
@xdzurman I just release ledgerjs 3.1.0 on npm so you can update to it: https://www.npmjs.com/package/@cardano-foundation/ledgerjs-hw-app-cardano

xdzurman:
Closing and re-opening due to the fact that heroku review apps killed this PR after 5 days of inactivity.

PR #984, opened 2021-05-10
Bulk export gradually more public keys

PeterBenc:
Motivation: 
 - for wallets that have a lot of account ist more convenient to export public key by batches bigger than 5

Changes: 
 - public keys are exported in batches of 5, 10, 20, 20.... (from 20 the batch stays the same, exporting more than 20 key from ledger takes a bit too long)

### DIFF
--- a/app/frontend/wallet/constants.ts
+++ b/app/frontend/wallet/constants.ts
@@ -116,6 +116,6 @@ export const TREZOR_ERRORS = {
   [CryptoProviderFeature.POOL_OWNER]: InternalErrorReason.TrezorPoolRegNotSupported,
 }
 
-export const MAX_ACCOUNT_INDEX = 30
+export const MAX_ACCOUNT_INDEX = 29
 
-export const MAX_BULK_EXPORT_AMOUNT = 5
+export const EXPORTED_PATHS_SLICE_SIZE = 5
--- a/app/frontend/wallet/constants.ts
+++ b/app/frontend/wallet/constants.ts
@@ -116,6 +116,4 @@ export const TREZOR_ERRORS = {
   [CryptoProviderFeature.POOL_OWNER]: InternalErrorReason.TrezorPoolRegNotSupported,
 }
 
-export const MAX_ACCOUNT_INDEX = 29
-
-export const EXPORTED_PATHS_SLICE_SIZE = 5
+export const MAX_ACCOUNT_INDEX = 30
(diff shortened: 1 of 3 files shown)


app/frontend/wallet/helpers/CachedDeriveXpubFactory.ts
```diff
@@ -53,25 +53,28 @@ function CachedDeriveXpubFactory(derivationScheme, shouldExportPubKeyBulk, deriv
     const accountIndex = derivationPath[2] - HARDENED_THRESHOLD
     const currentAccountPage = Math.floor(accountIndex / MAX_BULK_EXPORT_AMOUNT)
 
-    for (let i = 0; i < MAX_BULK_EXPORT_AMOUNT; i += 1) {
+    // we want to scale the number of exported keys up to 3 times the MAX_BULK_EXPORT_AMOUNT
+    const nAccounts = Math.min(currentAccountPage + 1, 3) * MAX_BULK_EXPORT_AMOUNT
```
refi93, line 57: This is getting really misleading - MAX_BULK_EXPORT_AMOUNT is not the maximum number of exported public keys anymore and the "current page" is not really the "current page" as these are now of variable size and not really "pages" as units of exported data.

I'd consider renaming the `MAX_BULK_EXPORT_AMOUNT` constant to `EXPORTED_PATHS_SLICE_SIZE`, `currentAccountPage` to `currentAccountSlice`, and to get the `nAccounts` I'd rather make a helper function `getExportedBulkSize` which would take as an argument the currentAccountSlice and have explicit `if-else` for returning the bulk size than a rather cryptic, albeit short formula

refi93 (changes requested):
see comment

app/frontend/wallet/constants.ts
```diff
@@ -120,6 +120,6 @@ export const TREZOR_ERRORS = {
   [CryptoProviderFeature.POOL_OWNER]: InternalErrorReason.TrezorPoolRegNotSupported,
 }
 
-export const MAX_ACCOUNT_INDEX = 30
+export const MAX_ACCOUNT_INDEX = 29
```
PeterBenc, line 123: Changed this to 29 since exporting 30 accounts means exporting up to index 29.

app/frontend/wallet/constants.ts
```diff
@@ -116,6 +116,4 @@ export const TREZOR_ERRORS = {
   [CryptoProviderFeature.POOL_OWNER]: InternalErrorReason.TrezorPoolRegNotSupported,
 }
 
-export const MAX_ACCOUNT_INDEX = 30
-
-export const MAX_BULK_EXPORT_AMOUNT = 5
+export const MAX_ACCOUNT_INDEX = 34
```
refi93, line 119: why did we have to bump this? I'd rather stay at 30 if we don't have a strong reason to bump it, 34 seems really arbitrary

refi93 (approved):
overall LGTM, just one question regarding MAX_ACCOUNT_INDEX

PR #988, opened 2021-05-12
Add support for catalyst voting

xdzurman:
- Adds a registration card to Voting Tab, which is used to open a 3 step registration modal
- The requirement for voting is that the round is active, the user has more than 500 ADA and their staking key is registered
- Introduces ability to send voting metadata in transactions
- Adds unit and integration tests for voting

TODOs:
- Rewrite "encryptWithPassword" function to use only libraries adalite used until now, and remove these new packages (chacha, pbkdf2). This encryption part is currently heavily inspired by yoroi

Questions:
- Will it be possible to extract types from trezor?


refi93:
In the review app I noticed that voting doesn't appear to be available. I'd consider exposing the tab, but adding something like "coming soon"

Also not sure if already added but we probably should at least warn if not prevent from proceeding further users that have their staking keys not registered and less than 450 ADA staking key balance. I'd personally rather warn all users about that (even if they have more so they don't decrease their balance until the snapshot is made/rewards distributed) and outright stop users that don't have it as there's little point for letting them go further until they fix their balance and it's more than likely such users would end up more confused at the end as nobody would remind them to top up their account later on.

If we add this "coming soon" banner and restrictions, I see no harm in releasing this functionality already so it isn't blocked and we can just push the release as the dates are clarified

app/package.json
```diff
@@ -18,7 +18,7 @@
   "homepage": "https://github.com/vacuumlabs/adalite#readme",
   "scripts": {},
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "^3.0.0",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "https://github.com/vacuumlabs/ledgerjs-cardano-shelley/releases/download/v3.1.0-catalyst-rc.5/cardano-foundation-ledgerjs-hw-app-cardano-v3.1.0-catalyst-rc.5.tgz",
```
refi93, line 21: this can be updated to the latest version published in npm

app/frontend/components/pages/voting/votingCard.tsx
```diff
@@ -40,7 +40,7 @@ const VotingCard = (): h.JSX.Element => {
   const activeAccount = useActiveAccount()
   const hasRegisteredStakingKey = hasStakingKey(activeAccount)
 
-  const getValidationMessage = (): string => {
+  const getUnmetPreconditionMessage = (): string => {
```
refi93, line 43: I guess this should be
```javascript
  const getUnmetPreconditionMessage = (): string | null => {
```

And I'm wondering - why didn't typescript catch it? :thinking:

refi93 (approved):

refi93:
LGTM, just one remark regarding typescript - worth checking why typescript didn't complain

app/frontend/components/pages/voting/votingCard.tsx
xdzurman: https://www.amitmerchant.com/strictly-check-null-undefined-typescript/ we have typescript's strictNullChecks off

refi93: ahhh I see - do you know what's the reason? I'd guess it's meant as a temporary solution as we migrate(d) from TS, so perhaps worth adding an issue for reenabling it in the backlog and fixing whatever violations remain in the code

xdzurman: Yes, it affects all the exhaustive ifs and switches as well. It brings a lot of value, but it would be an enormous PR as well.

refi93: I see, still IMHO probably worth adding explicitly in the backlog to do it eventually, if there's no other workstream as part of the broader type coverage chore

xdzurman: Yes of course, I agree, just saying

PR #991, opened 2021-05-17
Alter server error message if epoch transition is underway

xdzurman:


app/frontend/errors/errorMessages.ts
```diff
@@ -79,9 +80,12 @@ const internalErrorMessages: {[key in InternalErrorReason]: (params?: any) => st
   [InternalErrorReason.CryptoProviderError]: ({message}) => `CryptoProviderError: ${message}`,
   [InternalErrorReason.NetworkError]: () =>
     'NetworkError: Request to our servers has failed. Please check your network connection and if the problem persists, contact us.',
-  [InternalErrorReason.ServerError]: () =>
-    'ServerError: Our servers are probably down. Please try again later and if the problem persists, contact us.',
-
+  [InternalErrorReason.ServerError]: () => {
```
refi93, line 83: overriding the translation message based on some external context feels a bit sketchy - these are meant to be static/pure functions after all. I'd rather have a separate error reason for epoch boundary and create a helper function which would throw that error reason instead of the generic ServerError if the epoch boundary is detected. We can use that helper wherever we throw the server error (which should be in 2 places AFAIK) which would also be more transparent to whoever is inspecting the errors being thrown

app/frontend/helpers/isEpochBoundaryUnderway.ts
```diff
@@ -0,0 +1,16 @@
+import * as moment from 'moment'
+import {EPOCH_TRANSITION_MAX_MINUTES} from '../constants'
```
refi93, line 2: I'd consider just hardcoding the constant, I don't really see it being reused and whoever inspects this code would just have the overhead of having to look up the constant, and whoever inspects the constants file would wonder, what's that constant for even though it's rather unimportant compared to others

app/frontend/errors/errorMessages.ts
```diff
@@ -79,9 +80,12 @@ const internalErrorMessages: {[key in InternalErrorReason]: (params?: any) => st
...
+    if (isEpochBoundaryUnderway()) {
+      return 'Cardano is undergoing an epoch boundary. We should be back in a few minutes.'
```
refi93, line 85:
```suggestion
      return 'Our servers are temporarily down while Cardano is undergoing an epoch boundary. We should be back in a few minutes.'
```

app/frontend/helpers/epochBoundaryUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+  const dayDiff = now.diff(epochTransition, 'days')
+  const relativeMinuteDiff = now.subtract(dayDiff, 'days').diff(epochTransition, 'minutes')
+  // multiple of 5 days passed and it's no more than X minutes since the transition
+  if (dayDiff % 5 === 0 && relativeMinuteDiff < 60) {
+    return true
+  }
+  return false
+}
+
+export function probeEpochBoundary(params?: OptionalParams): void {
```
refi93, line 17: I'd rename it `throwIfEpochBoundary` to make it clear it throws an error and also to make it easier to follow when used together with throwing other errors in the logic that calls it

refi93 (approved):

PR #993, opened 2021-05-18
Check max utxo size also for change

PeterBenc:
Motivation: 
 - previously we were checking only outputs without change for max utxo size :facepalm: 

Changes: 
 - we check also change for max utxo size

Gotcha: 
- ideally we would create multiple outputs for transactions containing too many tokens but that would require significantly more work, I suggest we address it separately, since this anyway effect small minority of users.

### DIFF
--- a/app/frontend/wallet/shelley/shelley-transaction-planner.ts
+++ b/app/frontend/wallet/shelley/shelley-transaction-planner.ts
@@ -384,7 +384,11 @@ const validateTxPlan = (txPlanResult: TxPlanResult): TxPlanResult => {
     }
   }
 
-  if (outputs.some((output) => encode(cborizeSingleTxOutput(output)).length > MAX_TX_OUTPUT_SIZE)) {
+  if (
+    outputsWithChange.some(
+      (output) => encode(cborizeSingleTxOutput(output)).length > MAX_TX_OUTPUT_SIZE
+    )
+  ) {
     return {
       ...noTxPlan,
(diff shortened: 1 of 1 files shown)


refi93 (approved):

app/frontend/wallet/shelley/shelley-transaction-planner.ts
```diff
@@ -384,7 +384,11 @@ const validateTxPlan = (txPlanResult: TxPlanResult): TxPlanResult => {
     }
   }
 
-  if (outputs.some((output) => encode(cborizeSingleTxOutput(output)).length > MAX_TX_OUTPUT_SIZE)) {
+  if (
+    outputsWithChange.some(
```
refi93, line 388: just wondering - don't our transactions always have one output and at most one additional change output? If so, what's the point in having an "outputs" parameter?

PeterBenc: Hm thats a good point, I wasnt the one to introduce this, tho you are right. We also have an array for withdrawals even tho we always have one. Do u think its something worth refactoring?

refi93: until now it was true, but I also realized now we will likely need to introduce breaking down of big outputs into multiple ones and I guess we want to do it soon, as NFTs are failing because of that, if I understood correctly, so it probably makes sense to change this tx plan computation more significantly and perhaps make such renaming (if it would still make sense) part of that

PR #996, opened 2021-05-18
Add Changelly widget

DavidTranDucVL:
Add Changelly widget as temporary solution.


xdzurman:
Closing and reopening to force heroku deploy (which kills apps after 6 days of inactivity)

app/frontend/components/pages/exchange/exchange.module.scss
```diff
@@ -0,0 +1,32 @@
...
+.iframeWrapper {
+  box-sizing: content-box;
+  max-width: 568px;
+  width: 100%;
+  height: 398px;
+  @media screen and (max-width: $breakpoint-mobile) {
+    box-sizing: border-box;
+    height: unset;
+  }
+}
```
xdzurman, line 32: what's the point of this wrapper since it's just a subset of its parent?

app/frontend/components/pages/exchange/exchange.tsx
```diff
@@ -0,0 +1,25 @@
...
+          >
+            Can't load widget
+          </iframe>
+        </div>
+      </div>
+    </div>
+  </div>
+)
+
```
xdzurman, line 24: let's use this thread so that the topmost level is not spammed - Did you already test this as Michal asked?

app/frontend/components/pages/exchange/exchange.module.scss
DavidTranDucVL: Valid point, removed that part of code

app/frontend/components/pages/exchange/exchange.tsx
DavidTranDucVL: There is apparently minimum ADA required for a transaction. Do we have an account with more then 35 ADA?

app/frontend/components/pages/exchange/exchange.module.scss
xdzurman: Please remove the "iframeWrapper" CSS as well

xdzurman (approved):
LGTM, just test thoroughly

app/frontend/components/pages/exchange/exchange.module.scss
DavidTranDucVL: Fixed

PR #999, opened 2021-05-20
Fix parsing of ADA amount

### LINKED ISSUE #997: Rounding Issue in AdaLite
Hello, as you may or may not be aware, I have been creating a large number of Cardano NFT projects lately which rely on the concept of sending and receiving a specific, serialized dust value between wallets to confirm transactions. Since the time of the CryptoKnitties original launch (04/23/2021) through today, I routinely have been receiving support requests from purchasers asking why their transaction "timed out".

After a little bit of research and digging, we always come back to the same conclusion: The customer was using AdaLite wallet and the wallet sent a value that was "off by 1" to our dust address.

For example:
1. Customer visits website and receives a dust value of 8.131699 Ada. They must send this precise amount to the recipient wallet in order to receive their token within the allotted timeframe.
2. The dust address actually receives a value of 8.131698 Ada (See Transaction ID: [c156d601dc6085dc2ec5bc904f72ea6c0f327fdf5ab14406a6cd58c1191aabe8](https://explorer.cardano.org/en/transaction?id=c156d601dc6085dc2ec5bc904f72ea6c0f327fdf5ab14406a6cd58c1191aabe8)).

This does generally seem to happen when the last Lovelace digit of the transaction is a 9 so I'm not sure if that is helpful or not.

PeterBenc:
Fixes #997

### DIFF
--- a/app/frontend/helpers/validators.ts
+++ b/app/frontend/helpers/validators.ts
@@ -8,7 +8,8 @@ import {InternalErrorReason} from '../errors'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 const parseToLovelace = (str: string): Lovelace =>
-  Math.trunc(toCoins(parseFloat(str) as Ada)) as Lovelace
+  // Math.round to solve edge cases in floating number precision
+  Math.round(toCoins(parseFloat(str) as Ada)) as Lovelace
 
 const sendAddressValidator = (fieldValue: string) => {
   if (fieldValue === '') {
--- a/app/frontend/helpers/validators.ts
+++ b/app/frontend/helpers/validators.ts
@@ -8,7 +8,7 @@ import {InternalErrorReason} from '../errors'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 const parseToLovelace = (str: string): Lovelace =>
-  // Math.round to solve edge cases in floating number precision
+  // Math.round to solve edge cases in floating number precision like '8.131699'
   Math.round(toCoins(parseFloat(str) as Ada)) as Lovelace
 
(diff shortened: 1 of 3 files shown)


app/frontend/helpers/validators.ts
```diff
@@ -8,7 +8,8 @@ import {InternalErrorReason} from '../errors'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 const parseToLovelace = (str: string): Lovelace =>
-  Math.trunc(toCoins(parseFloat(str) as Ada)) as Lovelace
+  // Math.round to solve edge cases in floating number precision
+  Math.round(toCoins(parseFloat(str) as Ada)) as Lovelace
```
refi93, line 12: I'd consider adding a unit test covering this helper function with the exact edge case(s) where the old implementation failed - it would serve as documentation and prevention of reintroduction of the same error in the future

refi93 (changes requested):
see comment

app/frontend/helpers/validators.ts
PeterBenc: I added a test, tbh I find the comment more than self-explanatory, we can't explain the reason behind the errors in precision, nor we cant provide an example that explains it since they are basically infinite of them. I think anyone who would try to understand would need to figure it out themselves, and would google part of the comment "edge cases in floating number precision" and get a general idea of what is going on. I added an example to the comment, please let me know if you have any suggestions on how to make it more explanatory.

refi93: yeah, I didn't mean an exhausting explanation but rather concrete examples, from which you as a reader can get a better idea where to start from and what class of edge-cases is relevant for this specific piece of logic, and tests are a suitable place for such edge-case examples, but no harm in adding such in the comment as well

PR #1006, opened 2021-05-31
6.4.1

DavidTranDucVL:


refi93:
regarding Catlyst, I noticed `ADALITE_NEXT_VOTING_START` is set to `2021-01-01T00:00:00.000Z` - I guess we should set it to some future date before deploying this to prod, to avoid launching Catalyst prematurely

refi93 (approved):

PR #1008, opened 2021-05-31
Remove Ledger transport fallback to U2F, fix wallet logout

refi93:
Motivation: I was testing webhid on Chrome further and I realized that when the user hits "cancel" in the webhid pop-up, the window closes and Adalite seems to end up being stuck in an infinite "loading wallet data" state with the spinner. This seems to be caused by non-working U2F in Chrome 91 which Adalite automatically falls back to after getting a cancellation error from WebHID transport and can be quite confusing if users on Chrome actually face issues with WebHID

I'm not actually sure at this point, why we introduced the U2F fallback as Myether wallet doesn't seem to do it: https://github.com/MyEtherWallet/MyEtherWallet/blob/bd8416a199da87a29f24bc4b5668a0273f34488a/src/wallets/hardware/ledger/index.js#L148

Looking at the commit where we introduced it I'm none the wiser as it doesn't seem to point to any specific issue: https://github.com/vacuumlabs/adalite/commit/617c6b2cd4ef22b0c2b14bd67b45cf127997e882 - do you remember @PeterBenc  why did we need it? If it was because of Brave, the latest version seems to support WebHID just fine

Other fixes:
* unify "hw wallet suggestion" with "ledger troubleshooting suggestion" as it was a bit painful to maintain/combine both and in many cases the base suggestion of updating firmware is misleading, so better complement it with the suggestion to turn off all other apps
* I realized after logout and login to the wallet we are getting an "InvalidState" error which is apparently due to not closing the old transport first, so I added a module level `_activeTransport` variable in the ledger crypto provider to make sure to close the previously active transport before getting the new one. The other alternative was just returning the cached transport but I would be afraid from the old transport eventually "stalling" which is hard to test/debug, so I'm better creating a new one.

Testing:
* tested wallet login on Chrome/FF/Brave/Opera on Ubuntu 20.04
* you can notice that on current develop you get an error when you attempt to log in with ledger after logging out, this shouldn't be happening here anymore

### DIFF
--- a/app/frontend/errors/errorMessages.ts
+++ b/app/frontend/errors/errorMessages.ts
@@ -7,10 +7,8 @@ import {knownExternalErrors, InternalErrorReason} from '.'
 
 const {ADALITE_MIN_DONATION_VALUE} = ADALITE_CONFIG
 
-const ledgerTroubleshootingSuggestion =
-  'If you are using Ledger, please make sure Ledger Live app is closed and try connecting your device using the "Connect with WebUSB" functionality (the button underneath "Unlock with Ledger"). For more information please read the section concerning Ledger in our troubleshooting suggestions.'
-const updateHwWalletAppSuggestion =
-  'If you are using a hardware wallet, please make sure you are using the latest version of the Cardano application.'
+const hwWalletTroubleshootingSuggestion =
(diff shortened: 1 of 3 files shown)


PeterBenc (approved):
One minor remark, otherwise LGTM, I tested login on ubuntu, I wasnt able to reproduce the error when re-logging, it worked for also on develop. 

Regarding the question, I think the behavior that was set in mentioned commit was the request of Michal, and I don't think it has anything to do with brave.

PR #1010, opened 2021-06-01
Disable trezor on mobile devices

PeterBenc:
Changes: 
 - one fix in webpack related to this issue https://github.com/preactjs/preact/issues/1334
 - remove `platform` lib, it seems its no longer maintained and it didn't offer the API we need
 - add `react-device-detect-lib` as a better replacement
 
 Fixes: #1010

### DIFF
--- a/app/frontend/components/pages/login/hardwareAuth.tsx
+++ b/app/frontend/components/pages/login/hardwareAuth.tsx
@@ -7,6 +7,7 @@ import {useActions} from '../../../helpers/connect'
 import actions from '../../../actions'
 import {useState, useCallback} from 'preact/hooks'
 import {localStorageVars} from '../../../localStorage'
+import {isMobileOnly} from 'react-device-detect'
 
 const LoadByHardwareWalletSection = () => {
   const {loadWallet} = useActions(actions)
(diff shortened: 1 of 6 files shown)


app/frontend/components/pages/login/hardwareAuth.tsx
```diff
@@ -46,11 +47,12 @@ const LoadByHardwareWalletSection = () => {
             dangerouslySetInnerHTML={{__html: '&nbsp;'}}
           />
           <button
-            disabled={!ADALITE_CONFIG.ADALITE_ENABLE_TREZOR}
+            disabled={!ADALITE_CONFIG.ADALITE_ENABLE_TREZOR || isMobileOnly}
             {...tooltip(
               'Support for Trezor is temporarily disabled',
               !ADALITE_CONFIG.ADALITE_ENABLE_TREZOR
             )}
+            {...tooltip('Not supported on mobile devices', isMobileOnly)}
```
refi93, line 55: the tooltip seems to stay there "forever" until I switch to another screen, don't we have some convenient way to make it only temporarily visible? :thinking:

refi93 (approved):
tested on mobile and desktop, works as expected. I have one small remark, but if there's no simple way to address it, it's good enough to leave it like it is

PR #1025, opened 2021-06-07
Allow multiple change outputs

PeterBenc:
Changes: 
 - Splits change outputs based on the number of tokens

Gotcha:
 - first commit only moves stuff around, so I recommend to review by commits, 
 - Im currently working n tests


app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,268 @@
+import {InternalError, InternalErrorReason, UnexpectedErrorReason} from '../../../errors'
+import {CertificateType, Lovelace} from '../../../types'
+import {MAX_TX_OUTPUT_SIZE} from '../../constants'
```
refi93, line 3: does this limit make sense anymore if we chose to put a fixed threshold on the amount of tokens in a single output?

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,268 @@
...
+      success: false,
+      minimalLovelaceAmount: additionalLovelaceAmount,
+      estimatedFee: feeWithoutChange,
+      deposit,
+      error: {code: UnexpectedErrorReason.CannotConstructTxPlan},
+    }
+  }
+
+  // No change necessary, perfect fit
+  if (totalOutput + feeWithoutChange === totalInput && tokenDifference.length === 0) {
```
refi93, line 71: what if you are sending exactly one token out and it's a perfect match when it comes to ADA? Wouldn't the tokenDifference be non-empty, hence causing the code to produce an invalid change output with zero ADA and zero tokens?

refi93: so perhaps the check should be

```javascript
tokenDifference.every(({quantity}) => quantity === 0)
```

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
PeterBenc: In a way yes, we still validate every output if its not too big (even tho they should not be) and we split outputs only if they surpass this value. Initially, I thought we might wanna split outputs regardless of the max size, but then I changed it so we split them only in that case. What do you think?

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
PeterBenc: Agree, I think the situation you explained in the first comment wont happen since tokenDifference.length is 0 if its perfect fit (0 differences are omitted) but I would suggest creating a variable called `isTokenDifferenceEmpty` which would be defined as you proposed to emphasize this.

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
refi93: I'd prefer to keep it simple - if the concern are fees, the user may save some by producing the bigger utxo now, but that would "kick them back" anyway if they choose to send some of the tokens out as the respective change output would be big.

So I don't see that as some improvement worth the additional complexity as it may not even be the most profitable strategy in the general case.

Also bigger outputs mean higher chances of issues with Trezor which still doesn't have unlimited throughput and/or interoperability with wallets which still struggle with big outputs.

Or did you see some other benefit that makes it worth to not break outputs down until they hit the global limit?

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,268 @@
...
+  ) as Lovelace
+
+  const lovelaceChange = totalInput - totalOutput - changeAdditionalLovelaceAmount - feeWithChange
+
+  // lovelaceChange is added to the first change output
+  // so it has to be positive to keep its min ada value
+  if (tokenDifference.length > 0 && lovelaceChange < 0) {
+    return {
+      success: false,
+      minimalLovelaceAmount: additionalLovelaceAmount,
```
refi93, line 162: minlLovelaceAmount would be more consistent with other naming I guess

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
PeterBenc: My initial thought was that it "is ok to set a custom policy regarding change output splitting, but there is no need to set also custom `maxOutputSize` which we would kind of do if we split every output).  So basically my idea was to create a solution for an edge case, and don't influence also other cases if not necessary. Other than that I don't see any benefits to "not splitting all outputs based on token quatity".

Regarding the codebase, this is one condition we can easily remove so if you see any benefits to splitting every output (like the trezor support), lets do it.

refi93: then I'd do the splitting by the fixed amount of max tokens per output all times - the logic is already quite hard to follow with all the edge-cases so I'd optimize for simplicity

app/frontend/wallet/max-amount-calculator.ts
```diff
@@ -24,15 +26,27 @@ export const MaxAmountCalculator = () => {
         profitableInputs.map(({tokenBundle}) => tokenBundle)
       )
       const inputBalance = getInputBalance(profitableInputs)
-      const additionalLovelaceAmount =
+
+      const changeOutputs = splitTxOutput(
+        {
+          isChange: false,
```
refi93, line 32: probably doesn't matter here as it is just for the sake of a calculation, but it seems fishy the flag is false while the variable name is changeOutputs

app/tests/src/transaction-planner.js
```diff
@@ -1,6 +1,27 @@
 import assert from 'assert'
-import {selectMinimalTxPlan} from '../../frontend/wallet/shelley/shelley-transaction-planner'
+import {selectMinimalTxPlan} from '../../frontend/wallet/shelley/transaction'
 import {AssetFamily, TxType} from '../..//frontend/types'
+import {UnexpectedErrorReason} from '../../frontend/errors'
+// import range from '../../frontend/wallet/helpers/range'
```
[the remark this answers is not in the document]
PeterBenc: I have yet to finish these tests, the `WIP` prefix should have emphasize this but sorry, I should have left a comment.

app/frontend/wallet/shelley/transaction/splitTxOutput.ts
```diff
@@ -0,0 +1,16 @@
...
+import {computeMinUTxOLovelaceAmount} from './utils'
+
+export const splitTxOutput = (output: TxOutput, maxOutputTokens: number): TxOutput[] => {
+  const nOutputs = Math.ceil(output.tokenBundle.length / maxOutputTokens)
+  const outputs: TxOutput[] = []
+  for (let i = 0; i < nOutputs; i++) {
+    const tokenBundle = output.tokenBundle.slice(i * maxOutputTokens, (i + 1) * maxOutputTokens)
+    outputs.push({
+      ...output,
+      coins: computeMinUTxOLovelaceAmount(tokenBundle),
```
refi93, line 11: I guess this is really doable only for a change output. You wouldn't really want to overwrite the amount being sent out for a non-change output. I would at least rename the function to splitTxChangeOutput to make it clear it applies to change outputs only and perhaps leave here a clarifying comment warning that this doesn't preserve original amount being sent at all

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,268 @@
...
+      estimatedFee: feeWithChange,
+      deposit,
+      error: {code: UnexpectedErrorReason.CannotConstructTxPlan},
+    }
+  }
+
+  const firstChangeOutput: TxOutput = {
```
refi93, line 169: isn't there more value in just adding one more ada-only tx output? The ada in token outputs is imho kind of "locked up" so I'd try to "free it up" even at the cost of some fees which the user saves when performing ada-only transactions as the fee is linear

refi93: though I see - things would get hairy if the amount is less than minimum ada per utxo, though that should be a constant for ada-only utxos

PeterBenc: Thats a good point regarding the "freeing of ADA only outputs" but other than the const of fees, this would also require the output to be >= 1 ADA, unlike now when it only needs to be positive. What do you think? I

app/frontend/wallet/max-amount-calculator.ts
```diff
@@ -24,15 +26,27 @@ export const MaxAmountCalculator = () => {
...
+          address,
+          coins: computeMinUTxOLovelaceAmount(inputsTokenBundle) as Lovelace,
```
refi93, line 34: is there any point in passing a computed amount of coins if splitTxOutput overwrites the amounts?

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  const noTxPlan: TxPlanResult = {
+    success: false,
+    error: null,
+    estimatedFee: fee,
+    deposit,
+    minimalLovelaceAmount: additionalLovelaceAmount,
+  }
+
+  const outputsWithChange = change ? [...outputs, ...change] : outputs
```
refi93, line 276: empty array is truthy in JS, so I believe this actually always ends up in the left branch, or can the "change" be null? In that case I'd rathe rbe more explicit and check for `change != null`

PeterBenc: I totally missed this, thanks, will fix :+1:

app/frontend/wallet/shelley/transaction/constants.ts
```diff
@@ -0,0 +1,5 @@
+// policyId is 28 bytes, assetName max 32 bytes, together with quantity makes
+// max token size about 70 bytes, max output size is 4000 => 4000 / 70 ~ 50
+export const MAX_OUTPUT_TOKENS = 50
+
+export const MIN_UTXO_VALUE = 1000000
```
refi93, line 5: Isn't this rather ADA_ONLY_MIN_UTXO_VALUE ?

app/frontend/wallet/shelley/transaction/computeTxPlan.ts
```diff
@@ -0,0 +1,327 @@
...
+  auxiliaryData: TxAuxiliaryData
+): TxPlanResult {
+  const totalRewards = withdrawals.reduce((acc, {rewards}) => acc + rewards, 0)
+  const totalInput = inputs.reduce((acc, input) => acc + input.coins, 0) + totalRewards
+  const totalInputTokens = aggregateTokenBundles(inputs.map(({tokenBundle}) => tokenBundle))
+  const deposit = computeRequiredDeposit(certificates)
+  const totalOutput = outputs.reduce((acc, {coins}) => acc + coins, 0) + deposit
+  const totalOutputTokens = aggregateTokenBundles(outputs.map(({tokenBundle}) => tokenBundle))
+
+  // total amount of lovelace that had to be added to token-containing outputs
```
refi93, line 32: perhaps a single bigger multiline comment at the beginning outlining the steps in human language here would help a lot in figuring out what's going on. The one-line comments scattered around the code help to some extent, but it's already too hard to get a bigger picture out of them - could you write such comment @PeterBenc ?

refi93 (approved):
LGTM, I'd just suggest adding a big comment outliningthe steps taken when computing the tx plan

PR #1028, opened 2021-06-08
Ledger transport selection

DavidTranDucVL:
Remove Connect with WebUSB button.
Add drop-down for transport: Default, U2F, WebUSB, WebHID


app/frontend/actions/wallet.ts
```diff
@@ -60,29 +60,38 @@ export default (store: Store) => {
...
+    const config = {
+      ...ADALITE_CONFIG,
+      isShelleyCompatible,
+      shouldExportPubKeyBulk,
+      ledgerTransportType,
+    }
     try {
+      loadingAction(
+        state,
+        'Please allow access to your device. Select your device in dropdown and then click Connect.'
```
refi93, line 84: this message doesn't apply for U2F where there isn't any such dropdown

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
```diff
@@ -0,0 +1,36 @@
...
+import {LedgerTransportType} from '../../../../frontend/types'
+import styles from './ledgerTransportationSelect.module.scss'
+
+interface Props {
+  onSelect: (ledgerTransportType: LedgerTransportType) => void
+}
+
+const LedgerTransportationSelect = ({onSelect}: Props) => {
+  const dropdownAssetItems = [
+    LedgerTransportType.DEFAULT,
```
refi93, line 12: I'd consider moving the "smart" transport selection logic to the app (or exposing the function returning the enum from ledger crypto provider, and we would actually know what that "default" transport is) - this would allow us to display appropriate guidance to the users based on the transport chosen

app/frontend/actions/wallet.ts
refi93: I'd consider rewording it/reformatting it as a two-line text:

```
Loading wallet data...`
If a prompt appears, click on the Ledger device, then click "Connect."
```
The instruction `Please allow access to your device.` is imho too ambiguous and in some instances the users would not need to confirm anything but the public key on the device, contributing to further confusion

package.json
```diff
@@ -101,6 +101,7 @@
     "redis": ">=3.1.2",
     "hosted-git-info": ">=3.0.8",
     "lodash": ">=4.17.21",
-    "ws": ">=7.4.6"
+    "ws": ">=7.4.6",
+    "glob-parent": ">=5.1.2"
```
refi93, line 105: I don't see this package being used anywhere - was it added by accident?

refi93: ahh it's just a resolution to address something reported by yarn audit - correct?

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
```diff
@@ -0,0 +1,36 @@
...
+      displaySelectedItemClassName={`input dropdown ${styles.dropdown}`}
+      items={dropdownAssetItems}
+      displayItem={displayItem}
+      onSelect={onSelect}
+      showSearch={false}
+    />
+  )
+}
+
+export default LedgerTransportationSelect
```
refi93, line 36: `LedgerTransportSelect` would be a more consistent name

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
```diff
@@ -0,0 +1,36 @@
...
+    LedgerTransportType.DEFAULT,
+    LedgerTransportType.U2F,
+    LedgerTransportType.WEB_USB,
+    LedgerTransportType.WEB_HID,
+  ]
+
+  const displayItem = (item: LedgerTransportType) => <div>{item}</div>
+
+  return (
+    <SearchableSelect
```
refi93, line 21: why a searchable select? I'd expect an ordinary one (or the name is misleading)

package.json
DavidTranDucVL: yes

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
DavidTranDucVL: Because we don't have "ordinary" one. We already have this one "searchable" which can be easily used without the search functionality.

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
DavidTranDucVL: Refactored

app/frontend/components/pages/login/ledgerTransportationSelect.tsx
DavidTranDucVL: Refactored.

app/frontend/actions/wallet.ts
DavidTranDucVL: Fixed

app/frontend/actions/wallet.ts
```diff
@@ -60,29 +62,48 @@ export default (store: Store) => {
     {
       cryptoProviderType,
       walletSecretDef,
-      forceWebUsb,
+      selectedLedgerTransportType,
```
refi93, line 65: the "selected" prefix seems redundant to me and reflects on a concern irrelevant to the crypto provider (that the transport has been "selected", but that doesn't really matter)

app/frontend/actions/wallet.ts
```diff
@@ -17,6 +17,8 @@ import loadingActions from './loading'
 import {saveAs} from '../libs/file-saver'
 import {exportWalletSecretDef} from '../wallet/keypass-json'
 
+import {getDefaultTransport} from '../../frontend/helpers/transports'
```
refi93, line 20: this applies only to ledger, so I'd name it "getDefaultLedgerTransportType" and consider keeping the function within the ledger crypto provider and exporting it from there (keeping the "ledger stuff" together), or having a dedicated helper file

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -231,10 +231,10 @@ const SendAdaPage = ({
     [updateAmount]
   )
 
-  const displayDropdownSelectedItem = ({assetName, policyId}: DropdownAssetItem) => (
+  const displayDropdownSelectedItem = ({assetName, policyId, type}: DropdownAssetItem) => (
     <div className="wrapper">
       {assetName}
-      {selectedAsset.type === AssetFamily.TOKEN && (
+      {type === AssetFamily.TOKEN && (
```
refi93, line 237: is this related to the PR or just an improvement aside? I'd cosider naming the parameter `assetFamily`, type is too generic to give meaningful information to the reader

package.json
```diff
@@ -101,6 +101,8 @@
     "redis": ">=3.1.2",
     "hosted-git-info": ">=3.0.8",
     "lodash": ">=4.17.21",
-    "ws": ">=7.4.6"
+    "ws": ">=7.4.6",
+    "glob-parent": ">=5.1.2",
+    "normalize-url": ">=4.5.1 <5.0.0"
```
refi93, line 106: do we need to restrict the upper version?

Btw there's a tool https://www.npmjs.com/package/yarn-audit-fix that patches the vulnerable dependencies right in yarn.lock file - doesn't always solve everything, but when it does, at least you don't have to "hack" package.json with resolutions which stay and may eventually result in issues related to the package version forced being too old

app/frontend/components/common/loadingOverlay.tsx
```diff
@@ -2,16 +2,24 @@ import {h} from 'preact'
 import {useSelector} from '../../helpers/connect'
 
 const LoadingOverlay = () => {
+  const toArray = (message: string | string[]) =>
+    message ? (Array.isArray(message) ? message : [message]) : ['']
+
   const {loading, loadingMessage} = useSelector((state) => ({
-    loadingMessage: state.loadingMessage,
+    loadingMessage: toArray(state.loadingMessage),
```
refi93, line 9: I'd consider not overcomplicating this with extending the message to be an array of lines (which is kind of unexpected from the name and creeps into the rest of the codebase in form of extended definitions) and I'd just add a functionality to parse newline characters in the string passed and split it into proper lines at presentation. 

The line list IMHOof makes the API more bloated than it needs to be in the vast majority of cases and a "proper" abstraction, i.e. "messageLines" instead of messages seems like an overkill for one special case. If we ever need to refactor this, it should be rather straightforward to figure out what messages are multiline by checking where the component is instantiated

https://stackoverflow.com/questions/35351706/how-to-render-a-multi-line-text-string-in-react

app/frontend/components/pages/login/hardwareAuth.tsx
```diff
@@ -80,40 +86,42 @@ const LoadByHardwareWalletSection = () => {
...
+              }
+            >
+              Unlock with
+              <div className="ledger-logo-container">
+                <LedgerLogoWhite />
+              </div>
+            </button>
+            <div className={styles.transportWrapper}>
+              <div className={styles.transportLabel}>
+                Transport
```
refi93, line 112: Regular users won't have a clue what "transport" is, I'd put in the UI "Method" instead

refi93 (changes requested):
few minor suggestions, otherwise LGTM

app/frontend/actions/wallet.ts
DavidTranDucVL: It's not passed to crypto provider as `selectedLedgerTransportType`. Variable is named `selectedLedgerTransportType` in the UI, where it reflect's what user selected, after being passed to wallet logic, it's "transformed" to `ledgerTransportType` where the `Default` is changed to specific transport type.

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: It's not related to the PR, just improvement aside. It was using variable outside of function scope before. `type` is already existing member, it's not something introduced in this change.

app/frontend/components/common/loadingOverlay.tsx
DavidTranDucVL: Refactored

app/frontend/components/pages/login/hardwareAuth.tsx
DavidTranDucVL: Fixed

package.json
DavidTranDucVL: I removed the upper bound

app/frontend/actions/wallet.ts
DavidTranDucVL: Refactored

refi93 (approved):

PR #1043, opened 2021-06-16
6.5.0

PeterBenc:


app/tests/src/wallet/account.js
```diff
@@ -89,6 +89,7 @@ describe('Tx plan', () => {
     it(`should create the right tx plan for tx with ${name}`, async () => {
       const account = await accounts.ShelleyAccount0
       const txPlanResult = await account.getTxPlan({...setting.args})
+      // console.log(JSON.stringify(txPlanResult))
```
PeterBenc, line 92: I noticed I left some commented stuff in here, :facepalm:, I ll make a PR which removes it so its removed in the next release.

refi93 (approved):

Issue #1046, opened 2021-06-19
Error when signing pool transaction


kingtutPool:
When I connect Ledger nano x to adalite and try to sign a pool certificate I get the following error.

[...]

PR #1047, opened 2021-06-21
Fix change when registering pool and add tests

PeterBenc:
Motivation: 
- refactoring of tx plan computing and splitting change outputs introduces a bud in pool registration, change was assigned null instead of `[]`

Changes:
- fix bug with change
- add tests for pool registration so we don't forget to change plan-computing logic there as well

Fixes: 
 https://github.com/vacuumlabs/adalite/issues/1046

### DIFF
--- a/app/tests/src/common/pool-reg-tx-settings.js
+++ b/app/tests/src/common/pool-reg-tx-settings.js
@@ -1,0 +1,176 @@
+export const poolRegTxSettings = {
+  'regular pool registration': {
+    unsignedTxParsed: {
+      inputs: [
+        {
+          txHash: Buffer.from(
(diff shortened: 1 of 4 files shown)


app/tests/src/common/pool-reg-tx-settings.js
```diff
@@ -0,0 +1,176 @@
+export const poolRegTxSettings = {
```
refi93, line 1: poolRegTxFixtures would imho look more natural as a name. "Settings" evokes some generic configuration transcending a single pool registration tx, which isn't the case here

refi93: but I see at the same time the "settings" suffix is widely used, and it's probably not worth it deviating/changing the previous instances if it is already a coined term

refi93 (approved):

PR #1075, opened 2021-07-19
Add token registry support

DavidTranDucVL:
Changes:
- Add server side call to token registry
- Add front-end side call to BE to retrieve registered tokens metadata
- Change star icon next to ADA to ADA symbol
- Remove hashes from asset drop-down selection
- Make all asset names consistent, asset name is always displayed with logo/icon
- Add question mark icon for non-registered tokens
- Add support to search tokens by ticker
- Update transaction history to match token registry metadata
- Add decimal places to digital assets, transaction history, confirmation, validation, inputs, summary
- Remove local state from SearchableSelect, so there is only one source of truth


.env.example
```diff
@@ -66,3 +66,5 @@ ADALITE_ENABLE_SEARCH_BY_TICKER=true
 # catalyst next round start and end dates in ISO format (example: 2021-2030)
 ADALITE_NEXT_VOTING_START=2021-01-01T00:00:00.000Z
 ADALITE_NEXT_VOTING_END=2030-01-01T00:00:00.000Z
+
+TOKEN_REGISTRY_URL="https://tokens.cardano.org"
```
refi93, line 70: the convention is to namespace "adalite" env variables with the "adalite" prefix to eliminate collisions with unrelated env variables - for this case, it may look a bit awkward but I'd rather stick to our convention

DavidTranDucVL: fixed

app/frontend/helpers/printAda.ts
```diff
@@ -1,4 +1,3 @@
-// eslint-disable-next-line no-unused-vars
 import {Lovelace} from '../types'
 export default (coins: Lovelace, decimals = 6): string => {
```
refi93, line 2: I stumbled upon this "decimals" parameter and I think it's kind of misleading as it suggests ADA can have a different amount of decimals than 6, but it is actually meant to be the "precision" to use to display ADA so I'd consider renaming the parameter to "precision" to avoid future confusion

app/frontend/tokenRegistry/tokenRegistryApi.ts
```diff
@@ -0,0 +1,37 @@
...
+  const getTokensMetadata = async (
+    subjects: string[]
+  ): Promise<{[subject: string]: RegisteredTokenMetadata}> => {
+    const url = `${ADALITE_CONFIG.ADALITE_SERVER_URL}/api/tokenRegistry/getTokensMetadata`
+    const requestBody = {subjects}
+    const response = await request(url, 'POST', JSON.stringify(requestBody), {
+      'Content-Type': 'application/json',
+    })
+    if (response?.Right) {
+      return response.Right.reduce((acc, tokenMetadata) => {
```
refi93, line 16: I'd split this into the "fetch" and "parse" part and add some basic unit test for parsing so we have actual examples of how the data returned from the server vs the representation in our code looks like, which could serve as a valuable reference when trying to understand what's going on

app/frontend/types.ts
```diff
@@ -78,6 +78,24 @@ export type Token = Omit<TokenObject, 'quantity'> & {
...
+export type RegisteredTokenMetadata = {
+  subject: string
+  description: string
+  ticker?: string
+  url?: string
+  logoHex?: string
+  decimals?: number
+}
+
+export type UnregisteredToken = Token & {
```
refi93, line 90: I think this type neither `RegisteredToken` isn't currently used, right?

server/tokenRegistrySubmitter.js
```diff
@@ -0,0 +1,49 @@
+require('isomorphic-fetch')
```
refi93, line 1: Is this really a "submitter"? From what I gather "proxy" would be a more accurate description

app/frontend/wallet/account.ts
```diff
@@ -283,10 +283,14 @@ const Account = ({
     return txWitness
   }
 
-  async function getMaxSendableAmount(address: Address, sendAmount: SendAmount) {
+  async function getMaxSendableAmount(
+    address: Address,
+    sendAmount: SendAmount,
+    decimals: number = 0
```
refi93, line 289: it seems a bit weird to me that wallet would need to be aware of the amount of decimals to compute the max sendable amount - I would expect it to deal with the "raw" integer amounts at all times and the frontend should be the one taking care of formatting the amount properly. I guess I'm missing some non-trivial assumption/dependency here

DavidTranDucVL: It's returning `SendAmount` type, which includes both: `quantity` as "raw" number and formatted `fieldValue` which is string. I didn't change underlying logic, just added necessary information so we can still return the formatted value as before.

If I remember correctly, we needed to lift formatted value from local state to global store for some reason that I don't remember. I can try refactor it back to the local state and remove formatting from actions, but I assume this isn't very high in priority right now.

app/frontend/helpers/printAda.ts
DavidTranDucVL: Refactored

app/frontend/tokenRegistry/tokenRegistryApi.ts
DavidTranDucVL: Refactored and added test

app/frontend/types.ts
DavidTranDucVL: I removed dead code

server/tokenRegistrySubmitter.js
DavidTranDucVL: Refactored

app/tests/src/tokenRegistryApi.js
```diff
@@ -0,0 +1,128 @@
...
+
+    const expected = {
+      '1d7f33bd23d85e1a25d87d86fac4f199c3197a2f7afeb662a0f34e1e776f726c646d6f62696c65746f6b656e': {
+        subject:
+          '1d7f33bd23d85e1a25d87d86fac4f199c3197a2f7afeb662a0f34e1e776f726c646d6f62696c65746f6b656e',
+        description:
+          'WMT is a utility token created to connect the unconnected and bank the unbanked. With nearly 4 billion people left offline, digital exclusion is a significant problem. Together with Input Output, World Mobile will deliver a revolutionary, scalable network in Africa and beyond. Let’s kickstart the sharing economy, reboot telecoms, and open a new world!',
+        ticker: 'WMT',
+        url: 'https://worldmobiletoken.com',
+        logoHex:
```
refi93, line 119: this seems rather like base64 to me :thinking: but I guess that's how the original API already returns it, right?

app/frontend/tokenRegistry/tokenRegistryApi.ts
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return acc
+      }, {})
+    } else {
+      return {}
+    }
+  }
+
+  const getTokensMetadata = async (
+    subjects: string[]
```
refi93, line 36: Given this function faces towards our wallet logic, I'd consider not exposing the raw "subject" in the function signature, but rather let it take our Token type and do the transformation to "subject" within the module, to not leak this internal representation of tokens towards higher level logic - the benefit is that the function signature would be less cryptic (no generic strings/custom one-off types) and the transformation logic would be contained/centralized within this module

app/frontend/tokenRegistry/tokenRegistryApi.ts
```diff
@@ -0,0 +1,46 @@
...
+          decimals: tokenMetadata?.decimals?.value,
+        }
+        return acc
+      }, {})
+    } else {
+      return {}
+    }
+  }
+
+  const getTokensMetadata = async (
```
refi93, line 35: If I understood correctly, this function we currently don't cache the token metadata in the fronted when we reload the wallet even though the tokens of the user didn't change. I'd consider wrapping the fetchTokensMetadata in some memorization to prevent such redundant re-fetches of data: https://lodash.com/docs/4.17.15#memoize

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -71,6 +72,18 @@ const ShelleyWallet = ({config, cryptoProvider}: WalletParams) => {
     )
   }
 
+  function getTokensMetadata(
+    accountInfo: Array<AccountInfo>
+  ): Promise<{[subject: string]: RegisteredTokenMetadata}> {
+    return TokenRegistryApi().getTokensMetadata([
```
refi93, line 78: this looks a bit artificial and makes one question why TokenRegistryApi isn't either a proper class with configurable parameters like the server URL, or a simple helper function that would be directly called.

I'd consider making a TokenRegistry (API being a redundant suffix IMHO) class that would take in its constructor the server URL (similarly as BlockchainExplorer does, but it isn't a class - pre-TS legacy) and would offer the getTokenMetada method. parseTokenMetadata could remain a standalone helper function used within the class which we would export from the module just for the sake of unit-testing

app/frontend/tokenRegistry/tokenRegistryApi.ts
```diff
@@ -0,0 +1,46 @@
+import request from '../wallet/helpers/request'
+import {ADALITE_CONFIG} from '../config'
+import {RegisteredTokenMetadata} from '../types'
+import {TokenRegistryApi} from './types'
+
+export default (): TokenRegistryApi => {
+  const fetchTokensMetadata = (subjects: string[]): Promise<any> => {
+    const url = `${ADALITE_CONFIG.ADALITE_SERVER_URL}/api/tokenRegistry/getTokensMetadata`
```
refi93, line 8: to be consistent with our adalite backend's bulk calls, I'd consider making the route `api/bulk/tokens/metadata`

server/tokenRegistryProxy.js
```diff
@@ -0,0 +1,46 @@
+require('isomorphic-fetch')
+const Sentry = require('@sentry/node')
+const {backendConfig} = require('./helpers/loadConfig')
+const mockData = require('./_tokenRegistryMock')
+const cachedFetch = require('./helpers/cachedFetch')
+
+const CACHE_TIMEOUT = 24 * 60 * 60 * 1000 // 1 day
```
refi93, line 7: Do we know the actual rate limits of the API? I tried to check https://github.com/cardano-foundation/cardano-token-registry/blob/master/API_Terms_of_Use.md but didn't find any clear information on this

package.json
```diff
@@ -59,6 +59,7 @@
     "express-ipfilter": "^1.1.2",
     "express-mung": "^0.5.1",
     "isomorphic-fetch": "^3.0.0",
+    "memory-cache": "^0.2.0",
```
refi93, line 62: https://github.com/vacuumlabs/adalite/blob/d1a9da4cfe87f0a6c6b1bbc50b93a029e493eb1e/app/frontend/wallet/blockchain-explorer.ts#L54

Isn't this simple helper we already have in our code do pretty much the same? If so, I'd rather not import new libraries if we don't really have to

app/frontend/actions/wallet.ts
```diff
@@ -170,12 +172,14 @@ export default (store: Store) => {
     loadingAction(state, 'Reloading wallet info...')
     try {
       const accountsInfo = await wallet.getAccountsInfo(state.validStakepoolDataProvider)
+      const tokensMetadata = await wallet.getTokensMetadata(accountsInfo)
```
refi93, line 175: what about tokens present in transaction history that are no longer part of the current balance of the account?

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -160,10 +165,12 @@ const SendAdaPage = ({
     type: AssetFamily.ADA,
     policyId: null,
     assetName: 'ADA',
-    assetNameHex: null,
     fingerprint: null,
     quantity: balance,
-    star: true,
+    searchable: {
```
refi93, line 170: specifying explicitly the searchable attributes to a dropdown item seems to me like coupling higher (search) and lower (token representation in dropdown) concerns, so I'd consider keeping these keys at the same level as the rest of keys and let the search logic decide which keys to filter by - what do you think?

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -45,23 +46,21 @@ const SendValidation = ({sendFormValidationError, txSuccessTab}) =>
     )
   )
 
+type Searchable = {
+  searchable: {
+    ticker: string
+    assetNameReadable: string
```
refi93, line 52: I'd consider distinguishing rather assetNameHex and assetNameUtf8 in the codebase to make the respective encodings clearer to the reader which is in the end what matters when working with the data and may otherwise not be that obvious

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -45,23 +46,21 @@ const SendValidation = ({sendFormValidationError, txSuccessTab}) =>
...
+  }
+}
+
 type DropdownAssetItem = Token & {
   fingerprint: string
-  assetNameHex: string
   type: AssetFamily
-  star?: boolean
-}
+} & Searchable
```
refi93, line 59: why not just include these searchable keys besides fingerprint and type?

DavidTranDucVL:
Reopening to redeploy

app/tests/src/tokenRegistryApi.js
DavidTranDucVL: Refactored

app/frontend/tokenRegistry/tokenRegistryApi.ts
DavidTranDucVL: Refactored

server/tokenRegistryProxy.js
DavidTranDucVL: I didn't find any information on rate limits as well.

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: It's already possible to search by fingerprint:
https://github.com/vacuumlabs/adalite/blob/add-token-registry-support/app/frontend/components/pages/sendAda/sendAdaPage.tsx#L214
I guess the Searchable type was confusing, as it was possible to search by other members as well. Initial idea was to introduce `searchable` object to hold additional information that wasn't there before and also they have no other purpose then search.

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: Refactored

app/frontend/actions/wallet.ts
DavidTranDucVL: Fixed, transaction history is now considered as well.

package.json
DavidTranDucVL: Fixed, I reused suggested code.

app/frontend/tokenRegistry/tokenRegistryApi.ts
DavidTranDucVL: Refactored

app/frontend/wallet/shelley-wallet.ts
DavidTranDucVL: Refactored

app/frontend/tokenRegistry/tokenRegistryApi.ts
DavidTranDucVL: Fixed

app/frontend/wallet/shelley-wallet.ts
```diff
@@ -71,6 +82,19 @@ const ShelleyWallet = ({config, cryptoProvider}: WalletParams) => {
     )
   }
 
+  function getTokensMetadata(
+    accountInfo: Array<AccountInfo>
+  ): Promise<{[subject: string]: RegisteredTokenMetadata}> {
+    return tokenRegistry.getTokensMetadata([
```
refi93, line 88: the token registry API is not verified to be stable (and we don't know their exact rate limit yet) so I'd rather not risk the app loading to fail if this non-critical piece of information is missing, so I'd add a fall back to catch the error 500 and just return an empty dictionary. The downside would be users wouldn't see their tokens "nicely" but the wallet would still load and they would be able to interact with it.

server/tokenRegistryProxy.js
refi93: I raised a question in the repo: https://github.com/cardano-foundation/cardano-token-registry/issues/291 - feel free to follow it

app/frontend/wallet/shelley-wallet.ts
DavidTranDucVL: Yes, it's handled here: https://github.com/vacuumlabs/adalite/blob/816389d42b0872f4b8e00fb524f27b68aed0b75b/app/frontend/tokenRegistry/tokenRegistry.ts#L40

app/frontend/tokenRegistry/tokenRegistry.ts
```diff
@@ -0,0 +1,52 @@
...
+      }, {})
+    } else {
+      return {}
+    }
+  }
+
+  public readonly getTokensMetadata = async (
+    tokens: Token[]
+  ): Promise<{[subject: string]: RegisteredTokenMetadata}> => {
+    const subjects = tokens.map(({policyId, assetName}) => `${policyId}${assetName}`)
```
refi93, line 48: I'd consider making `subject` a branded string, e.g. `TokenRegistrySubject` so it's easy to look up where it comes from in the code and so static checks can "scream" if somebody passes something not asserted to be a proper subject in the code

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -222,25 +228,42 @@ const SendAdaPage = ({
...
           token: {
             policyId: dropdownAssetItem.policyId,
-            assetName: dropdownAssetItem.assetNameHex,
-            quantity: parseFloat(fieldValue),
+            assetName: dropdownAssetItem.assetName,
+            // `tokenDecimals` would result in previously selected asset value,
+            // so we have to make a lookup for new decimals value
+            quantity: parseTokenAmount(
+              fieldValue,
+              tokensMetadata[`${dropdownAssetItem.policyId}${dropdownAssetItem.assetName}`]
```
refi93, line 236: I'd rather centralize the logic to transform token id/asset name to "subjects" so TS can keep track whether the the string is being formed properly - related to the previous comment I had regarding a branded type for "subject"

app/frontend/tokenRegistry/tokenRegistry.ts
DavidTranDucVL: Refactored

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: Refactored

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -330,7 +330,7 @@ const SendAdaPage = ({
       <label className="send-label">Asset</label>
       <SearchableSelect
         wrapperClassName="no-margin"
-        defaultItem={selectedAsset}
+        selectedItem={selectedAsset}
```
refi93, line 341: as far as I checked, `value` seems to be the prevalent name to refer to the current form element value, e.g. in input fields, so I'd consider naming this "value" for the sake of consistency of the general interface of form inputs. I don't see a discrepancy with `displaySelectedItem` nor ` displaySelectedItemClassName` as these refer to properties specific to the select component which indeed has a notion of "selected item"

refi93:
I checked against the testnet metadata server and the faucet token obtainable at https://testnets.cardano.org/en/testnets/cardano/tools/faucet/ doesn't seem to appear properly in the sense that I don't see its ticker in the UI even though it should be "TEST" from what I gather: https://github.com/input-output-hk/metadata-registry-testnet/blob/fe0f5ce9e1e55ccbbaf3ba6315e17e80476b395e/registry/6b8d07d69639e9413dd637a1a815a7323c69c86abbafb66dbfdb1aa7.json

Is that expected?

DavidTranDucVL:
> I checked against the testnet metadata server and the faucet token obtainable at https://testnets.cardano.org/en/testnets/cardano/tools/faucet/ doesn't seem to appear properly in the sense that I don't see its ticker in the UI even though it should be "TEST" from what I gather: https://github.com/input-output-hk/metadata-registry-testnet/blob/fe0f5ce9e1e55ccbbaf3ba6315e17e80476b395e/registry/6b8d07d69639e9413dd637a1a815a7323c69c86abbafb66dbfdb1aa7.json
> 
> Is that expected?

Fixed

Issue #1077, opened 2021-07-20
Prioritize in utxo selection utxos that contain the token to be sent


refi93:
Problem: it seems we currently don't prioritize utxos containing the token to be sent when selecting them into the transaction, resulting in the behavior, that a token transaction contains a bunch of unrelated token inputs, which ultimately bloats transactions dealing with tokens

https://github.com/vacuumlabs/adalite/blob/307cbce574c7d061eac856383f4def13d18d152c/app/frontend/wallet/account.ts#L303-L308

PeterBenc:
already fixed

PR #1078, opened 2021-07-20
Optimize utxo selection

PeterBenc:
Motivation : #1077 

Changes: 
 - if a user is sending tokens, utxos that contain this token are selected first, then onlyAda utxos, and then the rest of the token utxos.
 
 Gotcha: 
  - I need to do some manual testing but checkout the tests

### DIFF
--- a/app/frontend/wallet/account.ts
+++ b/app/frontend/wallet/account.ts
@@ -13,6 +13,7 @@ import {
   TxType,
   Address,
   PoolOwnerTxPlanArgs,
+  AssetFamily,
 } from '../types'
 import {
   getAccountXpub as getAccoutXpubShelley,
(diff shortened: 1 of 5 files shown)


app/frontend/wallet/account.ts
```diff
@@ -294,30 +295,43 @@ const Account = ({
...
     const randomGenerator = PseudoRandom(seeds.randomInputSeed)
-    // TODO: ideally we want to do nonStaking, with token, only ada, the rest
-    if (txType === TxType.SEND_ADA || txType === TxType.CONVERT_LEGACY) {
+
+    if (
+      txPlanArgs.txType === TxType.SEND_ADA &&
+      txPlanArgs.sendAmount.assetFamily === AssetFamily.TOKEN
+    ) {
+      const {policyId, assetName} = txPlanArgs.sendAmount.token
+      const utxosWithToken = tokenUtxos.filter(({tokenBundle}) =>
```
refi93, line 310: This naming is possibly confusing as both the utxos "with" and "without" the token actually contain some token. Therefore I'd name this `utxosWithTargetToken` and `utxosWithoutTargetToken`

app/tests/src/wallet/account/utxo-settings.js
```diff
@@ -0,0 +1,80 @@
...
+    availableUtxos: [utxos.withTokens1, utxos.adaOnly],
+    selectedUtxos: [utxos.adaOnly, utxos.withTokens1],
+  },
+  'sending tokens': {
+    txPlanArgs: {
+      address:
+        'addr1qjag9rgwe04haycr283datdrjv3mlttalc2waz34xcct0g4uvf6gdg3dpwrsne4uqng3y47ugp2pp5dvuq0jqlperwj83r4pwxvwuxsgds90s0',
+      sendAmount: {
+        assetFamily: AssetFamily.TOKEN,
+        fieldValue: `${2}`,
```
refi93, line 68: why do we wrap these numbers into template strings? Also I'm curious - why is the fieldValue included among txPlanArgs? I guess it's some leftover from data passed from the UI without actual purpose, right?

app/frontend/wallet/account.ts
```diff
@@ -294,30 +295,43 @@ const Account = ({
...
+        tokenBundle.some((token) => token.policyId === policyId && token.assetName === assetName)
+      )
+      const utxosWithoutToken = tokenUtxos.filter(
+        ({tokenBundle}) =>
+          !tokenBundle.some((token) => token.policyId === policyId && token.assetName === assetName)
+      )
       return [
+        ...shuffleArray(utxosWithToken, randomGenerator),
```
refi93, line 318: general note - shuffling the individual "subcategories" of tokens has no benefit for the purpose of utxo "self-organization" as opposed to shuffling all the tokens in the beginning and then filtering the subcategories (provided the filtering respects the original order of elements which it does)

app/tests/src/wallet/account/utxo-settings.js
PeterBenc: I guess it's just a leftover, basically, the input has a field value, which is what is written in the input, (string), and a "normal" value which is of a branded type (Lovelace), but yes I guess this could be removed/refactored.

app/frontend/wallet/account.ts
PeterBenc: agreed, thanks :+1:

app/frontend/wallet/account.ts
PeterBenc: Hm, you are right, given that the filtering is respective to the shuffling, it should be enough to shuffle them all in the beginning. It didn't break any of the tests so I remove the shuffling of the filtered utxos.

app/tests/src/wallet/account/utxo-settings.js
```diff
@@ -0,0 +1,80 @@
...
+        token: {
+          policyId: 'ca37dd6b151b6a1d023ecbd22d7e881d814b0c58a3a3148b42b865a0',
+          assetName: '7365636f6e646173736574',
+          quantity: 2,
+        },
+      },
+      txType: TxType.SEND_ADA,
+    },
+    availableUtxos: [utxos.adaOnly, utxos.withTokens2, utxos.legacy, utxos.withTokens1],
+    selectedUtxos: [utxos.withTokens1, utxos.legacy, utxos.adaOnly, utxos.withTokens2],
```
refi93, line 78: I was about to write that this test leaves space for doubt because shuffling can interfere with its result and I realized there's another layer of confusion to this, as selectUtxos function doesn't really "select" them, just shuffles them in a specific way.

The latter can be addressed by renaming the function to let's say `arrangeUtxos` so it's clear that it isn't really filtering out any. The former is a bit trickier and for that I'd consider adding a flag "randomShuffle" to the arrangeUtxos function which would be on by default, but could be disabled for the sake of tests. The current solution where we have some preset seed in the tests is really fragile

PeterBenc: Agreed, I will rename it to `arrangeUtxos`, regarding the second point, I agree as well, the fact we have to manage the randomness in tests is a pain and for that reason, I made tests with only one utxo of each kind. 

If you agree, I ll create an issue for the removal of the randomness altogether, as you said, if they are not prioritized anywhere, they are basically shuffled.

refi93: agree, let's make the follow up issue

refi93 (approved):

PR #1085, opened 2021-07-30
Dashboard header message

PeterBenc:


refi93 (approved):
LGTM only minor naming remarks, also inputs in send form should be disabled while tx is being submitted

PR #1107, opened 2021-08-30
Reintroduce token registry

DavidTranDucVL:
Changes
- Increase max size for requests to express server
  - I increased the limit from default 100kB to 256kB, I'm not sure where we want to set the limit.
- Add polyfills to fix crashing of ES6+ code in old browsers
- If user has too many tokens, we load wallet regardless and ignore token metadata

Notes
- You can test the wallet with regular ledger seed and `abcd` passphrase


app/frontend/tokenRegistry/tokenRegistry.ts
```diff
@@ -18,11 +18,15 @@ export class TokenRegistry {
       : this._fetchTokensMetadata
   }
 
-  private readonly _fetchTokensMetadata = (subjects: string[]): Promise<any> => {
+  private readonly _fetchTokensMetadata = async (subjects: string[]): Promise<any> => {
```
refi93, line 21: I'd make the return type explicit

refi93 (approved):

PR #1117, opened 2021-09-06
Refactor testnet networks

xdzurman:
TODO on merge: Change heroku env var from MARY_TESTNET to TESTNET!


app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -312,7 +312,12 @@ function ShelleySignedTransactionStructured(
   }
 
   function encodeCBOR(encoder) {
-    return encoder.pushAny([txAux, txWitnesses, txAuxiliaryData])
+    return encoder.pushAny([
+      txAux,
+      txWitnesses,
+      true, // fixed alonzo "isValid" Validation tag
```
refi93, line 318: Just curious - I thought this is not part of the submitted transaction but it's rather computed on-chain after the transaction is submitted - what would be the point in putting here false?

xdzurman:
```
 the IsValid tag, which indicates whether all scripts that are executed during the script
execution phase validate. The correctness of the tag is verified as part of the ledger rules,
and the block is deemed to be invalid if it is applied incorrectly. If it is set to False, then the
block can be re-applied without script re-validation. This tag cannot be signed, since it is
applied by the block producer.
```

refi93:
what address can I test this with? Where is the faucet for this testnet?

refi93 (approved):
LGTM, please squash redundant commits and I'd merge it to develop so it's ready and we'd already have this way a mainnet staging environment (adalite-staging.herokuapp.com) to validate the deployment with at the moment of the HF

refi93:
turns out the changes to the serialized transaction format (the additional isValidating "true" flag) weren't needed so I'm "un-approving" this PR to get other related changes done instead

refi93 (approved):

PR #1128, opened 2021-09-09
Optimize token registry metadata processing

DavidTranDucVL:
Changes
- Load token metadata in non-blocking matter. After wallet is loaded we would show tokens without linked metadata until the request fetches data from remote service and then update UI accordingly.
- Change request to token registry to bulk request.
  - Maximum amount of tokens per request arbitrarily set to 100.
- Cache entries on level of token metadata, instead of functions calls.
- Set maximum limit for wallet for which we support token metadata, currently set arbitrary to 2000.

### DIFF
--- a/app/frontend/actions/wallet.ts
+++ b/app/frontend/actions/wallet.ts
@@ -47,12 +47,12 @@ export default (store: Store) => {
 
   const loadAsyncWalletData = async (): Promise<void> => {
     const asyncFetchAndUpdate = async <T extends Partial<State>>(
-      fetch: () => Promise<T>,
+      fetchFn: () => Promise<T>,
       fallbackValue: T,
       debugMessage: string = null
     ): Promise<void> => {
       try {
-        setState(await fetch())
+        setState(await fetchFn())
       } catch (e) {
         debugMessage ?? debugLog(debugMessage)
         setState(fallbackValue)
--- a/app/frontend/actions/wallet.ts
+++ b/app/frontend/actions/wallet.ts
@@ -20,20 +20,6 @@ import {exportWalletSecretDef} from '../wallet/keypass-json'
 import {getDefaultLedgerTransportType} from '../wallet/shelley/helpers/transports'
 import sleep from '../helpers/sleep'
 
-// TODO: (refactor), this should not call "setState" as it is not action
-const fetchConversionRates = async (conversionRates, setState) => {
-  try {
-    setState({
-      conversionRates: await conversionRates,
-    })
-  } catch (e) {
-    debugLog('Could not fetch conversion rates.')
-    setState({
-      conversionRates: null,
-    })
-  }
-}
-
 // TODO: we may be able to remove this, kept for backwards compatibility
 const getShouldShowSaturatedBanner = (accountsInfo: Array<AccountInfo>) =>
   accountsInfo.some(({poolRecommendation}) => poolRecommendation.shouldShowSaturatedBanner)
@@ -56,8 +42,36 @@ const accountsIncludeStakingAddresses = (
 export default (store: Store) => {
   const {loadingAction} = loadingActions(store)
   const {setError} = errorActions(store)
-  const {setState} = store
+  const {getState, setState} = store
   const {setWalletOperationStatusType} = commonActions(store)
+
+  const loadAsyncWalletData = async (): Promise<void> => {
+    const asyncFetchAndUpdate = async <T extends Partial<State>>(
+      fetch: () => Promise<T>,
+      fallbackValue: T,
+      debugMessage: string = null
+    ): Promise<void> => {
+      try {
+        setState(await fetch())
+      } catch (e) {
+        debugMessage ?? debugLog(debugMessage)
+        setState(fallbackValue)
+      }
+    }
+
+    await Promise.all([
+      asyncFetchAndUpdate(
+        async () => ({conversionRates: await getConversionRates(getState())}),
+        {conversionRates: null},
+        'Could not fetch conversion rates.'
+      ),
+      asyncFetchAndUpdate(
+        async () => ({tokensMetadata: await wallet.getTokensMetadata(getState().accountsInfo)}),
+        {tokensMetadata: new Map()},
+        'Could not fetch tokens metadata.'
+      ),
+    ])
+  }
 
   const loadWallet = async (
     state: State,
@@ -116,10 +130,8 @@ export default (store: Store) => {
 
       const validStakepoolDataProvider = await wallet.getStakepoolDataProvider()
       const accountsInfo = await wallet.getAccountsInfo(validStakepoolDataProvider)
-      const tokensMetadata = await wallet.getTokensMetadata(accountsInfo)
       const shouldShowSaturatedBanner = getShouldShowSaturatedBanner(accountsInfo)
 
-      const conversionRatesPromise = getConversionRates(state)
       const usingHwWallet = wallet.isHwWallet()
       const maxAccountIndex = wallet.getMaxAccountIndex()
       const shouldShowWantedAddressesModal = accountsIncludeStakingAddresses(
@@ -135,7 +147,6 @@ export default (store: Store) => {
       setState({
         validStakepoolDataProvider,
         accountsInfo,
-        tokensMetadata,
         maxAccountIndex,
         shouldShowSaturatedBanner,
         walletIsLoaded: true,
@@ -157,7 +168,7 @@ export default (store: Store) => {
         // shelley
         isShelleyCompatible,
       })
-      await fetchConversionRates(conversionRatesPromise, setState)
+      loadAsyncWalletData()
     } catch (e) {
       setState({
         loading: false,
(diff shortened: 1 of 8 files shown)


app/frontend/tokenRegistry/tokenRegistry.ts
```diff
@@ -22,6 +24,12 @@ export class TokenRegistry {
   private readonly _fetchTokensMetadata = async (
     subjects: string[]
   ): Promise<TokenMetadataResponse> => {
+    if (subjects.length > MAX_REQUEST_SIZE) {
```
refi93, line 27: MAX_REQUEST_SIZE can be mistook for the byte size, something like MAX_SUBJECTS_COUNT would be less ambiguous

server/tokenRegistryProxy.js
```diff
@@ -1,29 +1,95 @@
...
+      const subjects = req.body.subjects
+
+      if (subjects.length > MAX_REQUEST_SIZE) {
+        return res.json({
+          statusCode: 400,
+          Left: 'Request over max limit',
+        })
+      }
+
+      // Retrieve subjects that are cached form cache and determine
```
refi93, line 46:
```suggestion
      // Retrieve subjects that are cached from cache and determine
```

app/frontend/types.ts
```diff
@@ -415,3 +415,11 @@ export type WalletOperationStatusType =
   | 'txSuccess'
   | 'txFailed'
   | 'reloadFailed'
+
+export type ConvectionRates = {
```
refi93, line 419:
```suggestion
export type ConvertionRates = {
```

server/tokenRegistryProxy.js
```diff
@@ -1,29 +1,95 @@
 require('isomorphic-fetch')
 const Sentry = require('@sentry/node')
 const {backendConfig} = require('./helpers/loadConfig')
-const cacheResults = require('./helpers/cacheResults')
+const chunk = require('./helpers/chunk')
+const Cache = require('./helpers/cache')
 
-const CACHE_TIMEOUT = 24 * 60 * 60 * 1000 // 1 day
+const REQUEST_CHUNK_SIZE = 100
```
refi93, line 7: SUBJECTS_PER_REQUEST would imho be a more descriptive name

server/tokenRegistryProxy.js
```diff
@@ -1,29 +1,95 @@
...
+      if (subjects.length > MAX_REQUEST_SIZE) {
+        return res.json({
+          statusCode: 400,
+          Left: 'Request over max limit',
+        })
+      }
+
+      // Retrieve subjects that are cached form cache and determine
+      // which subjects need to be fetched from remote
+      const {cached: cachedTokensMetadata, toFetch} = subjects.reduce(
```
refi93, line 48: I'd consider splitting this into two separare filter+map transformations which could even be one-liners so it's easier to follow:

```javascript
const cached = subjects.filter((s) => cache.get(s) !== null).map((s) => cache.get(s))
const toFetch = subjects.filter((s) => cache.get(s) === undefined)
```

server/helpers/cache.js
```diff
@@ -0,0 +1,38 @@
...
+    if (this.get(key) === undefined) this.size++
+    this.cache[key] = {
+      timestamp: Date.now(),
+      data: value,
+    }
+    if (this.size > this.limit) this.prune()
+  }
+
+  prune() {
+    const pruneSize = Math.round(PRUNE_RATIO * this.limit)
```
refi93, line 29: why not use a native Map? https://stackoverflow.com/questions/18541940/map-vs-object-in-javascript

A Map preserves the order of insertion, so you wouldn't even have to sort the keys, though arguably, last access time to the key is more relevant than the insertion time for the sake of cache pruning

server/helpers/cache.js
```diff
@@ -0,0 +1,38 @@
+const TIMEOUT = 24 * 60 * 60 * 1000 // 1 day
+const PRUNE_RATIO = 0.1
+
+class Cache {
+  constructor(limit, maxAge = TIMEOUT) {
+    this.limit = limit
+    this.maxAge = maxAge
+    this.cache = {}
+    this.size = 0
```
refi93, line 9: why do we have to keep the size explicitly? Were you concerned about Object.keys().length being inefficient?

server/helpers/cache.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  get(key) {
+    if (key in this.cache && this.cache[key].timestamp + this.maxAge > Date.now()) {
```
refi93, line 13: I'd consider deleting the stale keys right away and then you could split the concern of enforcing the record max age and accessing non-stale records of the cache

server/helpers/cache.js
```diff
@@ -0,0 +1,38 @@
...
+  set(key, value) {
+    if (this.get(key) === undefined) this.size++
+    this.cache[key] = {
+      timestamp: Date.now(),
+      data: value,
+    }
+    if (this.size > this.limit) this.prune()
+  }
+
+  prune() {
```
refi93, line 28: I guess this method isn't really supposed to be called from outside so I'd underscore it (declaring it "private" would be ideal, but we don't have TS server side :/)

server/tokenRegistryProxy.js
```diff
@@ -1,29 +1,95 @@
...
+        (acc, subject) => {
+          const cachedValue = cache.get(subject)
+          if (cachedValue === undefined) {
```
refi93, line 51: if the cache had a proper "has()" method, these checks for undefined could be rewritten to use it, making for an easier to follow code, removing the hassle around null vs undefined

refi93 (changes requested):
generally looks good, I left suggestions to improve readability and I think a significant part of the "heavy lifting" in the cache can be removed if you represent it via Map instead of an ordinary object

app/frontend/actions/wallet.ts
```diff
@@ -56,9 +42,37 @@ const accountsIncludeStakingAddresses = (
 export default (store: Store) => {
   const {loadingAction} = loadingActions(store)
   const {setError} = errorActions(store)
-  const {setState} = store
+  const {getState, setState} = store
   const {setWalletOperationStatusType} = commonActions(store)
 
+  const loadAsyncWalletData = async (): Promise<void> => {
+    const asyncFetchAndUpdate = async <T extends Partial<State>>(
+      fetch: () => Promise<T>,
```
refi93, line 50: fetchFn would imho be better, to not confuse it with native "fetch" function in js

app/frontend/tokenRegistry/tokenRegistry.ts
DavidTranDucVL: Refactored

server/tokenRegistryProxy.js
DavidTranDucVL: Refactored

app/frontend/types.ts
DavidTranDucVL: Refactored

server/tokenRegistryProxy.js
DavidTranDucVL: Refactored

server/tokenRegistryProxy.js
DavidTranDucVL: Refactored

server/helpers/cache.js
DavidTranDucVL: I wasn't aware that map preserves the order, refactored.

server/helpers/cache.js
DavidTranDucVL: Yes, from what I found, `Object.keys().length` is O(n)

server/helpers/cache.js
DavidTranDucVL: Not sure if I understood this comment correctly. I added deletion of old entries right away.

server/helpers/cache.js
DavidTranDucVL: Refactored

server/tokenRegistryProxy.js
DavidTranDucVL: Refactored

app/frontend/actions/wallet.ts
DavidTranDucVL: Refactored

server/helpers/cache.js
refi93: I'd still leave here a comment clarifying that the order of insertion is preserved so an unaware reader isn't left wondering based on what do we actually prune the keys

server/tokenRegistryProxy.js
```diff
@@ -1,29 +1,85 @@
...
+          Left: 'Request over max limit',
+        })
+      }
+
+      // Retrieve subjects that are cached from cache and determine
+      // which subjects need to be fetched from remote
+      const toFetch = subjects.filter((s) => !cache.has(s))
+
+      // filter both - undefined (to fetch) and null (no metadata) entries
+      const cachedTokensMetadata = subjects.map((s) => cache.get(s)).filter((s) => s != null)
```
refi93, line 49: won't the `.filter((s) => s != null)` result in items in the null cache being ignored? I think the correct transformation should look like this:

```javascript
subjects.filter((s) => cache.has(s)).map((s) => cache.get(s))
```

refi93 (changes requested):
see further comments - should be rather small things, high level LGTM

server/tokenRegistryProxy.js
DavidTranDucVL: Yes that's the point, otherwise it would fill the array with null values, like `[null, null, ..., null]`. Current logic expects that subjects that don't have associated metadata are just not returned. Token registry service behaves in the same way.

refi93: ah I see - I'd name it then cachedNonNullTokensMetadata otherwise as a reader I may be confused why the "null" metadata are not considered here and may genuinely believe it's a bug

refi93 (approved):

server/helpers/cache.js
refi93: yes, that's what I meant :+1:

server/helpers/cache.js
DavidTranDucVL: Fixed

server/tokenRegistryProxy.js
DavidTranDucVL: Fixed

PR #1131, opened 2021-09-13
Reward warning, filtering of failed txs, news

xdzurman:
image

image

### DIFF
--- a/app/frontend/components/common/infoModal.tsx
+++ b/app/frontend/components/common/infoModal.tsx
@@ -59,11 +59,24 @@ class InfoModal extends Component<Props, {dontShowAgainCheckbox: boolean; should
           <div className="welcome-body">
             <h2 className="welcome-title">AdaLite News</h2>
             <NewsSection
+              date={'09/13/2021'}
+              children={
+                <Fragment>
+                  <Article title="Possibly missing rewards in staking history" icon="">
+                    One of the cardano components we use may be missing some of the rewards. For
+                    this reason, staking history and CSV exports may be wrong. However, your staking
+                    balance is unaffected by this issue and you can freely withdraw your correct
+                    amount of rewards.
+                  </Article>
+                </Fragment>
+              }
+            />
+            <NewsSection
               date={'09/07/2021'}
               children={
                 <Fragment>
                   <Article title="Trezor signing planned outage" icon="">
-                    On September 8th, between 14:00 and 15:00 UTC, signing transactions with Trezor
+                    On September 15th, between 14:00 and 15:00 UTC, signing transactions with Trezor
                     hardware wallets is expected to fail while we update Adalite to support the
                     upcoming Trezor firmware.
                   </Article>
--- a/app/frontend/components/common/infoModal.tsx
+++ b/app/frontend/components/common/infoModal.tsx
@@ -72,7 +72,7 @@ class InfoModal extends Component<Props, {dontShowAgainCheckbox: boolean; should
               }
             />
             <NewsSection
-              date={'09/07/2021'}
+              date={'09/13/2021'}
               children={
                 <Fragment>
                   <Article title="Trezor signing planned outage" icon="">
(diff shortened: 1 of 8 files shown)


app/frontend/components/pages/delegations/stakingHistoryPage.tsx
```diff
@@ -190,6 +191,11 @@ class StakingHistoryPage extends Component<Props> {
     return (
       <div className="staking-history card">
         <h2 className="card-title">Staking and Rewards History</h2>
+        <div className="staking-history-warning">
+          <Alert alertType="warning">
+            Temporarily, some rewards may be missing in the history.
```
refi93, line 196:
```suggestion
            Some rewards may be missing in the history.
```

app/frontend/components/pages/txHistory/transactionHistory.tsx
```diff
@@ -244,6 +245,11 @@ const TransactionHistory = (): h.JSX.Element => {
           <ExportCSV transactionHistory={transactionHistory} stakingHistory={stakingHistory} />
         </div>
       </div>
+      <div className="staking-history-warning">
+        <Alert alertType="warning">
+          CSV exports might be wrong due to possibly missing rewards.
```
refi93, line 250:
```suggestion
          CSV exports might be incorrect due to possibly missing rewards.
```

app/frontend/components/common/infoModal.tsx
```diff
@@ -58,12 +58,25 @@ class InfoModal extends Component<Props, {dontShowAgainCheckbox: boolean; should
...
+                    One of the cardano components we use may be missing some of the rewards. For
+                    this reason, staking history and CSV exports may be wrong. However, your staking
+                    balance is unaffected by this issue and you can freely withdraw your correct
+                    amount of rewards.
+                  </Article>
+                </Fragment>
+              }
+            />
             <NewsSection
               date={'09/07/2021'}
```
refi93, line 75: I'd bump also the date of this message so its clear that it's a recent update

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -169,13 +170,19 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         return await _getAddressInfos(addresses.slice(beginIndex, beginIndex + gapLimit))
       })
     )
-    const addressTokenBundles = txHistory.map((txHistoryEntry) => {
+    const validTxHistory = txHistory.map((txHistoryEntry) => {
```
refi93, line 173: we use just the balance from the address info response, so if I'm not mistaken there's no reason to filter the tx history as we don't do anything with it here

I think that the names are just misleading and what we call "txHistory" here should rather be "addressInfos"

Here's a sample response I looked up:

```
{"Right":{"caAddresses":[<addresses requested>],"caTxNum":248,"caBalance":{"getCoin":"1671509","getTokens":[]},"caTxList":[<bunch of txs we don't really care about here>]}}
```

refi93: If we should drop invalid txs somewhere, it should be in `getTxHistory()` if I'm not mistaken

refi93: so I think we don't have to change anything here, though the variables probably deserve renaming to make it clear we care about overall address infos and tx history is thrown away

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -100,8 +101,8 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
       if (!tx.ctbId) captureMessage(`Tx without hash: ${JSON.stringify(tx)}`)
       return prepareTxHistoryEntry(tx, addresses)
     })
-
-    return txHistoryEntries.sort((a, b) => b.ctbTimeIssued - a.ctbTimeIssued)
+    const validTransactions = filterValidTransactions(txHistoryEntries)
```
refi93, line 104: I'd add a TODO comment along the lines of `// TODO temporary workaround until collateral inputs in adalite-backend are exposed to properly show effect of invalid transactions in UI/csv export`

refi93 (changes requested):
see comments - one minor remark regarding news date bump and I believe there's redundant filtering happening in the call to fetch wallet balance

refi93 (approved):

PR #1135, opened 2021-09-14
Disable tx creation while catalyst registration is being submitted

refi93:
As the submit tx UI has been updated to be non-blocking, we probably forgot to update the catalyst registration screen, which now shows to the user the "Confirm transaction" button as non-disabled while the previously confirmed tx is being submitted, which can lead to unpredictable behavior.

This PR is fixing the issue above and disables the button while the wallet is not ready to submit further txs

Other fixes:
* disable closing of the tx confirmation modal if user clicks outside by introducing a configurable property of the common modal element
* if user rejected the transaction while signing, UI got stuck on waiting for ledger - apparently `stopLoadingAction` call was missing in the "finally" block of the tx signing/submission handler

It can be tested with "odor match ..." mnemonic or on Ledger with abandon... mnemonic

Scenarios covered:
* clicking outside voting/tx confirmation modal no longer closes it
* if tx rejected on Ledger, error modal appears and "waiting for Ledger" disappears shortly thereafter
* "confirm transaction" no longer clickable while the transaction is being submitted

Known issue:
* probably not a big deal but the user can still close the modal/cancel the transaction from the UI and continue signing the transaction on Ledger and it gets submitted. We would have to be able to somehow abort the transaction or disable closing the modals while tx is being confirmed on hw wallet, but IMHO it's an overkill and I would expect a reasonable user to reject the transaction on the device as well

### DIFF
--- a/app/frontend/actions/transaction.ts
+++ b/app/frontend/actions/transaction.ts
@@ -244,8 +244,8 @@ export default (store: Store) => {
         error: e,
       })
       setState({
-        shouldShowTransactionErrorModal: true,
         shouldShowVotingDialog: false,
+        shouldShowTransactionErrorModal: true,
       })
 
       closeConfirmationDialog(state)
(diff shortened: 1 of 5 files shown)


PeterBenc (approved):
LGTM, tested, thanks for fixing this.

PR #1143, opened 2021-09-20
Refactor IP extraction to account for firewall

refi93:
Since we setup an application firewall, it appends its IPs to the "fwd" header, skewing our metrics based around IP addresses. This PR fixes the call to take only the first IP (the on of the client itself) from the respective header

More context: https://serverfault.com/questions/846489/can-x-forwarded-for-contain-multiple-ips

Testing: included temporary logging to see in logs if ips are being extracted properly once deployed, is to be reverted afterwards


PeterBenc (approved):
LGTM, thanks for fixing this

PR #1148, opened 2021-09-24
Invalidate browser cache after new version release

refi93:
Ensures that users get latest js as soon as we make a new release of the app. A bit hacky, but given the effort it would take to rework serving the initial page (injecting the content hashes into HTML, meaning it would have to be done by webpack or keep it in expresses but with even uglier hacks) I think it's a reasonable workaround and at least I tried to clarify the rationale with comments


PeterBenc (approved):
Thanks a lot for fixing this.

PR #1157, opened 2021-09-30
Add 30 second block time between refreshes

xdzurman:


PeterBenc (approved):

app/frontend/actions/wallet.ts
```diff
@@ -182,18 +183,25 @@ export default (store: Store) => {
     return true
   }
 
-  const reloadWalletInfo = async (state: State) => {
+  let lastReloadTime = new Date()
+  const reloadWalletInfo = async (state: State): Promise<void> => {
```
PeterBenc, line 187: I would give this function a parameter for overriding the time condition. I think we should reload wallet info after transaction submission no matter what. In case a user reloads wallet and then sends tx under 30 seconds his balances and tx history would not update.

app/frontend/components/pages/accounts/accountsDashboard.tsx
```diff
@@ -127,7 +127,7 @@ const AccountsDashboard = () => {
             <button
               className="button secondary balance refresh"
               disabled={shouldDisableSendingButton(walletOperationStatusType)}
-              onClick={reloadWalletInfo}
+              onClick={() => reloadWalletInfo(true)}
```
refi93, line 130: I think it would be more sensible to limit the debouncing to the reload button, as @PeterBenc pointed out, it could otherwise impact other parts of the app that need reloading, and managing that can get unnecessarily complex, or did you have some strong reason to implement it at "reloadWalletInfo" level?

If not, I guess something like this https://www.npmjs.com/package/debounce (let's port the code to our codebase if we don't have an equivalent) should work and the idea is that the onClick handler would have an effect only once every 30 seconds

refi93: The implication is that every refresh button we have would have a separate debouncing though. An alternative could be to introduce a decorated "debouncedReloadWalletInfo" which would be a wrapper over reloadWalletInfo with debouncing added which would apply globally for all its invocations, and I think it would be cleaner than having a flag for that as it would be clear from the name what it's doing.

app/frontend/actions/wallet.ts
```diff
@@ -208,6 +210,19 @@ export default (store: Store) => {
     }
   }
 
+  let lastReloadTime = new Date()
```
refi93, line 213: In theory, lastReloadTime should probably be bumped by `reloadWalletInfo()` and it should be initially set to some value indicating the wallet wasn't reloaded yet ever, like `new Date(0)`  - this way we would ensure that the wallet as such gets reloaded every 30 seconds even if other parts of the app use the non-debounced version of the function, or did you intentionally bump it here?

refi93: and I'd name the variable `_lastWalletReloadTime` to make it a bit more explicit

xdzurman: -> discussed offline

refi93 (approved):
LGTM, one last remark - I think it's sensible to reset the last reloading time if the loading fails by any chance so that users don't have to wait 30 seconds to get a new state, or even get misled by the fake reload that the reloading was successful which could lead to pretty big frustration in the end

PR #1160, opened 2021-10-04
Run yarn-audit-fix and add resolutions

xdzurman:
Fixes audit errors

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -44,6 +44,7 @@
     "sinon": "^7.3.2"
   },
   "resolutions": {
-    "lodash": ">=4.17.21"
+    "lodash": ">=4.17.21",
+    "ansi-regex": "<6.0.0"
   }
 }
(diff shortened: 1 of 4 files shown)


app/package.json
```diff
@@ -44,6 +44,7 @@
     "sinon": "^7.3.2"
   },
   "resolutions": {
-    "lodash": ">=4.17.21"
+    "lodash": ">=4.17.21",
+    "ansi-regex": "<6.0.0"
```
refi93, line 48: shouldn't this rather be 
```suggestion
    "ansi-regex": ">=5.0.1"
```
?

xdzurman: They rollbacked it to https://github.com/chalk/ansi-regex/releases/tag/v5.0.1 because v 6.0.1 fails with webpack module errors. However, yarn lock automatically goes for v6.0.1 if you specify >=5.0.1 which breaks it.

xdzurman: Should I rather add two entries there? One for >= and the other one for <=? I did not want to specify exact match : "5.0.1."

refi93 (approved):

PR #1161, opened 2021-10-04
Update support emails throughout the site

xdzurman:
AFTER MERGE TODO: UPDATE ENV ADALITE_SUPPORT_EMAIL!

### DIFF
--- a/app/frontend/components/common/contactForm.tsx
+++ b/app/frontend/components/common/contactForm.tsx
@@ -40,9 +40,7 @@ class ContactForm extends Component<Props, {submitted: boolean}> {
               troubleshooting suggestions
             </a>
             <span> before contacting us. You can also reach us out via </span>
-            <a href={`mailto:${ADALITE_CONFIG.ADALITE_SUPPORT_EMAIL}`}>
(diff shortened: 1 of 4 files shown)


app/frontend/components/common/contactForm.tsx
```diff
@@ -40,9 +40,7 @@ class ContactForm extends Component<Props, {submitted: boolean}> {
               troubleshooting suggestions
             </a>
             <span> before contacting us. You can also reach us out via </span>
-            <a href={`mailto:${ADALITE_CONFIG.ADALITE_SUPPORT_EMAIL}`}>
-              {ADALITE_CONFIG.ADALITE_SUPPORT_EMAIL}
-            </a>
+            <a href={'mailto:info@adalite.io'}>info@adalite.io</a>
```
refi93, line 43: why not use the ADALITE_SUPPORT_EMAIL constant here as well? Also, I'm wondering if we can't remove the env variable altogether and keep only the constant here

xdzurman: Acting according to internal slack communication. But yes, this constant seems redundant.

Issue #1086, opened 2021-07-30
News footer not fixed at bottom for big screens


refi93:
I was trying adalite.io on a big screen when I noticed that the "close" option is not visible until I scrolled to the bottom of the news modal. On my smaller screen, this doesn't happen. Screenshot to show what I mean:

![Screenshot from 2021-07-30 16-26-20](https://user-images.githubusercontent.com/4980147/127668106-adb6c35d-e731-4cc5-96be-9465fe2848a9.png)

Issue #1056, opened 2021-07-02
Update fee coefficient to match shelley era


refi93:
Currently we use still the old Byron-era coefficient of 43.946, but it has apparently been rounded to 44 in Shelley era: https://hydra.iohk.io/build/6510764/download/1/mainnet-shelley-genesis.json

Issue #867, opened 2021-03-05
Migrate to bigint


PeterBenc:
Currently Adalite only supports numbers but cardano transactions allow also big integers as values for lovelace, asset amount, fee etc.

refi93:
let's use https://www.npmjs.com/package/bignumber.js to handle easily also formatting of decimal numbers and avoid issues with older versions of browsers which may not support native JS BigInt

PR #1192, opened 2021-10-26
Enable strict null checks

coreyar:
This PR enables strict null checks in the typescript config.

Many changes were straightforward like explicitly declaring null or undefined types or enforcing a final type from logical operators using typecasting or ternaries.

Less straightforward changes include handling a missing wallet (it could be set to null when logged out). In these cases, I wrapped logic required the existence of a value in a conditional statement. I also added a function for type narrowing to check if the transaction plan result is a success or error and added an error case for if the transaction plan wasn't executed because of a missing wallet.

### DIFF
--- a/app/frontend/errors/unexpectedErrorReason.ts
+++ b/app/frontend/errors/unexpectedErrorReason.ts
@@ -11,4 +11,6 @@ export enum UnexpectedErrorReason {
   MissingStakingPath = 'MissingStakingPath',
   MissingStakingXpub = 'MissingStakingXpub',
   MissingDnsName = 'MissingDnsName',
+  AccessedMissingSourceAccount = 'AccessedMissingSourceAccount',
+  AccessedMissingActiveAccount = 'AccessedMissingActiveAccount',
 }
--- a/app/frontend/state.ts
+++ b/app/frontend/state.ts
@@ -1,6 +1,6 @@
 import {ADALITE_CONFIG} from './config'
 import {MainTabs} from './constants'
-import {InternalErrorReason} from './errors'
+import {InternalErrorReason, UnexpectedError, UnexpectedErrorReason} from './errors'
 import {StakepoolDataProvider} from './helpers/dataProviders/types'
 import {localStorageVars} from './localStorage'
 import {
@@ -235,7 +235,19 @@ export type SetStateFn = (newState: Partial<State>) => void
 export type GetStateFn = () => State
 export type Store = {getState: GetStateFn; setState: SetStateFn}
 
-export const getSourceAccountInfo = (state: State) => state.accountsInfo[state.sourceAccountIndex]
-export const getActiveAccountInfo = (state: State) => state.accountsInfo[state.activeAccountIndex]
+export const getSourceAccountInfo = (state: State) => {
+  const sourceAccountInfo = state.accountsInfo[state.sourceAccountIndex]
+  if (!sourceAccountInfo) {
+    throw new UnexpectedError(UnexpectedErrorReason.AccessedMissingSourceAccount)
+  }
+  return sourceAccountInfo
+}
+export const getActiveAccountInfo = (state: State) => {
+  const activeAccountInfo = state.accountsInfo[state.activeAccountIndex]
+  if (!activeAccountInfo) {
+    throw new UnexpectedError(UnexpectedErrorReason.AccessedMissingActiveAccount)
+  }
+  return activeAccountInfo
+}
 
 export {initialState}


app/frontend/wallet/shelley/types.ts
```diff
@@ -12,9 +12,9 @@ export type TxAux = {
   ttl: number
   certificates: TxCertificate[]
   withdrawals: TxWithdrawal[]
-  auxiliaryData: TxAuxiliaryData
-  auxiliaryDataHash: HexString
-  validityIntervalStart: number
+  auxiliaryData: TxAuxiliaryData | null
+  auxiliaryDataHash: HexString | undefined
```
refi93, line 16: I would rather keep this as `HexString | null` and modify the upstream code to pass such values instead of undefined, so it's clear where such potentially undefined values enter our logic and that we handle them properly, reducing the risk of bugs due to accidentally omitting some parameter only for it to end up as undefined without us noticing

app/frontend/actions/common.ts
```diff
@@ -46,10 +46,14 @@ export default (store: Store) => {
...
   }
 
-  const prepareTxPlan = async (args: TxPlanArgs): Promise<TxPlanResult> => {
+  const prepareTxPlan = async (args: TxPlanArgs): Promise<TxPlanResult | undefined> => {
     const state = getState()
+    const wallet = await getWallet()
     try {
-      return await getWallet()
+      if (!wallet) {
+        throw new Error('Wallet is not loaded')
```
refi93, line 54: I would rather extract this logic into a `getWalletOrThrow` helper function which would be guaranteed to return the wallet, making the places where we invoke it less cluttered with checks that shouldn't really trigger at runtime unless there's a bug

app/frontend/actions/accounts.ts
```diff
@@ -54,7 +60,7 @@ export default (store: Store) => {
       targetAccountIndex: accountIndex,
     })
     const targetAddress = await getWallet()
-      .getAccount(accountIndex)
+      ?.getAccount(accountIndex)
```
refi93, line 63: the wallet object shouldn't be undefined/null here, otherwise it could lead to inonsisten behavior down the line. As suggested in another comment, I'd rather introduce a getWalletOrThrow() method which would assert that the wallet is properly loaded

app/frontend/actions/delegate.ts
```diff
@@ -34,9 +36,13 @@ export default (store: Store) => {
     if (hasPoolIdentifiersChanged(state)) {
       return
     }
-    const poolInfo = !state.shelleyDelegation?.selectedPool?.name
-      ? await getWallet().getPoolInfo(state.shelleyDelegation?.selectedPool?.url)
-      : {}
+    const poolInfo = (!state.shelleyDelegation?.selectedPool?.name &&
```
refi93, line 39: This overriding of potentially undefined values with empty ones seems a bit fragile and I'd have a hard time reasoning about the possible consequences of it. I think it can be rewritten more sensibly like this:

```javascript
    const selectedPool = state.shelleyDelegation?.selectedPool
    const poolInfo = (!selectedPool?.name && selectedPool?.url) ?
      (await getWallet()?.getPoolInfo(selectedPool?.url))
      : null

    if (hasPoolIdentifiersChanged(state)) {
      return
    }
    const newState = getState()
    setState({
      shelleyDelegation: {
        ...state.shelleyDelegation,
        selectedPool: selectedPool ? {
          ...selectedPool,
          ...poolInfo,
        } : null,
```

@PeterBenc please check if I'm not missing anything by any chance as I think you were originally working on this

refi93:
Thanks a lot @coreyar for your PR and sorry for the late feedback. I left a few comments/suggestions which I believe should simplify the solution

app/frontend/actions/delegate.ts
PeterBenc: Yeah I agree, we shouldn't overide the values, that handled in `stakePoolInfo.tsx` file, What @refi93 proposes should imho work as expected so I am curious what was the reason to change the previous behavior, the difference being only doing `...null` instead of `...{}` which should be the same.

app/tests/src/transaction-planner.js
```diff
@@ -106,7 +108,7 @@ describe('Succesful transaction plans', () => {
     it(`should ${name}`, () => {
       const {utxos, changeAddress, fee, args} = setting
       const txPlanResult = selectMinimalTxPlan(utxos, changeAddress, args)
-      if (txPlanResult.success === true) {
+      if (isTxPlanResultSuccess(txPlanResult)) {
```
refi93, line 111: I would rather replace the one-line helper with its explicit implementation as it doesn't really simplify the logic, but it arguably adds one more layer of indirection to whoever wants to understand what this check does. Also it seems a bit weird for a function that seems t expect a tx plan to also allow null/undefined and hide it in its signature, possibly misleading someone casually checking this code that the tx plan must be defined

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -401,11 +404,11 @@ const ShelleyTrezorCryptoProvider = async ({
         return {
           finalizedTxAux: ShelleyTxAux({
             ...txAux,
-            auxiliaryDataHash: auxiliaryDataSupplement.auxiliaryDataHash,
+            auxiliaryDataHash: auxiliaryDataSupplement?.auxiliaryDataHash,
```
refi93, line 407: auxiliaryDataSupplement shouldn't really be undefined here, by the contract of the function called it should exist, so I'd rather add here a runtime assertion

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -504,11 +507,11 @@ const ShelleyLedgerCryptoProvider = async ({
         return {
           finalizedTxAux: ShelleyTxAux({
             ...txAux,
-            auxiliaryDataHash: auxiliaryDataSupplement.auxiliaryDataHashHex,
+            auxiliaryDataHash: auxiliaryDataSupplement?.auxiliaryDataHashHex,
           }),
           txAuxiliaryData: cborizeTxAuxiliaryVotingData(
             txAux.auxiliaryData,
-            auxiliaryDataSupplement.catalystRegistrationSignatureHex
+            auxiliaryDataSupplement?.catalystRegistrationSignatureHex
```
refi93, line 514: same as in ledger crypto provider

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -345,11 +345,11 @@ const cborizeTxVotingRegistration = ({
 
 const cborizeTxAuxiliaryVotingData = (
   txAuxiliaryData: TxAuxiliaryData,
-  signatureHex: string
+  signatureHex: string | undefined
```
refi93, line 348: I see in the upstream logic you relaxed the type of the signatureHex parameter, but it actually should always be defined

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -64,7 +64,9 @@ import {TxRelayType, TxStakepoolOwner, TxStakepoolRelay} from './helpers/poolCer
 import assertUnreachable from '../../helpers/assertUnreachable'
 
 let _activeTransport: Transport | null = null
-const getLedgerTransport = async (ledgerTransportType: LedgerTransportType): Promise<Transport> => {
+const getLedgerTransport = async (
+  ledgerTransportType: LedgerTransportType
+): Promise<Transport | null> => {
```
refi93, line 69: this code should not deal with any other ledger transport type than the ones handled here (webhid, u2f, webusb) and should throw in such case, so it should either throw or return a proper Transport. The "default" option matters really just for the UI. I would propose refactoring it like this, even within this PR as it shouldn't be much work and would be less confusing than relaxing the type signature of this function

1. Introduce a `LedgerTransportChoice` type which would be equivalent to the current `LedgerTransportType` and rename all existing instances in the code to LedgerTransportChoice
2. introduce `type LedgerTransportType = Exclude<LedgerTransportChoice, LedgerTransportChoice.DEFAULT>` and migrate from `LedgerTransportChoice` to `LedgerTransportType` in `getLedgerTransport()` output type  and `getDefaultLedgerTransportType()` function
3. `getLedgerTransport()` now can be refactored to `return assertUnreachable` in the default branch of the switch statement

app/frontend/actions/delegate.ts
coreyar: The problem was that type for `Stakepool` doesn't have any optional properties. As it was written if the poolInfo or selectedPool can't be fetched or doesn't exist then you end up with an empty object. I was thinking it would be better to have defaults than to type all the properties as optional. 

I am updating with the solution proposed by @refi93 because it does make more sense

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
coreyar: I am seeing in the docs that it can be undefined https://github.com/trezor/connect/blob/2955b5ef92a3c77ccf3b1353f74f37fa4a2ff337/docs/methods/cardanoSignTransaction.md#flowtype-1

I can still replace it with a runtime assertion

app/frontend/wallet/shelley/shelley-transaction.ts
coreyar: I allowed for this to be undefined because it is called with `auxiliaryDataSupplement.catalystSignature` as the value for `signatureHex` which can be undefined

https://github.com/trezor/connect/blob/2955b5ef92a3c77ccf3b1353f74f37fa4a2ff337/src/ts/types/networks/cardano.d.ts#L201

I can keep this type as defined and add a runtime check for where it is called with the undefined value

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
refi93: yes, indeed the docs aren't very clear on that now that I checked it deeper, sorry for the confusion, but it's supposed to be defined anyway for catalyst registrations and if not, this code wouldn't work as we need these data in such case. So a runtime assertion makes sense

app/frontend/wallet/shelley/shelley-transaction.ts
refi93: as explained in https://github.com/vacuumlabs/adalite/pull/1192#discussion_r741223653 let's add a runtime check

coreyar:
@refi93 Thank you for the feedback. I have updated the PR with your feedback and left comments where clarification was needed.

app/frontend/actions/reloadWallet.ts
```diff
@@ -26,13 +26,15 @@ export default (store: Store) => {
     const previousWalletReloadTime = _lastWalletReloadTime
     try {
       updateLastReloadTime()
-      const accountsInfo = await wallet.getAccountsInfo(state.validStakepoolDataProvider)
-      const tokensMetadata = await wallet.getTokensMetadata(accountsInfo)
+      const accountsInfo = state.validStakepoolDataProvider
+        ? await wallet?.getAccountsInfo(state.validStakepoolDataProvider)
```
refi93, line 30: `getWalletOrThrow()` should ensure that the wallet is defined
```suggestion
        ? await wallet.getAccountsInfo(state.validStakepoolDataProvider)
```

app/frontend/actions/wallet.ts
```diff
@@ -23,18 +30,24 @@ const getShouldShowSaturatedBanner = (accountsInfo: Array<AccountInfo>) =>
...
-let wallet: Wallet
-export const setWallet = (w: Wallet) => {
+let wallet: Wallet | null
+export const setWallet = (w: Wallet | null) => {
   wallet = w
 }
-export const getWallet = (): Wallet => wallet
+export const getWallet = (): Wallet | null => wallet
 
+export const getWalletOrThrow = (): Wallet => {
```
refi93, line 39: I see that getWallet() is not used anymore anywhere else, so I'd  rather rename `getWalletOrThrow()` to `getWallet()` and get rid of the old `getWallet()` function that wasn't checking whether the wallet is ready as the `OrThrow` suffix is arguably just exposing a lower-level concern than the callers would care about most of the time

app/frontend/actions/send.ts
```diff
@@ -36,16 +38,16 @@ export default (store: Store) => {
     if (state.sendAmount.assetFamily === AssetFamily.TOKEN) {
       const {policyId, assetName, quantity} = state.sendAmount.token
       // TODO: we should have a tokenProvider to get token O(1)
-      const tokenBalance = getSourceAccountInfo(state).tokenBalance.find(
-        (token) => token.policyId === policyId && token.assetName === assetName
-      ).quantity
+      const tokenBalance = getSourceAccountInfo(state)?.tokenBalance?.find(
```
refi93, line 41: `getSourceAccountInfo()` should throw if there's no source account, the original code would fail anyway if the source account isn't defined, so I'd add the assertion right in the declaration of `getSourceAccountInfo()`

app/frontend/actions/delegate.ts
```diff
@@ -64,59 +70,64 @@ export default (store: Store) => {
...
-        type: TxType.DELEGATE,
-        deposit: txPlanResult.txPlan.deposit,
-        stakePool: newState.shelleyDelegation?.selectedPool,
+      const newState = getState()
+      if (hasPoolIdentifiersChanged(newState)) {
+        return
+      }
+      const balance = getSourceAccountInfo(state).balance as Lovelace
+
+      if (isTxPlanResultSuccess(txPlanResult)) {
```
refi93, line 86: My comment https://github.com/vacuumlabs/adalite/pull/1192#discussion_r741082442 referred to all other instances of this helper as well

app/frontend/components/common/viewPort.tsx
```diff
@@ -31,8 +31,8 @@ const isSmallerThanDesktop = (screenType: ScreenType): boolean => {
   return [ScreenType.MOBILE, ScreenType.TABLET].includes(screenType)
 }
 
-const isBiggerThanMobile = (screenType: ScreenType): boolean => {
-  return [ScreenType.TABLET, ScreenType.DESKTOP].includes(screenType)
+const isBiggerThanMobile = (screenType: ScreenType | undefined): boolean => {
+  return !!(screenType && [ScreenType.TABLET, ScreenType.DESKTOP].includes(screenType))
```
refi93, line 35: I'd rather keep the interface of this function oblivious to the fact screenType might be undefined and fix the upstream logic to handle that

app/frontend/actions/delegate.ts
```diff
@@ -64,59 +70,64 @@ export default (store: Store) => {
...
-    }
-    const balance = getSourceAccountInfo(state).balance as Lovelace
-
-    if (txPlanResult.success === true) {
-      setState({
-        shelleyDelegation: {
-          ...newState.shelleyDelegation,
-          delegationFee: (txPlanResult.txPlan.fee + txPlanResult.txPlan.deposit) as Lovelace,
-        },
+    if (stakingAddress && poolHash) {
```
refi93, line 73: I'd rather put here a runtime assertion that stakingAddress must not be null as the original logic apparently relied on that and I don't see it working properly without that assumption being fulfilled and it doesn't really make sense to delegate without the staking address being set

app/frontend/actions/delegate.ts
```diff
@@ -169,13 +182,15 @@ export default (store: Store) => {
...
     loadingAction(state, 'Preparing transaction...')
-
-    const txPlanResult = await prepareTxPlan({
-      txType: TxType.DEREGISTER_STAKE_KEY,
-      rewards,
-      stakingAddress: sourceAccount.stakingAddress,
-    })
-    if (txPlanResult.success === true) {
+    let txPlanResult: TxPlanResult | null | undefined = null
+    if (sourceAccount.stakingAddress) {
```
refi93, line 186: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741304487

app/frontend/actions/transaction.ts
```diff
@@ -327,25 +335,29 @@ export default (store: Store) => {
       return
     }
     loadingAction(state, 'Preparing transaction...')
+    let txPlanResult: TxPlanResult | null | undefined = null
     // TODO: rewards should be of type Lovelace
     const rewards = getSourceAccountInfo(state).shelleyBalances.rewardsAccountBalance as Lovelace
     const stakingAddress = getSourceAccountInfo(state).stakingAddress
-    const txPlanResult = await prepareTxPlan({rewards, stakingAddress, txType: TxType.WITHDRAW})
+    if (stakingAddress) {
```
refi93, line 342: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741304487

app/frontend/actions/voting.tsx
```diff
@@ -54,21 +56,25 @@ export default (store: Store) => {
...
     const sourceAccount = getSourceAccountInfo(state)
-    const txPlanResult = await prepareTxPlan({
-      txType: TxType.REGISTER_VOTING,
-      votingPubKey,
-      stakePubKey,
-      stakingAddress: sourceAccount.stakingAddress,
-      nonce: BigInt(nonce),
-    })
+    let txPlanResult: TxPlanResult | null | undefined = null
+    if (sourceAccount.stakingAddress) {
```
refi93, line 67: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741304487

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -338,7 +339,7 @@ const ShelleyLedgerCryptoProvider = async ({
     const poolOwners: LedgerTypes.PoolOwner[] = stakepoolOwners.map((owner) => {
       // TODO: helper function for slicing first bit from staking address so its stakingKeyHash
       return !Buffer.compare(
```
refi93, line 341: Overriding the value to an empty string seems really fragile, I'd rather do something like this

```suggestion
      return owner.stakingKeyHashHex && !Buffer.compare(
```

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -347,7 +348,7 @@ const ShelleyLedgerCryptoProvider = async ({
         }
         : {
           type: LedgerTypes.PoolOwnerType.THIRD_PARTY,
-          params: {stakingKeyHashHex: owner.stakingKeyHashHex},
+          params: {stakingKeyHashHex: owner.stakingKeyHashHex || ''},
```
refi93, line 351: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741308487

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -212,7 +212,7 @@ function cborizeStakepoolRegistrationCert(
     ),
     Buffer.from(poolRegistrationParams.rewardAccountHex, 'hex'),
     poolRegistrationParams.poolOwners.map((ownerObj) => {
-      return Buffer.from(ownerObj.stakingKeyHashHex, 'hex')
+      return Buffer.from(ownerObj.stakingKeyHashHex || '', 'hex')
```
refi93, line 215: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741308487

app/frontend/wallet/shelley/shelley-transaction.ts
```diff
@@ -349,7 +349,7 @@ const cborizeTxAuxiliaryVotingData = (
 ): CborizedVotingRegistrationMetadata => [
   new Map<number, Map<number, Buffer | BigInt>>([
     cborizeTxVotingRegistration(txAuxiliaryData),
-    [61285, new Map<number, Buffer>([[1, Buffer.from(signatureHex, 'hex')]])],
+    [61285, new Map<number, Buffer>([[1, Buffer.from(signatureHex || '', 'hex')]])],
```
refi93, line 352: this fallback to empty string seems very fragile and should no longer be needed

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -153,8 +153,8 @@ const ShelleyTrezorCryptoProvider = async ({
     addressToAbsPathMapper: AddressToPathMapper
   ): TrezorTypes.CardanoInput {
     return {
-      ...(input.address && {path: addressToAbsPathMapper(input.address)}),
-      prev_hash: input.txHash,
+      path: input.address ? addressToAbsPathMapper(input.address) : [],
+      prev_hash: input.txHash || '',
```
refi93, line 157: this fallback to empty string shouldn't be needed

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -332,7 +335,7 @@ const ShelleyTrezorCryptoProvider = async ({
 
   const prepareByronWitness = (witness: TrezorTypes.CardanoSignedTxWitness): TxByronWitness => {
     const publicKey = Buffer.from(witness.pubKey, 'hex')
-    const chainCode = Buffer.from(witness.chainCode, 'hex')
+    const chainCode = Buffer.from(witness.chainCode || '', 'hex')
```
refi93, line 338: byron witnesses should have a chain code, so this fallback to empty string is really fragile. A runtime asserttion would be more sensible

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -373,10 +376,10 @@ const ShelleyTrezorCryptoProvider = async ({
         return {
           catalystRegistrationParameters: {
             votingPublicKey: txAuxiliaryData.votingPubKey,
-            stakingPath: txAuxiliaryData.rewardDestinationAddress.stakingPath,
+            stakingPath: txAuxiliaryData.rewardDestinationAddress.stakingPath || '',
```
refi93, line 379: rather than fallback to empty string which if triggered would break the code unpredictably, please put here a runtime assertion

app/frontend/actions/voting.tsx
```diff
@@ -54,21 +56,25 @@ export default (store: Store) => {
     state = getState()
 
     const stakePubKey = xpub2pub(
-      Buffer.from(getSourceAccountInfo(state).stakingXpub.xpubHex, 'hex')
+      Buffer.from(getSourceAccountInfo(state).stakingXpub?.xpubHex || '', 'hex')
```
refi93, line 59: I'd rather put here a runtime assertion that this xpubHex exists - if it wouldn't, this code would break unpredictably

app/frontend/actions/voting.tsx
```diff
@@ -54,21 +56,25 @@ export default (store: Store) => {
...
     ).toString('hex')
-    const nonce = await getWallet()
-      .getAccount(state.sourceAccountIndex)
-      .calculateTtl()
+    const nonce =
+      (await getWalletOrThrow()
+        .getAccount(state.sourceAccountIndex)
+        .calculateTtl()) || ''
```
refi93, line 64: this fallback to empty string should no longer be needed

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -172,14 +176,17 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
...
+        acc.push(curr?.caBalance.getTokens.map((token) => parseToken(token)))
+      }
+      return acc
+    }, [])
     const tokenBundle = aggregateTokenBundles(addressTokenBundles).filter(
       (token) => token.quantity > 0
     )
     const coins = addressInfos.reduce(
-      (acc, elem) => acc + parseInt(elem.caBalance.getCoin, 10),
+      (acc, elem) => acc + parseInt(elem?.caBalance.getCoin || '', 10),
```
refi93, line 189:
```suggestion
      (acc, elem) => acc + parseInt(elem?.caBalance.getCoin || '0', 10),
```

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -39,8 +39,8 @@ export const encodeAssetFingerprint = (policyIdHex: HexString, assetNameHex: Hex
   return bech32.encode('asset', data)
 }
 
-export const encodeCatalystVotingKey = (votingKey: HexString): string => {
-  return bech32.encode('ed25519_pk', Buffer.from(votingKey, 'hex'))
+export const encodeCatalystVotingKey = (votingKey: HexString | null): string => {
```
refi93, line 42: I would rather make this function accept only a defined votingKey and let the upstream logic handle the case when it isn't (if it can actually happen at all)

app/frontend/wallet/shelley/helpers/poolCertificateUtils.ts
```diff
@@ -167,33 +167,33 @@ const parseStakepoolRelays = (relays: _PoolRelay[]): TxStakepoolRelay[] =>
...
+            ipv6: relay.ipv6 ? ipv6BufToAddress(relay.ipv6) : undefined,
           },
         }
       case TxRelayTypes.SINGLE_HOST_NAME:
         return {
           type: TxRelayType.SINGLE_HOST_NAME,
           params: {
             portNumber: relay.portNumber,
-            dnsName: relay.dnsName,
+            dnsName: relay.dnsName || '',
```
refi93, line 179: this fallback to empty string seems really fragile as it is apparently a required property as per the type signature. A runtime assertion on the property being defined would be more sensible, same below

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -338,7 +339,7 @@ const ShelleyLedgerCryptoProvider = async ({
     const poolOwners: LedgerTypes.PoolOwner[] = stakepoolOwners.map((owner) => {
       // TODO: helper function for slicing first bit from staking address so its stakingKeyHash
       return !Buffer.compare(
-        Buffer.from(owner.stakingKeyHashHex, 'hex'),
+        Buffer.from(owner.stakingKeyHashHex || '', 'hex'),
```
refi93, line 342: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741308487

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -236,7 +236,7 @@ const ShelleyTrezorCryptoProvider = async ({
     const {data} = bech32.decode(certificate.stakingAddress)
     const owners = certificate.poolRegistrationParams.poolOwners.map((owner) => {
       // TODO: helper function stakingAddress2StakingKeyHash
-      return !Buffer.compare(Buffer.from(owner.stakingKeyHashHex, 'hex'), data.slice(1))
+      return !Buffer.compare(Buffer.from(owner.stakingKeyHashHex || '', 'hex'), data.slice(1))
```
refi93, line 239: same as https://github.com/vacuumlabs/adalite/pull/1192/files#r741308487

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -165,7 +167,7 @@ const SendAdaPage = ({
 
   const adaAsset: DropdownAssetItem = {
     type: AssetFamily.ADA,
-    policyId: null,
+    policyId: '',
```
refi93, line 170: non-null policyId doesn't make sense for ADA asse. The type of DropdownAssetItem should rather be fixed to have policyId as null for this case and as not-null for the case when the type is AssetFamily.TOKEN, leveraging possibly type union

app/frontend/actions/delegate.ts
coreyar: This is doing more then just abstracting the logic to check for the `.success === true`. It is narrowing the Union so that `txPlanResult` inside the conditional is assigned the success type and outside the conditional is assigned the error type. This is useful because it makes accessing different properties cleaner based on the narrowing.

I am happy to change it back but I believe that type narrowing is useful here.

app/frontend/components/common/viewPort.tsx
coreyar: This could also be handled by setting a default value for the state managed by the `useViewport` hook. Which would be similar to setting a fallback when the caller checks if the value is undefined or not.

app/frontend/actions/delegate.ts
coreyar: Since this logic will be repeated in three places does it make sense to throw this error in `prepareTxPlan` and relax the `stakingAddress` type in the signature?

refi93: I looked deeper into that and the only place where we set the stakingKey to null is the initial state of the app where we set an initial value to the `accountsInfo` property, but that doesn't really make sense. Let's just set `accountsInfo` there to an empty array and then we should be good to enforce that stakingAddress must be always defined

I mean here: https://github.com/vacuumlabs/adalite/blob/3119857be681ba54376a032b215ffa9544e5474b/app/frontend/state.ts#L223

Do you see some reason @PeterBenc why we would have to set some default "null" account in the initial state of the app, or is that just legacy from the times when we had a single account in the state?

refi93:
@coreyar I tried running the wallet with `yarn dev` and I got a `wallet is not loaded` error in the console and a white screen, so it seems that the assertion doesn't work everywhere it was put, here's the stacktrace:

```
Uncaught Error: Wallet is not loaded
    getWalletOrThrow wallet.ts:41
    default wallet.ts:54
    default reloadWallet.ts:10
    default transaction.ts:32
    default delegate.ts:24
    default accounts.ts:15
    default actions.ts:18
    mapActions util.js:9
    Preact 20
    reload walletApp.js:98
    js walletApp.js:105
    Webpack 6
```

coreyar:
> @coreyar I tried running the wallet with `yarn dev` and I got a `wallet is not loaded` error in the console and a white screen, so it seems that the assertion doesn't work everywhere it was put, here's the stacktrace:
> 
> ```
> Uncaught Error: Wallet is not loaded
>     getWalletOrThrow wallet.ts:41
>     default wallet.ts:54
>     default reloadWallet.ts:10
>     default transaction.ts:32
>     default delegate.ts:24
>     default accounts.ts:15
>     default actions.ts:18
>     mapActions util.js:9
>     Preact 20
>     reload walletApp.js:98
>     js walletApp.js:105
>     Webpack 6
> ```

Ok, I am not seeing any webpack errors building right now but let me investigate. I've pushed updates to the comments left so far

coreyar:
@refi93 I found the error. The wallet was being loaded in an outer scope but used in the scope of inner functions. I solved this by fetching the wallet inside the functions where it is used so that it isn't prematurely loaded

app/frontend/actions/delegate.ts
PeterBenc: Agreed, I dont see any reason why `accountInfos` cant be initially `[]`. Its not used until the wallet is loaded and then just assigned. 

BUT, I think I remember that setting account info to `[]` was actually a problem when logging out. I think that after logging out, some components still (for a brief moment I guess) "used" `accountInfos[state.activeAccount]` and accessed its keys so errors like `cant... something of undefined` popped up and the page crashed. Surely that probably points to an issue on its own.

So that was probably the reason for keeping it the way it is now, but I just tried setting it to `[]` and it works as expected. So I guess I either remember it wrong the issue causing it got fixed.

Anyway, just something to consider but a bit unrelated to this, having the initial `accountsInfo` as `[]` seem to work and makes sense.

coreyar:
I left the null checks in `loadWallet` because if `cryptoProvider` isn't fetched then the wallet won't be created so it could be null. Another option would be to wrap all the subsequent logic in the cryptoProvider conditional block or wrap it in another block checking for the wallet. 

It isn't absolutely necessary because it is wrapped in a try block with the catch setting an error that the wallet couldn't be loaded.

app/frontend/actions/voting.tsx
```diff
@@ -52,23 +55,28 @@ export default (store: Store) => {
     }
     loadingAction(state, 'Preparing transaction...')
     state = getState()
-
+    const sourceAccountInfo = getSourceAccountInfo(state)
+    if (!sourceAccountInfo.stakingXpub?.xpubHex) {
```
refi93, line 59: for these runtime errors I would rather use a standard assert(condition) statement - these shouldn't really happen at runtime and should rather be "technical" assertions and are more concise than "if ... then ... throw".

We have an `assert()` helper for which you can find in app/src/utls/assertion.ts. Same for other places where you introduced this pattern. Sorry for not claryfing that earlier

app/frontend/components/common/asset.tsx
```diff
@@ -20,7 +20,7 @@ export const LinkToAsset = ({policyIdHex, assetNameHex}: LinkToAssetProps) => (
 )
 
 export type FormattedAssetItemProps = Token & {
-  fingerprint: string
+  fingerprint: string | null
```
refi93, line 23: assets should always have a fingerprint. If this can be null, I'd rather check the root cause and assert there

refi93: I see it comes from DropdownAssetItem which indeed can be ADA as well, ok nevermind

refi93:
ok let's do it like this - this PR should already be close enough to what we want but needs further refinements and testing to tackle edge-cases that may not be as obvious and it would be more efficient for us to finish that, so I'd merge this to a enable-strict-null-checks branch in adalite repo and make he refinements there and then merge it to our develop branch

coreyar:
@refi93 Sounds good. I've got time to pickup another issue if that would be helpful

refi93:
Thanks a lot @coreyar for your contribution, here's the follow up PR https://github.com/vacuumlabs/adalite/pull/1206 where we'd refactor it/fixi it a bit more so it can be merged

refi93:
@coreyar great! - one issue that has been limiting us for a while is this one: https://github.com/vacuumlabs/adalite/issues/867 - currently the amounts are ordinary JS numbers which breaks as soon as you work with values bigger than max safe integer in JS which can happen especially with cardano tokens

Another, smaller issue that comes to my mind is this one: https://github.com/vacuumlabs/adalite/issues/1086

refi93:
The biginit migration would be pretty big, it probably makes sense to do so in several "layers", going from the innermost logic to the outermost one (or vice versa), if you choose to do it, so it can be split into multiple smaller PRs

refi93:
@coreyar another small issue worth doing can be this one: https://github.com/vacuumlabs/adalite/issues/1056

PR #1193, opened 2021-10-27
Display metadata token name instead of onchain asset name

DavidTranDucVL:
Currently, we display asset name that is on-chain. In metadata, name field is mandatory. For tokens that actually have metadata, it makes sense to display metadata name to user instead.

### DIFF
--- a/app/frontend/components/common/asset.tsx
+++ b/app/frontend/components/common/asset.tsx
@@ -1,6 +1,6 @@
 import {h} from 'preact'
 import printAda from '../../helpers/printAda'
-import {AssetFamily, Lovelace, Token} from '../../types'
+import {AssetFamily, Lovelace, RegisteredTokenMetadata, Token} from '../../types'
 import {AdaIcon, HomePageIcon, QuestionFillIcon, StarIcon} from './svg'
 import LinkIcon from './linkIcon'
 import {StringEllipsis} from '../pages/stringEllipsis'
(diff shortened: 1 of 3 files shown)


app/frontend/components/common/asset.tsx
```diff
@@ -74,13 +74,11 @@ export const FormattedAssetItem = ({
...
     return (
       <span>
-        {assetName ? (
-          <span className={styles.assetName}>{assetNameHex2Readable(assetName)}</span>
-        ) : (
-          <span className={styles.empty}>
-            {'<'}no name{'>'}
+        {
+          <span className={metadata || assetName ? styles.assetName : styles.empty}>
+            {metadata?.name ?? assetNameHex2Readable(assetName) ?? '<no name>'}
```
refi93, line 79: This would work incorrectly if the token has no metadata and an empty asset name hex - we would just show an empty string whereas before this PR this got formatted to `<no name>` which makes more sense.

I would suggest refactorng this logic into a pure helper function `formatAssetName()` which would explicitly receive the token assetNameHex and metadata (if present) and have explicit conditions that are easy to follow as I personally got a bit lost when evaluating this composite condition at first and it can hide pretty nasty edge-cases as the one I pointed out

app/frontend/components/common/asset.tsx
```diff
@@ -70,17 +70,25 @@ export const FormattedAssetItem = ({
     }
   }
 
+  const formatAssetName = (metadata: RegisteredTokenMetadata, assetNameHex: string) => {
+    if (metadata?.name) {
```
refi93, line 74: this contradicts the type signature which apparently expects metadata to be defined. If the metadata can actually be null/undefined I'd include it explicitly in the type signature or consider refactoring the upstream code so it doesn't happen. Also, I'd consider passing the assetNameHex first as that value is always defined.

app/frontend/components/common/asset.tsx
```diff
@@ -70,17 +70,24 @@ export const FormattedAssetItem = ({
...
     return (
       <span>
-        {assetName ? (
-          <span className={styles.assetName}>{assetNameHex2Readable(assetName)}</span>
-        ) : (
-          <span className={styles.empty}>
-            {'<'}no name{'>'}
+        {
+          <span className={metadata || assetName ? styles.assetName : styles.empty}>
+            {formatAssetName(assetName, metadata)}
```
DavidTranDucVL, line 88: Note: There is an issue with incorrect inference of type from `Map.prototype.get()` - which return type infers only to `T` instead of `T | undefined` .This is caused by disabled strict null checks. This issue is solved by: https://github.com/vacuumlabs/adalite/pull/1192

app/frontend/components/common/asset.tsx
DavidTranDucVL: Fixed

app/frontend/components/common/asset.tsx
DavidTranDucVL: Fixed

refi93 (approved):

PR #1206, opened 2021-11-03
Enable strict null checks

refi93:
follow up PR to #1192 

Changes respective to the original PR concern mostly reduction of changes to make reviewing easier and restoring narrowness of certain types, especially UTxO/TxPlan by putting explicit assertions on conditions that cannot happen at runtime as these would result in errors/inconsistent state anyway

Another minor refactor was the refactor of ledger transport selection - separating the enum for the "choice", including the "default option" vs transports themselves, cleaning up the transport selection logic in the ledger crypto provider.

Needs manual testing to verify that the functionality, especially sending transactions/sending pool registrations (where the parsing of inputs had to be changed to accommodate the type that expected addresses and other stuff not actually available) has been preserved


app/frontend/wallet/shelley/helpers/stakepoolRegistrationUtils.ts
```diff
@@ -58,12 +58,13 @@ const parseCliTokens = (tokenBundle: _MultiAsset[]): Token[] =>
...
 const parseCliInputs = (inputs: _Input[]): TxInput[] => {
   return inputs.map((input) => {
+    // TODO refactor to not override the types
     return {
       txHash: input.txHash.toString('hex'),
       outputIndex: input.outputIndex,
-      coins: null,
-      tokenBundle: null,
-      address: null,
+      coins: 0 as Lovelace,
```
refi93, line 65: I'm not sure if this works actually, we will need to test it

app/frontend/wallet/shelley/shelley-ledger-crypto-provider.ts
```diff
@@ -166,7 +169,7 @@ const ShelleyLedgerCryptoProvider = async ({
 
   async function displayAddressForPath(
     absDerivationPath: BIP32Path,
-    stakingPath?: BIP32Path
+    stakingPath: BIP32Path
```
refi93, line 172: stakingPath being optional seems to be a left-over from byron era and the general CryptoProvider interface had it actually enforced

PeterBenc (commented):
I went through the code, codewise LGTM, tho I found couple of bugs.

1. This one is IMHO related to this [comment](https://github.com/vacuumlabs/adalite/pull/1192#discussion_r742107888)
After logging out the page crashes.

image

2. I suppose that setting something to `''` instead of null cause that instead of showing `<Unknown pool>` we show empty string for the pool names. I noticed that this problem is also in production in the `Previous rewards` section :thinking:

image

3. I also tried voting, (works as expected) and pool registrations. The pool registration txs I have are pretty old but they used to work. Now they dont. Ledger fails with security policy error but it also fails on mainnet. I haven't been following the latest changes in ledger API so it might be something I missed. TLDR: I wasn't able to test pool registration. (feel free to ping me for more info)

Other than this, I haven't found any other bugs.

refi93:
@PeterBenc

"unknown pool" parsing fixed - the condition to map the pool hash to its name was refactored incorrectly, setting the name to undefined

The logout issue turned out to be trickier. We've been apparently setting the account info to some empty/partial data to work around the fact that during logout which was overwriting the state, components depending on account info (e.g. staking tab) may still be connected to it.

I found it more resilient to refactor the logout to do a page refresh and keep the account infos array in the initial state empty, while adding assertions to the helper that fetches activeAccountInfo rather than passing some malformed data to it. If some component depending on account info would actually have to be rendered with no account info, then it's probably more sensible to make these components handle that case explicitly (which currently isn't the case), but right now I'm not aware of any scenario (apart from logout) where that could possibly happen, so I just put assertions in the account info getters, to ensure that it wouldn't lead to further inconsistencies and make the app crash instead

PR #1207, opened 2021-11-04
Apply sticky footer CSS outside of media query

### LINKED ISSUE #1086: News footer not fixed at bottom for big screens

coreyar:
The css that makes the footer sticky currently has a max height of 1366px so it doesn't get applied on large screens.
I removed this so that the css is applied on all screen sizes as a base that can be overridden on smaller screens.

Fixes #1086 

### Notes
- I've found it very helpful to use storybooks with visual regression snapshots to understand the effects of css changes

### DIFF
--- a/app/public/css/0-1366px.css
+++ b/app/public/css/0-1366px.css
@@ -11,17 +11,6 @@
     padding-right: 56px;
   }
 
-  .welcome-footer {
-    position: -webkit-sticky;
-    position: sticky;
-    width: 100%;
-    bottom: -56px;
-    border-top: 1px solid var(--color-border);
-    padding: 24px 56px;
-    margin-top: 0;
-    background-color: white;
-  }
-
   .welcome .credits {
     border-bottom-width: 0;
   }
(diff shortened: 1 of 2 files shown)


app/public/css/0-1366px.css
```diff
@@ -1,28 +1,26 @@
-@media screen and (max-height: 1366px) {
-  .welcome {
-    position: relative;
-    margin-bottom: -112px;
-    margin-left: -56px;
-    margin-right: -56px;
-  }
+ .welcome {
```
refi93, line 1: I would rather look for the root cause of the missing footer for big resolutions and fix the base CSS that remove the restriction on screen resolutions from a CSS meant for a certain resolution range (0-1366px) as its name suggests

coreyar: ok, I can update the base welcome-footer class with the correct styles for the footer and leave the rest within the original resolution range.

refi93:
approving and creating a branch in our repo to do final polishing/rebasing and merge it

refi93 (approved):

PR #1212, opened 2021-11-09
Bitbox02 support final polishing

refi93:
follow up PR based on #1187 to enable CI/review app and implement last fixes (if needed)

changes/refactors (2nd commit):
* get rid of isHwWallet and hwWalletName state variables and crypto provider methods, replace them with pure helpers based on crypto provider type
* streamline feature checks in UI by exposing `cryptoProviderInfo` in the state together with supported features by the currently used provider/device
* show warning regarding tokes for BirBox02 users in the UI, prevent multiasset transactions from being signed with bitbox

TODOs:

- [x] fix failing unit tests
- [x] test end-to-end
- [x] refactor logic for determining in UI whether feature is supported (Voting/pool registration)
- [x] update affiliate link (https://github.com/vacuumlabs/adalite/pull/1187#discussion_r736424643)
- [x] link usage tutorial for users


benma:
Fyi, we just released BitBoxApp v4.30.0 containing the firmware with Cardano support. In other words, feel free to deploy at any time :smile: 

https://shiftcrypto.ch/download/

https://github.com/digitalbitbox/bitbox-wallet-app/releases/tag/v4.30.0

refi93:
Hi @benma - while testing I realized bitbox doesn't seem to support Cardano tokens. The problem is that in Cardano, tokens are bound to the unspent output and therefore if a bitbox user receives a token, they would not be able to spend the ADA associated with it either. So somebody can in theory send a bitbox user ADA together with a token only for the ADA to be unspendable by the bitbox user (until bitbox supports cardano tokens), which in theory can be done on purpose by the sender and can cause issues especially if it is e.g. used as a cold wallet to manage some exchange.

Did you account for this issue when you decided not to implement tokens support? I would recommend adding tokens support if possible to prevent this kind of issues for Bitbox users before going live. An alternative (if bitbox is fully interoperable

app/tests/src/tokenRegistry.js
```diff
@@ -111,6 +111,7 @@ describe('Token registry api', () => {
     const expected = new Map(
       Object.entries({
         '1d7f33bd23d85e1a25d87d86fac4f199c3197a2f7afeb662a0f34e1e776f726c646d6f62696c65746f6b656e': {
+          name: 'World Mobile Token',
```
refi93, line 114: not sure why the test was working fine before, seems like the upgrade of webpack led to `Map`s being compared properly

app/frontend/components/pages/common.tsx
```diff
@@ -16,3 +21,17 @@ export const EpochDateTime = ({
...
+
+export const BitBox02MultiAssetAlert = () => {
+  const cryptoProviderInfo = useSelector((state) => state.cryptoProviderInfo)
+  const isMultiAssetSupported = useIsWalletFeatureSupported(CryptoProviderFeature.MULTI_ASSET)
+  return !isMultiAssetSupported && cryptoProviderInfo?.type === CryptoProviderType.BITBOX02 ? (
+    <Alert alertType="warning">
+      BitBox02 currently does not support sending Cardano tokens. If you need to send/receive
+      tokens, wait for the firmware update or use Ledger instead.{' '}
+      {
+        // TODO add link to more info
```
refi93, line 33: planning to add bitbox tutorial once done/published by them

benma: Please link to this guide: 

https://shiftcrypto.support/help/en-us/35-adalite-cardano

We will flesh it out some more, but the link will stay the same.

app/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/lodash": "^4.14.168",
     "babel-polyfill": "6.26.0",
     "bip39-light": "^1.0.7",
+    "bitbox02-api": "^0.11.0",
```
xdzurman, line 30: When building the app, I'm getting

```
WARNING in ./app/node_modules/bitbox02-api/lib/bitbox02-api-go.js 4:181-188
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./app/node_modules/bitbox02-api/lib/bitbox02.js 9:21-52
 @ ./app/node_modules/bitbox02-api/lib/index.js 45:14-38
 @ ./app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts 298:45-106
 @ ./app/frontend/wallet/shelley/shelley-crypto-provider-factory.ts 3:0-79 11:23-52
 @ ./app/frontend/actions/wallet.ts 49:0-93 179:45-91
 @ ./app/frontend/actions.ts 23:0-45 26:440-453
 @ ./app/frontend/components/autoLogout.ts 16:0-33 78:7-14
 @ ./app/frontend/components/app.tsx 10:0-38 35:38-48
 @ ./app/frontend/walletApp.js 96:14-49
```

I recall that this might have been discussed, but is this going to get fixed? Perhaps in the new version that benma suggested?

app/frontend/state.ts
```diff
@@ -61,7 +62,7 @@ export interface State {
     mnemonicInputError: {code: InternalErrorReason} | null
     formIsValid: boolean
   }
-  hwWalletName?: string
+  cryptoProviderInfo?: CryptoProviderInfo
```
xdzurman, line 65: Not sure if it relates exactly to cryptoProviderInfo, but when developing with `ADALITE_DEVEL_AUTO_LOGIN = true` and trying to log out, the app does not stay logged out but immediately logs user in again, making it possible to actually log out.

app/frontend/components/pages/txHistory/transactionHistory.tsx
```diff
@@ -242,7 +242,7 @@ const TransactionHistory = (): h.JSX.Element => {
       </div>
       <div className="staking-history-warning">
         <Alert alertType="warning">
-          CSV exports might be wrong due to possibly missing rewards.{' '}
+          Some rewards may miss from the CSV export.{' '}
```
xdzurman, line 245: I think `be missing` is the correct way to say it in english.

app/frontend/components/pages/voting/votingCard.tsx
```diff
@@ -26,16 +31,21 @@ const AppDownloadInfo = ({url, imageSrc}: {url: string; imageSrc: string}) => (
   </div>
 )
 
-const VotingCard = (): h.JSX.Element => {
+const VotingCard = (state): h.JSX.Element => {
```
xdzurman, line 34: Is this left over?

app/frontend/errors/errorMessages.ts
```diff
@@ -87,6 +87,14 @@ const internalErrorMessages: {[key in InternalErrorReason]: (params?: any) => st
     'ServerError: Our servers are probably down. Please try again later and if the problem persists, contact us.',
   [InternalErrorReason.EpochBoundaryUnderway]: () =>
     'Our servers are temporarily down while Cardano is undergoing an epoch boundary. We should be back in a few minutes.',
+  [InternalErrorReason.BitBox02OutdatedFirmwareError]: ({message}) =>
+    `BitBox02OutdatedFirmwareError: Your BitBox02 firmware (version ${message}) is outdated. Please update your firmware to version ${
+      BITBOX02_VERSIONS[CryptoProviderFeature.MINIMAL].major
+    }.${BITBOX02_VERSIONS[CryptoProviderFeature.MINIMAL].minor}.${
+      BITBOX02_VERSIONS[CryptoProviderFeature.MINIMAL].patch
```
xdzurman, line 94: Not too relevant for this PR, but we have a lot of this duplicated and spacious versioning code which could be quite prettily solved by creating a helper function, maybe we should refactor it in the future.

app/frontend/state.ts
```diff
@@ -196,7 +197,7 @@ const initialState: State = {
   sendTransactionSummary: {
     // this should be called only transactionSummary
     type: TxType.SEND_ADA,
-    address: null,
+    address: null as any,
```
xdzurman, line 200: Is this also related to the `// TODO avoid casting to any` comment?

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -0,0 +1,350 @@
...
+type CryptoProviderParams = {
+  network: Network
+  config: any
+}
+
+const ShelleyBitBox02CryptoProvider = async ({
+  network,
+  config,
+}: CryptoProviderParams): Promise<CryptoProvider> => {
+  const bitbox02API = await import(/* webpackChunkName: "bitbox02-api" */ 'bitbox02-api')
```
xdzurman, line 54: Can you please add an explanatory comment of what this does and why it's needed?

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -0,0 +1,350 @@
...
+      case CryptoProviderFeature.BYRON:
+      case CryptoProviderFeature.POOL_OWNER:
+      case CryptoProviderFeature.MULTI_ASSET:
+      case CryptoProviderFeature.VOTING:
+        return false
+      default:
+        return BITBOX02_VERSIONS[feature]
+          ? hasRequiredVersion(version, BITBOX02_VERSIONS[feature])
+          : hasRequiredVersion(version, BITBOX02_VERSIONS[CryptoProviderFeature.MINIMAL])
+    }
```
xdzurman, line 156:
```suggestion
        return hasRequiredVersion(version, BITBOX02_VERSIONS[feature] ?? BITBOX02_VERSIONS[CryptoProviderFeature.MINIMAL])
    }
```

same as a few lines below

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -0,0 +1,350 @@
...
+          keypathPayment: absDerivationPath,
+          keypathStake: stakingPath,
+        },
+      })
+    })
+  }
+
+  function getWalletSecret(): void {
+    throw new UnexpectedError(UnexpectedErrorReason.UnsupportedOperationError, {
+      message: 'Unsupported operation!',
```
xdzurman, line 196: Inconsistent with `message: 'Operation not supported',` a few lines higher

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -0,0 +1,350 @@
...
+      prevOutHash: Buffer.from(input.txHash, 'hex'),
+      prevOutIndex: input.outputIndex,
+    }
+  }
+
+  function prepareOutput(output: TxOutput): CardanoOutput {
+    if (output.tokenBundle.length > 0 && !isFeatureSupported(CryptoProviderFeature.MULTI_ASSET)) {
+      throw new InternalError(InternalErrorReason.BitBox02MultiAssetNotSupported, {
+        message:
+          'Sending tokens is currently not supported on BitBox02. Please wait for the firmware update or use Ledger instead.',
```
xdzurman, line 216:
```suggestion
          'Sending tokens is currently not supported on BitBox02. Please wait for the firmware update or use other wallet options (Ledger, Trezor, Mnemonic) instead.',
```

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -0,0 +1,350 @@
...
+    const certificates = txAux.certificates.map((certificate) =>
+      prepareCertificate(certificate, addressToAbsPathMapper)
+    )
+    const validityIntervalStart = txAux.validityIntervalStart
+      ? `${txAux.validityIntervalStart}`
+      : null
+
+    if (
+      txAux.auxiliaryData?.type === 'CATALYST_VOTING' &&
+      isFeatureSupported(CryptoProviderFeature.VOTING)
```
xdzurman, line 293: Isn't this supposed to be `!isFeatureSupported(CryptoProviderFeature.VOTING)` instead?

xdzurman (commented):
Minor comments and suggestions, I'll properly test it whole tomorrow or on Thursday.

app/frontend/actions/poolOwner.ts
```diff
@@ -76,12 +76,14 @@ export default (store: Store) => {
     try {
       // TODO: refactor feature support logic
       const supportError = getWallet().ensureFeatureIsSupported(CryptoProviderFeature.POOL_OWNER)
+      const {cryptoProviderInfo} = state
+
       if (supportError) {
         throw new InternalError(supportError.code, {message: supportError.params.message})
       }
-      if (usingHwWalletSelector(state)) {
+      if (cryptoProviderInfo && isHwWallet(cryptoProviderInfo.type)) {
```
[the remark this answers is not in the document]
refi93: I can relax the isHwWallet type guard to accept `any` and then we could directly write `isHwWallet(cryptoProviderInfo?.type)` - what do you think?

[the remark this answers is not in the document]
xdzurman: Or perhaps have `cryptoProviderInfo: CryptoProviderInfo | null` insetad of `cryptoProviderInfo?: CryptoProviderInfo` and let the isHwWallet accept `CryptoProviderInfo | null`?

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
refi93: actually, only Ledger seems to be fully interoperable with Bitbox (i.e. same keys/addresses produced from the same mnemonic), that's why I mentioned explicitly only Ledger. I'd expect it to be elaborated on in the bitbox tutorial for Adalite that's being prepared

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
refi93: I noticed the same inconsistency in the rest of crypto providers, but I agree it's worth unifying it for good, will do :+1:

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
xdzurman: Oh. In that case I would specify this compatibility somehow, e.g. `or use Ledger (with the same seed phrase) instead.`

app/package.json
refi93: yes, I called out the same issue in the original PR, there doesn't seem to be a simple fix to it right now, but should be solved in some later release of the library: https://github.com/vacuumlabs/adalite/pull/1187#issuecomment-963408318

app/frontend/actions/common.ts
```diff
@@ -24,7 +24,7 @@ export default (store: Store) => {
       sendTransactionSummary: {
         // TODO: we should reset this to null
         type: TxType.SEND_ADA,
-        address: null,
+        address: null as any, // TODO avoid casting to any
```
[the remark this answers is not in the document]
refi93: I was actually a bit hesitant to do it, but you convinced me :) I pushed the refactor to make sendTransactionSummary in the state optional

app/frontend/state.ts
refi93: yes, as mentioned in the comment above, I removed that

app/frontend/state.ts
refi93: I see, good catch - that's actually a side-effect of a refactor done in https://github.com/vacuumlabs/adalite/pull/1206 where "soft" logout was replaced by full page refresh to simplify reasoning about the state which can be in a weird intermediary state as you overwrite the wallet data at once which broke assumptions around strict null checks.

I don't see a nice way to solve it right now and I would not really consider worth it solving it soon as it's a development only issue and you can still test logout properly, just with autologin off

app/frontend/actions/poolOwner.ts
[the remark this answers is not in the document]
refi93: Yeah I ended up doing something like that, though setting `cryptoProviderInfo: CryptoProviderInfo | undefined` for some reason broke type inference in places where I did `isHwWallet(cryptoProviderInfo?.type)` and TS started complaining that the value may be undefined in code depending on that guard passing, so I just left it at `cryptoProviderInfo?: CryptoProviderInfo`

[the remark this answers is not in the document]
refi93: ok, it's apparently happening regardless, I'm working on fixing it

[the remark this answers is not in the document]
refi93: It wasn't as smooth as I hoped due to this issue: https://github.com/microsoft/TypeScript/issues/38136

What I did is I created a selector to obtain the cryptoProvider type directly which is a workaround to apply the type guard on the same value that would be used to print the device brand name

app/bitbox02-api.d.ts
```diff
@@ -0,0 +1,105 @@
...
+      network: CardanoNetwork,
+      scriptConfig: ScriptConfig,
+      display?: boolean
+    ): Promise<string>
+    cardanoSignTransaction(params: {
+      network: CardanoNetwork
+      inputs: CardanoInput[]
+      outputs: CardanoOutput[]
+      fee: string
+      ttl: string
```
refi93, line 97: @benma btw ttl in cardano transactions is optional since Mary era, similarly as `validityIntervalStart` - we actually didn't implement that in Adalite so far, but I'm making a PR to fix that: https://github.com/vacuumlabs/adalite/pull/1226/files

So alongside multiasset support it would be great if you included also support for optional TTL for the sake of consistency

benma: The BitBox02 as well as the bitbox02-api package actually accept undefined/null for ttl and validityIntervalStart and handle it properly. You can simply adjust the type here to `string | null` like with `validityIntervalStart`.

refi93: I tried passing ttl=null, but the signature returned by bitbox didn't match with the tx body that our code serialized. Do you have some test fixture showing how bitbox serializes a transaction with null ttl? null ttl should be missing completely from the transaction body as per the CDDL

benma: We have test fixtures, but none where ttl is zero at the moment.

The relevant cbor serialization code is here:

https://github.com/digitalbitbox/bitbox02-firmware/blob/d7e59db2fde8823354c49b2325b2656992a0e32b/src/rust/bitbox02-rust/src/hww/api/cardano/sign_transaction/cbor.rs#L123-L126

I'll need a bit more time to make a test and compare it to other implementations to see where the root of the problem is.

Adalite always sets a ttl though, right?

refi93: Adalite UI yes, but the crypto providers are generic and we allow also pool registrations passed as raw CBOR and these can have TTL not defined. BTW, looking at the code you attached - ttl = 0 is in theory a semantically valid TTL the transaction can have - advanced users may leverage that to sign a transaction and ensure it won't ever be accepted by the blockchain - it's used e.g. with Ledger when CLI users need to sign separately voting registration metadata (but the device allows signing only the full tx) and therefore want to throw away the rest of the signed transaction. (of course, they can use ttl=1 which would do as well, but mapping 0 to null seems worth avoiding for the sake of predictability of the interface)

So it would be ideal, if your implementation distinguished between ttl being `null` and `0` and treated `0` as a possible TTL value, to avoid confusion

benma: Right, the firmware so far treats ttl=0 the same as if it was missing. Interesting, I wasn't aware that a zero (but not missing) TTL was used with a good reason. I will add this change to the next release, fixing this part.

That does not explain yet why `ttl=null` in your test yielded different serializations by the firmware and the Adalite code. I will investigate and come back to you.

Note that the BitBox02 doesn't support pool registrations (also disabled in the BB02 crypto provider in Adalite), so I think this edge case is not a blocker for release. If the issue turns out to be in the BitBox02 firmware, we will resolve it by the next release.

xdzurman:
Trying to send tokens:
image
The device shows TADA amount and fee, confirms and then results in an error:
image

xdzurman:
When you turn on the device, it explicitly sends you to BitBox app. However, when you try to interact with BitBox on Adalite while BitBoxApp is opened, no devices are shown: 
image
I think we should be more explicit about the error message that follows:
image
and mention that other apps (including BitBoxApp) should be closed.

benma:
> When you turn on the device, it explicitly sends you to BitBox app. However, when you try to interact with BitBox on Adalite while BitBoxApp is opened, no devices are shown:

Good point - we will make sure to mention it in our [guide](https://shiftcrypto.support/help/en-us/35-adalite-cardano) as well. The BitBox02 can only handle one open connection to a wallet app at a time. Improving the error message makes sense.

refi93:
@xdzurman - sorry, the review app was deployed from an old commit, I fixed that. Regarding handling of tokens - currently users see a big warning in the send/receive screens and if they try to send a token, they get an error message clearly stating why the transaction failed and it doesn't let them sign some malformed tx on the device.

As a follow up, we can consider even disabling the tokens in the dropdown in the send screen, I just didn't want to complicate the solution too much for such an edge-case

app/frontend/wallet/helpers/cryptoProviderUtils.ts
```diff
@@ -25,8 +25,12 @@ export function getWalletName(providerType: CryptoProviderType): WalletName {
 }
 
 export function isHwWallet(
-  providerType: CryptoProviderType
+  providerType: CryptoProviderType | undefined | null
 ): providerType is HwWalletCryptoProviderType {
+  if (providerType == null) {
+    return false
+  }
```
xdzurman, line 32: If it's undefined, it will bubble down to `.includes` and always return false, wouldn't it be better to have 
```suggestion
  if (providerType == null || providerType == undefined) {
    return false
  }
```
so that it's more transparent and caught earliest possible?

xdzurman (commented):
Refactors review - LGTM, just one comment

Is there anything else that's planned in this PR? I'll approve if not.

app/frontend/wallet/helpers/cryptoProviderUtils.ts
refi93: I personally see `== null` check as an idiomatic way of expressing "equal to null or undefined" so I would soo the `== undefined` check as redundant and may be actually a bit distracting as one may legitimately wonder why there is such otherwise redundant check

It's even regarded as the standard way to check for null and undefined simultaneously: https://stackoverflow.com/a/21273362/2657249

refi93:
> Is there anything else that's planned in this PR? I'll approve if not.

No, I wasn't planning to do anything more here, there's a follow-up PR to deal with undefined TTL (#1226 ), but that's fixing a pre-existing bug in our app, but I still need to sort out the Bitbox02 side there regarding how it handles null ttl but I didn't get to it yet and it's not high priority

app/bitbox02-api.d.ts
refi93: > Note that the BitBox02 doesn't support pool registrations (also disabled in the BB02 crypto provider in Adalite), so I think this edge case is not a blocker for release.

Yes, it isn't a blocker, we are planning to deal with that null ttl issue separately in #1226 :+1:

app/frontend/wallet/helpers/cryptoProviderUtils.ts
xdzurman: Ohhhh I didn't notice that it's `==`, sorry. I'm too used to being very explicit :D

xdzurman (approved):

app/bitbox02-api.d.ts
benma: @refi93 

Good news: sending an undefined ttl actually works, and the BitBox02 and Adalite both serialize the tx the same way. I just send an onchain transaction using this branch:

https://github.com/benma/adalite/commit/a58ebbb33a3879b46084b347de602701abdf68f0

Since the type of `ttl` can't currently be null or undefined, I changed the type to be `number | undefined`, which resulted in these two important changes:

a)

Don't serialize the ttl if it is not undefined, not if it is not null. Note that the current check `!== null` does not do anything as in the current type, it cannot be null).

https://github.com/benma/adalite/commit/a58ebbb33a3879b46084b347de602701abdf68f0#diff-32ff170d6209da4ac450ad17009a98373855e630758dba679e4afc7f4b0c6a99R94

b)

Don't do `toString()` without checking for undefined first:

https://github.com/benma/adalite/commit/a58ebbb33a3879b46084b347de602701abdf68f0#diff-3a3f4139e69bc5cbc4f08f613bb13fc943236d3825d2e105dd467f9de58014a8R309

I suspect that one of these changes were missing from your experiment.

PR #1213, opened 2021-11-11
Remove redundant ADALITE_SUPPORT_EMAIL env var, hardcode formspree link

refi93:
Motivation: ADALITE_SUPPORT_EMAIL env var was currently "abused" to provide the formspree link suffix (which we migrated to from suffixing the link with an email a while ago) which introduced confusion to the code for whoever lacked the background.

Changes: Given in other places we already get the support email from constants, I removed the env variable completely and hardcoded the formspree link as it isn't supposed to change

Testing: Please send a test email to check that formspree works as intended (I'm not receiving emails from there)


xdzurman (approved):
Lgtm, tested, successfully sent an email, theres just a lint error that needs to be fixed

PR #1223, opened 2021-11-22
Handle undefined pool info

refi93:
Motivation: fix potentially undefined pool info (possibly for some pool that has been unregistered). We've got reports by users with the following breadcrumbs:

```javascript
      const getPool = async (poolHash: string): Promise<StakePoolInfo | HostedPoolMetadata> => {
      const stakePool = validStakepoolDataProvider.getPoolInfoByPoolHash(poolHash)
      if (stakePool?.name) {
        return stakePool
      }
      const poolInfo = await getPoolInfo(stakePool.url)
```
and error: `Cannot read properties of undefined (reading 'url')` (Chrome) or `stakePool is undefined` (Firefox)

Changes: fix signature of getPoolInfoByPoolHash/Ticker to "acknowledge" it may return also null and the typescipt naturally warns in places where this breaks the assumption the data is always defined, which turned out to be only the place where this error got reported.

Testing: I don't have a pool to test it with but I emulated the situation locally by modifying the `getPoolInfoByPoolHash()` function to return `undefined`, managed to reproduce the error on `develop` and checked that the proposed fix makes the error disappear and the pool appears just as "Unknown pool" which is desired behavior in such case

Sidenote:
`yarn audit` failed on `universal-analytics` and `cypress` package so I'm bumping its version as part of this PR as well. I also ported a fix for cypress tess in CI so the job properly awaits on the FE being build/server launched which should fix the flakiness of CI on integration tests we've been experiencing so far (original "wait-on" parameter was basically useless - it's not a parameter cypress would listen on, and rather than the server, the cypress test suite should be waiting but there was no such parameter for it set up)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,12 +52,13 @@ jobs:
       - run:
           name: Start server with correct env for cypress tests
           command: yarn cypress:dev
-          wait-on: https://localhost:3000
           background: true
       - run:
           name: Run Cypress tests
           store_artifacts: true
-          command: VIDEO=true yarn cypress:run
+          command: |
+            npx wait-on https://localhost:3000 --timeout 120000
+            VIDEO=true yarn cypress:run
       - store_artifacts:
           path: app/cypress/videos
       - store_artifacts:
--- a/app/package.json
+++ b/app/package.json
@@ -25,6 +25,7 @@
     "@ledgerhq/hw-transport-webusb": "^5.53.1",
     "@sentry/browser": "^6.15.0",
     "@types/lodash": "^4.14.168",
+    "babel-polyfill": "6.26.0",
     "bip39-light": "^1.0.7",
     "borc": "^2.1.2",
     "cardano-crypto.js": "^6.1.0",
--- a/app/package.json
+++ b/app/package.json
@@ -30,6 +30,7 @@
     "cardano-crypto.js": "^6.1.0",
     "chacha": "^2.1.0",
     "lodash": "^4.17.21",
+    "moment": "^2.29.1",
     "pbkdf2": "^3.1.2",
     "preact": "^10.5.11",
     "react-device-detect": "^1.17.0",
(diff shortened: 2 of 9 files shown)


app/package.json
```diff
@@ -38,13 +39,14 @@
...
+    "cypress": "^9.0.0",
     "fetch-mock": "^6.5.1",
     "fetch-readablestream": "^0.2.0",
     "sinon": "^7.3.2"
   },
   "resolutions": {
     "lodash": ">=4.17.21",
-    "ansi-regex": "<6.0.0"
+    "ansi-regex": "<6.0.0",
+    "cypress/**/json-schema": ">=0.4.0"
```
refi93, line 50: for cypress it turns out I needed to override the json-schema package version as the latest version seems to still recursively depend on the vulnerable one

app/package.json
```diff
@@ -30,6 +30,7 @@
     "cardano-crypto.js": "^6.1.0",
     "chacha": "^2.1.0",
     "lodash": "^4.17.21",
+    "moment": "^2.29.1",
```
refi93, line 33: the update of cypress (technically not needed but there shouldn't be any harm in doing so) apparently dropped the 'moment' recursive dependency and we weren't importing it explicitly

refi93: I directly added a commit t enable "no-extraneous-dependencies" rule and fixed other packages (babel-polyfill and glob) that were apparently also implicitly imported

.circleci/config.yml
```diff
@@ -52,12 +52,13 @@ jobs:
       - run:
           name: Start server with correct env for cypress tests
           command: yarn cypress:dev
-          wait-on: https://localhost:3000
           background: true
```
refi93, line 55: the original `wait-on` parameter wasn't doing anything apparently as parameters are used for circleci commands if I understand correctly, but this is just a direct invocation of a bash command (https://circleci.com/docs/2.0/reusing-config/#using-the-parameters-declaration)

package.json
```diff
@@ -58,13 +58,14 @@
...
     "express-mung": "^0.5.1",
+    "glob": "^7.2.0",
     "isomorphic-fetch": "^3.0.0",
     "mocha-headless-chrome": "^2.0.3",
     "moment": "^2.27.0",
     "normalize-url": "^4.0.0",
     "redis": "^3.1.2",
     "redisscan": "^2.0.0",
-    "universal-analytics": "^0.4.20",
+    "universal-analytics": "^0.5.0",
```
refi93, line 68: As per the package changelog, the changes were just routine maintenance: https://github.com/peaksandpies/universal-analytics/commit/4736a56c234ad67edecbe207cee8bba7e8ff4757

PeterBenc (approved):
LGTM, thanks for fixing also the other stuff and adding the  `"import/no-extraneous-dependencies": "error"` rule. Nice :+1:

PR #1226, opened 2021-11-24
Handle optional ttl

refi93:
Motivation: I have yet to test it, but while testing bitbox02 I realized pool registrations with missing ttl were almost certainly not working. TTL became optional in Mary era: https://github.com/input-output-hk/cardano-ledger/blob/master/eras/shelley-ma/test-suite/cddl-files/shelley-ma.cddl which we already support, but for missing TTL our code would try to generate the TTL, making the tx to be actually signed inconsistent with the tx body supplied

Change: Add proper handling of null TTL so that pool registrations without TTL can work, extend related type signatures and add test case for transaction with null TTL

### DIFF
--- a/app/frontend/actions/poolOwner.ts
+++ b/app/frontend/actions/poolOwner.ts
@@ -92,7 +92,7 @@ export default (store: Store) => {
       const {plan, ttl, validityIntervalStart} = state.poolRegTransactionSummary
       const txAux = await getWallet()
         .getAccount(state.sourceAccountIndex)
-        .prepareTxAux(plan, ttl || undefined, validityIntervalStart || undefined)
+        .prepareTxAux(plan, ttl, validityIntervalStart ?? undefined)
       const witness = await getWallet()
         .getAccount(state.sourceAccountIndex)
         .witnessPoolRegTxAux(txAux)
--- a/app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
+++ b/app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
@@ -315,6 +315,14 @@ const ShelleyBitBox02CryptoProvider = async ({
 
     const ttl = txAux.ttl
     if (ttl == null) {
+      // not done via isFeatureSupported() because it would be too much hassle
+      // and we suppose the new new FW will support that together with missing multiasset
+      // support which once added, we'd bump the minimal suported version to that
+      //
+      // TODO: bump minimum supported version of Ledger/Trezor as well to the version
+      // that added multiasset and optional TTL support (the respective libs handle the
+      // feature checks for us and it's an edge-case for pool registration, that's why
+      // we aren't using isFeatureSupported() there either)
       throw new UnexpectedError(UnexpectedErrorReason.UnsupportedOperationError, {
         message: 'Optional TTL not supported by Bitbox02 wallet',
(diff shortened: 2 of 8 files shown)


app/frontend/actions/poolOwner.ts
```diff
@@ -92,7 +92,7 @@ export default (store: Store) => {
       const {plan, ttl, validityIntervalStart} = state.poolRegTransactionSummary
       const txAux = await getWallet()
         .getAccount(state.sourceAccountIndex)
-        .prepareTxAux(plan, ttl || undefined, validityIntervalStart || undefined)
+        .prepareTxAux(plan, ttl ?? undefined, validityIntervalStart ?? undefined)
```
refi93, line 95: this was working incorrectly for TTL = 0 which got modified to `undefined`

refi93, line 95: this was working incorrectly for TTL = 0 which got modified to `undefined`

app/tests/src/common/tx-settings.js
```diff
@@ -91,6 +91,65 @@ const transactionSettings = {
     ttl,
     txHash: '059cce8e6c1d404798237b36ee45f74dafb0cf463ad236def4bf305474b6a666',
   },
+  sendAdaNullTtl: {
```
refi93, line 94: worth noting that this test would pass even on current develop, the actual runtime changes were made just to the transformations of data in the crypto providers for hw wallets, but we are not testing them currently and I wouldn't introduce that just for this small modification. Still, it seems worthwhile explicitly covering that we can handle serialization/planning of a transaction with null TTL

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -299,16 +296,31 @@ const ShelleyBitBox02CryptoProvider = async ({
       })
     }
 
+    const ttl = txAux.ttl
+    if (ttl == null) {
+      // not done via isFeatureSupported() because it would be too much hassle
+      // and we suppose the new new FW will support that together with missing multiasset
+      // support which once added, we'd bump the minimal suported version to that
+      //
+      // TODO: bump minimum supported version of Ledger/Trezor as well to the version
```
refi93, line 305: not doing now, because it would be too big of a change, I'd do it as some bigger, systematic cleanup at some point

benma:
Heads up: we're about to release BitBox02 firmware v9.9.0, which allows optional TTL and also TTL=0 (https://github.com/digitalbitbox/bitbox02-api-js/pull/68). This PR will then need to be updated slightly.

I'll open a separate PR to enable token support, which also lands in v9.9.0,

benma:
> Heads up: we're about to release BitBox02 firmware v9.9.0, which allows optional TTL and also TTL=0 ([digitalbitbox/bitbox02-api-js#68](https://github.com/digitalbitbox/bitbox02-api-js/pull/68)). This PR will then need to be updated slightly.
> 
> I'll open a separate PR to enable token support, which also lands in v9.9.0.

The PR is now open: https://github.com/vacuumlabs/adalite/pull/1264 - it includes token support and also the updated typing for the ttl to allow undefined.

refi93:
note: was testing with pool registration that had the TTL removed and it still didn't work, likely needs further fixing

PR #1232, opened 2021-11-29
Disable Firefox for BitBox02 wallet

refi93:
Motivation: BitBox02 can't connect on Firefox because Adalite firewall is apparently appending the upgrade-insecure-requests header which doesn't allow connecting with the bitbox-bridge running on localhost over an insecure connection, resulting in a "Bitbox is busy" error


refi93:
not needed, fixed at firewall level

PR #1239, opened 2021-12-13
Update trezor-connect version

refi93:
Fixes issues with Catalyst registration introduced in previous trezor-connect release


gabrielKerekes (approved):
Works fine. Tested catalyst registration with FW 2.4.3 and 2.4.2. I also tested a regular transaction with both versions.

PR #1264, opened 2022-01-24
bitbox02: enable token support

benma:
The BitBox02 supports tokens since firmware v9.9.0.

### DIFF
--- a/app/bitbox02-api.d.ts
+++ b/app/bitbox02-api.d.ts
@@ -37,8 +37,14 @@ declare module 'bitbox02-api' {
     prevOutIndex: number
   }
 
+  type CardanoAssetGroupToken = {
+    assetName: Uint8Array,
+    value: string,
+  }
+
   type CardanoAssetGroup = {
     policyId: Uint8Array
+    tokens: CardanoAssetGroupToken[]
   }
 
   type CardanoOutput = {
@@ -99,7 +105,7 @@ declare module 'bitbox02-api' {
       inputs: CardanoInput[]
       outputs: CardanoOutput[]
       fee: string
-      ttl: string
+      ttl: string | null
       certificates: CardanoCertificate[]
       withdrawals: CardanoWithdrawal[]
       validityIntervalStart: string | null
--- a/app/bitbox02-api.d.ts
+++ b/app/bitbox02-api.d.ts
@@ -37,10 +37,15 @@ declare module 'bitbox02-api' {
     prevOutIndex: number
   }
 
+  type CardanoAssetGroup = {
+    policyId: Uint8Array
+  }
+
   type CardanoOutput = {
     encodedAddress: string
     value: string
     scriptConfig?: ScriptConfig
+    assetGroups: CardanoAssetGroup[]
   }
 
   type CardanoCertificate =
--- a/app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
+++ b/app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
@@ -1,4 +1,5 @@
 import {
+  CardanoAssetGroup,
   CardanoInput,
   CardanoOutput,
   CardanoCertificate,
@@ -8,6 +9,7 @@ import {
 import * as cbor from 'borc'
 import CachedDeriveXpubFactory from '../helpers/CachedDeriveXpubFactory'
 import {ShelleySignedTransactionStructured, cborizeTxWitnesses} from './shelley-transaction'
+import {orderTokenBundle} from '../helpers/tokenFormater'
 import {hasRequiredVersion} from './helpers/version-check'
 import {BITBOX02_ERRORS, BITBOX02_VERSIONS} from '../constants'
 
@@ -22,6 +24,7 @@ import {
   DerivationScheme,
   AddressToPathMapper,
   CertificateType,
+  TokenBundle,
 } from '../../types'
 import {
   CryptoProviderType,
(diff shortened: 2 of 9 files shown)


benma:
cc @refi93

app/bitbox02-api.d.ts
```diff
@@ -37,10 +37,15 @@ declare module 'bitbox02-api' {
     prevOutIndex: number
   }
 
+  type CardanoAssetGroup = {
+    policyId: Uint8Array
+  }
```
refi93, line 42: Seems like the "tokens" field is missing here

app/bitbox02-api.d.ts
```diff
@@ -94,7 +99,7 @@ declare module 'bitbox02-api' {
       inputs: CardanoInput[]
       outputs: CardanoOutput[]
       fee: string
-      ttl: string
+      ttl?: string
```
refi93, line 102:
```suggestion
      ttl: string | null
```
this would be consistent with `validityIntervalStart` field type, or is it not possible for some reason?

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
```diff
@@ -212,14 +214,27 @@ const ShelleyBitBox02CryptoProvider = async ({
...
-    if (output.tokenBundle.length > 0 && !isFeatureSupported(CryptoProviderFeature.MULTI_ASSET)) {
+  function prepareTokenBundle(tokenBundle: TokenBundle): CardanoAssetGroup[] {
+    if (tokenBundle.length > 0 && !isFeatureSupported(CryptoProviderFeature.MULTI_ASSET)) {
       throw new InternalError(InternalErrorReason.BitBox02MultiAssetNotSupported, {
-        message:
-          'Sending tokens is currently not supported on BitBox02. If you received tokens and cannot spend them, please contact support@shiftcrypto.ch',
+        message: 'Please update your BitBox02 firmware for token support.',
       })
     }
+    const orderedTokenBundle = orderTokenBundle(tokenBundle)
```
refi93, line 223: @benma does bitbox preserve the order in which the tokens are passed to the device? Also, I'd suggest ensuring in the device firmware (if you aren't already doing it) that you explicitly check for duplicate entries as it can result in a misleading information being displayed to the user (same token may be serialized/shown twice in the same tx output, but the node will evaluate only one entry AFAIK)

benma: > does Bitbox preserve the order in which the tokens are passed to the device? 

Yes. We assume the host wallet would send it in the same order in which it constructs the tx, so it should always lead to the same serialization. If the HW wallet orders it internally, the serialization could become different, right?

> which can result in valid txs with undefined behavior

Can you elaborate? Can such a tx be mined? If so, what are the semantics of it? If it cannot be mined, then confirming it on the device doesn't lead to a valid transaction.

app/bitbox02-api.d.ts
benma: good catch. it is possible - the upstream package is still not typed and both null and undefined is accepted. 

maybe changing both to undefined is better (for both), as then the fields can be missing from the object, letting the library user specify only the fields they need.

app/bitbox02-api.d.ts
benma: not sure where that went ;) thanks for fixing it! feel free to merge it like this, your change looks good to me.

fyi, the next release v9.9.1 likely out next week, will support Byron recipient addresses (up to v9.9.0, only Shelley recipient addresses were supported).

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
refi93: IOHK told us a CBOR map with duplicate keys isn't invalid at protocol level, the nodes apparently just pick the last of the keys when evaluating a CBOR map with duplicate keys - so the user would see two times the same token being sent on the hw device, but actually it would be sent only one time

app/bitbox02-api.d.ts
refi93: Will there any code changes in Adalite be needed to handle the Byron addresses with bitbox? I think not as we apparently didn't handle it anywhere so I guess it just causes the wallets with older firmware to crash, right? (We actually got a user's report related to that who later on contacted you too and got the issue solved)

benma: No changes needed, only a firmware upgrade - the BitBox02 is provided the address string and handles parsing, validating and cbor-serializing it.

app/frontend/wallet/shelley/shelley-bitbox02-crypto-provider.ts
benma: That means it is a consensus rule to pick the last of the keys? That is a bit of an unfortunate choice. if that is the case we will have to check for duplicates. Thanks for making us aware of it!

benma: Here is the PR: https://github.com/digitalbitbox/bitbox02-firmware/pull/869

refi93: > That means it is a consensus rule to pick the last of the keys? That is a bit of an unfortunate choice. if that is the case we will have to check for duplicates. Thanks for making us aware of it!

yes, that's the case unfortunately, you're welcome :)

refi93 (approved):

PR #1265, opened 2022-01-28
Improve tx submission handling

refi93:
Fixes/improvements:
1. Fix audit errors, clean up resolutions
2. Improve tx submission handling to deal properly with transactions from blockfrost (handling code 425 - mempool full), and changing UI message from "transaction failed" to "transaction error" so it's not misleading if there was an error, but the tx itself didn't necessarily fail
3. Getting pool metadata request may take very long if the metadata endpoint is unresponsive. So far, adalite's endpoint to get that was awaiting indefinitely, needlessly blocking the request and Adalite UI when initially loading the wallet (in prod it would timeout after 30 seconds, since heroku kills the request after such time) - happened for me on testnet for `odor match...` mnemonic
4. Fix to a bug in UI while withdrawing - "address verifying" was shown while signing the tx. Fixed the app state to contain the distinction between whether we are awaiting a transaction or address verification

Note: diff seems big due to the first commit where I fixed the audit errors, but the important diff is a lot smaller and can be checked by commits which are split by the thing they were fixing

### DIFF
--- a/app/frontend/wallet/blockchain-explorer.ts
+++ b/app/frontend/wallet/blockchain-explorer.ts
@@ -219,6 +219,13 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         throw new InternalError(InternalErrorReason.TransactionRejectedByNetwork, {
           message: response.Left,
         })
+      }
+      if (response.statusCode === 425) {
+        // error 425 can be returned by blockfrost API (used by adalite-backend)
+        // https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1tx~1submit/post
+        throw new InternalError(InternalErrorReason.TransactionRejectedMempoolFull, {
+          message: response.Left,
+        })
       } else if (response.statusCode === 504) {
         // Usually happens when a gateway timeout is received from the adalite-backend
         // where a gateway-timeout is sent from the load balancer after the backend not
@@ -432,7 +439,7 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         return stakePool
       }
       const poolInfo = stakePool ? await getPoolInfo(stakePool.url) : null
-      return poolInfo && ('name' in poolInfo)
+      return poolInfo && 'name' in poolInfo
         ? poolInfo
         : ({name: UNKNOWN_POOL_NAME} as HostedPoolMetadata)
     }
--- a/package.json
+++ b/package.json
@@ -71,7 +71,8 @@
     "universal-analytics": "^0.5.3",
     "webpack": "^5.62.1",
     "webpack-dev-middleware": "^5.2.1",
-    "webpack-hot-middleware": "^2.25.1"
+    "webpack-hot-middleware": "^2.25.1",
+    "ws": "^8.4.2"
   },
   "devDependencies": {
     "@types/node": "^16.11.6",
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -103,15 +104,10 @@
...
-    "redis": ">=3.1.2",
-    "hosted-git-info": ">=3.0.8",
-    "lodash": ">=4.17.21",
-    "ws": ">=7.4.6",
-    "glob-parent": ">=5.1.2",
-    "set-value": ">=4.0.1",
-    "acorn": ">=5.7.4",
-    "ansi-regex": "<6.0.0",
-    "postcss": ">=8.2.13"
+    "redisscan/redis": "^3.1.2",
```
refi93, line 107: `^` seems to be more "foolproof" than `>=`, because that ends up automatically updating the dependency to the latest version each time you run `yarn install` which can have hard to spot unintended consequences as it can happen while updating an unrelated package

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -215,7 +215,9 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         throw new InternalError(InternalErrorReason.ServerError)
       }
 
-      if (response.statusCode === 400) {
+      // 425 code is returned by blockfrost API and indicates a full mempool
+      // https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1tx~1submit/post
+      if (response.statusCode === 400 || response.statusCode === 425) {
```
xdzurman, line 220: Doesn't this result in showing the users the
```
[InternalErrorReason.TransactionRejectedByNetwork]: () =>
    'TransactionRejectedByNetwork: Submitting the transaction into Cardano network failed. We received this error and we will investigate the cause.',
```

Doesn't it make sense to create a different message for the 425 case - explaining the congestion, etc.?

server/helpers/dropSensitiveEventData.js
```diff
@@ -1,3 +1,5 @@
+// source: https://stackoverflow.com/questions/46946380/fetch-api-request-timeout
```
xdzurman, line 1: Isn't this comment supposed to be placed in `server/helpers/fetchWithTimeout.js`?

app/frontend/wallet/blockchain-explorer.ts
refi93: actually yes, sorry I somehow thought I handled it correctly but apparently forgot to get back to it after experimenting and fixing the other stuff :)

xdzurman (approved):
After implementing the 425 case -> LGTM

refi93:
@xdzurman the handling of error 425 can be checked with the mokku addon, overriding the following request:

`POST https://localhost:3000/api/txs/submit`
to return
```json
{
	"Left": "Transaction network error - {\"status_code\":425,\"error\":\"Mempool Full\",\"message\":\"Mempool is full, please try resubmitting again later.\"}",
	"statusCode": 425
}
```
(with HTTP status code 200)

app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -215,12 +215,15 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         throw new InternalError(InternalErrorReason.ServerError)
       }
 
-      // 425 code is returned by blockfrost API and indicates a full mempool
```
xdzurman: I actually liked the blockfrost API reference comment being there :D Can you please put it back?

refi93: yes, I thought that whoever wants to check it should just look up adalite-backend, but I agree putting it here explicitly is indeed more sensible

PR #1269, opened 2022-02-02
Add link below tx pending state

AndrejZelonka:
Adding link below tx pending state.

image


PeterBenc (approved):
LGTM, thanks

PeterBenc (commented):
Sorry, approved too soon and forgot to check it really, I think it looks a bit "not great". 

The whole page jumps when the link appears and creates a space between the tabs and the content

image

Not sure if this is supposed to look like this on mobile devices
image

app/frontend/components/pages/dashboard/walletOperationStatus.tsx
```diff
@@ -24,7 +24,21 @@ export const WalletOperationStatusType = () => {
...
+        <Fragment>
+          <PinkSpinner message={'Transaction pending...'} />
+          <p className="loading-helper-text">
+            Taking too long? Read why{' '}
+            <a
+              href="https://github.com/vacuumlabs/adalite/wiki/Troubleshooting#my-transaction-has-been-pending-for-a-long-time--what-should-i-do"
+              target="_blank"
+            >
+              here
+            </a>
```
PeterBenc, line 37: Just a suggestion, If we wanted to make it cool, we could start a `timer` on `useEffect` and show the message lets say after 10 seconds.

app/frontend/components/pages/dashboard/walletOperationStatus.tsx
```diff
@@ -25,19 +46,23 @@ export const WalletOperationStatusType = () => {
...
-              target="_blank"
-            >
-              here
-            </a>
-            .
-          </p>
-        </Fragment>
+        <PinkSpinner
+          message={'Transaction pending...'}
+          helperText={
```
AndrejZelonka, line 51: Seems easier to add helper text absolutely to PinkSpinner as it has hardcoded margin values for responsive designs.

app/frontend/components/pages/dashboard/walletOperationStatus.tsx
```diff
@@ -1,20 +1,41 @@
...
-const PinkSpinner = ({message}) => (
-  <div className="header-message">
-    {message.split('\n').map((line, i) => (
-      <p className="loading-message black" key={i}>
-        {line}
-      </p>
-    ))}
-    <div className="spinner pink">
-      <span />
+const DelayRendering = ({children}) => {
```
AndrejZelonka, line 8: Rather then creating condition whitin `PinkSpinner` `useEffect` I've created new component for delay rendering, Seems more elegant imho

app/public/css/768-1024px.css
```diff
@@ -32,6 +32,10 @@
     margin: 16px 0px -42px;
   }
 
+  .helper-text-wrapper {
+    margin-bottom: -37px;
```
PeterBenc, line 36: Nitpick: I think the policy is that the px should be multiple of 4 or at least an even number. If it looks worse with an even number then feel free to ignore this. Although, I think the "Taking too long" text could be a pixel or two close to the helper text and farther from the bottom.

PeterBenc (approved):
LGTM

PR #1274, opened 2022-02-08
Release 6.9.18

refi93:
Enable tokens for bitbox02 9.9.0 firmware

Tested with physical device both that for 9.8.0 token transactions fail with the expected error, prompting the user to upgrade (and non-token pass) and for 9.9.0 both transactions pass


PeterBenc (approved):
LGTM

PR #1277, opened 2022-02-24
Unescape pool names strings

AndrejZelonka:
Converting the HTML entitie in staking pool names to their corresponding characters.

Hopefully I've not missed any other place where we display pool name.
 previously (`&amp;` instead of &amp;)
image

after changes
image
image
image
image


refi93:
closing because the issue has been fixed in adalite-backend which was the source of the incorrectly formatted pool names

PR #1284, opened 2022-03-21
Update NuFi message

AndrejZelonka:
Updating Sidebar NuFi message

image

### DIFF
--- a/app/public/css/styles.css
+++ b/app/public/css/styles.css
@@ -126,6 +126,14 @@ p.hey {
   margin-left: 8px;
 }
 
+.my-6 {
+  margin: 6px 0px;
+}
+
+.mb-6 {
+  margin-bottom: 6px;
+}
+
 .space-between {
   display: flex;
   justify-content: space-between;
@@ -940,6 +948,12 @@ button[data-balloon] {
 
 .alert.nufi-announcement:before {
   background-image: url('../assets/alert_icon_nufi.svg');
+}
+
+.alert.nufi-announcement > .alert-content {
+  max-width: fit-content;
+  max-height: 300px;
+  padding-right: 5px;
 }
 
 .alert.news {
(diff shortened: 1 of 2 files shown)


app/public/css/styles.css
```diff
@@ -126,6 +126,14 @@ p.hey {
   margin-left: 8px;
 }
 
+.my-6 {
```
refi93, line 129: are these classes actually used anywhere?

app/frontend/components/common/alert.tsx
```diff
@@ -15,11 +15,12 @@ interface Props {
...
     | 'nufi-announcement sidebar'
   children: ComponentChildren
+  alertContentType?: 'nufi-announcement'
 }
 
-const Alert = ({children, alertType = 'success'}: Props) => (
+const Alert = ({children, alertType = 'success', alertContentType}: Props) => (
   <div className={`alert ${alertType}`}>
-    <div className="alert-content">{children}</div>
+    <div className={`alert-content ${alertContentType || ''}`}>{children}</div>
```
refi93, line 23: I'd reuse the alertType here - I don't see in the CSS any pre-existing rules that would cause breaking changes in the existing instances of alert

AndrejZelonka: Not really sure if you mean it like this but if I
```
<div className={`alert-content ${alertType || ''}`}>{children}</div>
```
the styles are broken.

I ended up removing it and used CSS selector for that purpose

AndrejZelonka: [ded8d40](https://github.com/vacuumlabs/adalite/pull/1284/commits/ded8d40638cd68a1d722c529951f9b6fcc199fe8)

app/public/css/styles.css
AndrejZelonka: [ded8d40](https://github.com/vacuumlabs/adalite/pull/1284/commits/ded8d40638cd68a1d722c529951f9b6fcc199fe8)

app/frontend/components/common/alert.tsx
refi93: Ah I see, probably some css rule concerning the alertType alone broke it, CSS selectors sound good to me :+1:

refi93 (approved):

PR #1288, opened 2022-03-22
Release 6.9.20

refi93:
Main changes:
* (hotfix) increase ttl of requests cache to 15 seconds to reduce repeated requests in most cases as response times from our backend have increased lastly, causing the cache to be stale prematurely
* upgrade wording of NuFi announcement

### DIFF
--- a/app/frontend/wallet/blockchain-explorer.ts
+++ b/app/frontend/wallet/blockchain-explorer.ts
@@ -73,7 +73,7 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     return 'Right' in result ? result.Right : undefined
   }
 
-  const _getAddressInfos = cacheResults(5000)(_fetchBulkAddressInfo)
+  const _getAddressInfos = cacheResults(15000)(_fetchBulkAddressInfo)
 
   async function getTxHistory(addresses: Array<string>): Promise<TxSummaryEntry[]> {
     const chunks = range(0, Math.ceil(addresses.length / gapLimit))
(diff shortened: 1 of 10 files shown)


app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -73,7 +73,7 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
     return 'Right' in result ? result.Right : undefined
   }
 
-  const _getAddressInfos = cacheResults(5000)(_fetchBulkAddressInfo)
+  const _getAddressInfos = cacheResults(15000)(_fetchBulkAddressInfo)
```
PeterBenc, line 76: I remember that I set this to 5 seconds, because a bigger catching time caused some problems, mainly, if you submitted tx and it all went through faster than 5 seconds, (15 in this case), the balance didn't update since cahced results were used. Not sure if this is still relevant, just raising it.

PeterBenc (approved):
LGTM

app/frontend/wallet/blockchain-explorer.ts
refi93: I see, this seems like an unlikely edge case right now and for the foreseeable future, I can still imagine somebody refreshing the wallet right before the tx going through which would cause a similar issue and this is an inherent problem of this simplistic caching. If the cache could be invalidated after submitting a tx that would probably solve the problem and it's worth revisiting, possibly as part of fixing the whole problem with slow requests.

I will add a comment acknowledging this problem for future reference

refi93: I also made a follow up issue in our Jira to come up with a proper solution, probably involving the aforementioned cache invalidation

PR #1289, opened 2022-03-22
Invalidate the request cache after a transaction is submitted

DavidTranDucVL:
https://vacuum.atlassian.net/browse/ADLT-1033

### DIFF
--- a/app/frontend/actions/reloadWallet.ts
+++ b/app/frontend/actions/reloadWallet.ts
@@ -24,6 +24,13 @@ export default (store: Store) => {
   const reloadWalletInfo = async (state: State): Promise<void> => {
     setWalletOperationStatusType(state, 'reloading')
     const wallet = getWallet()
+
+    // submitting transaction and not clearing cache can cause stale values
+    // user would see old balance and history, but UTXOs for planning would be up to date
+    // we are interested only in caching for the sake of not fetching
+    // multiple requestests within context of related requests
+    wallet.invalidateCache()
+
     const previousWalletReloadTime = _lastWalletReloadTime
     try {
       updateLastReloadTime()
--- a/app/frontend/helpers/cacheResults.ts
+++ b/app/frontend/helpers/cacheResults.ts
@@ -1,18 +1,32 @@
 import getHash from './getHash'
 
-const cacheResults = (maxAge: number, cache_obj: Object = {}) => <T extends Function>(fn: T): T => {
+type Cache<T> = {
+  fn: T
+  invalidate: () => void
+}
+
+const cacheResults = (maxAge: number) => <T extends Function>(fn: T): Cache<T> => {
+  let cacheObj = {}
+
   const wrapped = (...args) => {
     const hash = getHash(JSON.stringify(args))
-    if (!cache_obj[hash] || cache_obj[hash].timestamp + maxAge < Date.now()) {
-      cache_obj[hash] = {
+    if (!cacheObj[hash] || cacheObj[hash].timestamp + maxAge < Date.now()) {
+      cacheObj[hash] = {
         timestamp: Date.now(),
         data: fn(...args),
       }
     }
-    return cache_obj[hash].data
+    return cacheObj[hash].data
   }
 
-  return (wrapped as any) as T
+  const invalidate = () => {
+    cacheObj = {}
+  }
+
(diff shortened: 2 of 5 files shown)


app/frontend/wallet/blockchain-explorer.ts
```diff
@@ -214,6 +210,7 @@ const blockchainExplorer = (ADALITE_CONFIG) => {
         ...(token ? {token} : {}),
       }
     )
+    invalidateGetAddressInfosCache()
```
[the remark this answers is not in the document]
refi93: or as discussed offline, let's invalidate it from reloadWalletInfo which is already the central point to reload rendered wallet data anyway

refi93:
while testing this I found out that we are performing redundant network requests when computing tx fees in the send form - turns out our `getChangeAddress` call performs address discovery only to retrieve the first address anyway, needlessly slowing down the fee computation. I'd suggest addressing that as a follow-up

refi93 (approved):

app/frontend/actions/reloadWallet.ts
```diff
@@ -24,6 +24,7 @@ export default (store: Store) => {
   const reloadWalletInfo = async (state: State): Promise<void> => {
     setWalletOperationStatusType(state, 'reloading')
     const wallet = getWallet()
+    wallet.invalidateCache()
```
refi93, line 27: I'd clarify here with a comment why we need to invalidate and what edge-cases are we protecting against

app/frontend/helpers/cacheResults.ts
```diff
@@ -1,18 +1,32 @@
...
+    return cacheObj[hash].data
   }
 
-  return (wrapped as any) as T
+  const invalidate = () => {
+    cacheObj = {}
+  }
+
+  return {
+    function: (wrapped as any) as T,
```
PeterBenc, line 27: Nitpick: I would rather call this `fn`, function is reserved syntax word so it seems kind of strange to us it like this.

app/frontend/actions/reloadWallet.ts
PeterBenc, line 27: Haven't tested, but if I for example click the `reload` button 500x times, it would try to fetch all 500x times right? 
Comparing this to react-query, It seems to me that the invalidation should happen in (after) mutation (tx submission in this case), this is basically like calling the invalidation before querying right?

PeterBenc (commented):
Other than the comments LGTM, one suggestion about moving the invalidation to transaction submission logic.

app/frontend/actions/reloadWallet.ts
DavidTranDucVL: The problem is when the user submits tx, which times out. If we would cache it, then he will spam the reload button and can get into a state, where the transaction is submitted, but not fetched, because it will use the cached value.

Regarding fetching it 500 times, this function is already debounced.

PeterBenc: I see that the debouncing fixes the issue with clicking the reload button multiple times. I am fine with the current state, but conceptually its still seems to me that we should call the invalidation in the `waitForTxToAppearOnBlockchain` just before the reload and in the `catch` of the `submitTx` fn (or some other way which would fix the cache issue you mentioned, I guess there is a way to do it). Feel free to leave it as it is. :+1:

PeterBenc (approved):
I agree with rafo, I would comment the motivation for these changes somewhere, otherwise lgtm.

app/frontend/actions/reloadWallet.ts
refi93: Currently we rely on the whole account info being reloaded by `reloadWalletInfo` therefore I'd lean towards keeping the invalidation there (to ensure everything is fetched is sync) rather than ad-hoc invalidating the cache (which is a valid approach, but given how simplistic the caching is, I can see that getting out of hand)

refi93: basically, the purpose of caching here is to just prevent re-fetching data by calls depending on the same data while we (re)load the wallet and in that context it IMHO makes sense to manage the cache from there

PeterBenc: Makes sense, thnx :+1:

PR #1292, opened 2022-03-25
[HOTFIX] fix laggy UI

AndrejZelonka:
`Send` and `Delegate Stake`  form inputs are updating the global state.
This cause every component to re-render on every input change

Proper fix requires to move this logic from global state to components

Adding `React.memo` to skip re-rendering of render heavy components (long lists. TX history, Staking Rewards).

before:
image

after
image

EDIT:
[c8806da](https://github.com/vacuumlabs/adalite/pull/1292/commits/c8806dae3139d1f865bf11eb0e45f0b188071129)

Even better solution seems to be to to pass deep compare function into `useSelector`, as some of the child is updating the state and most likely changing the reference of store data, then the default shallow comparison does not do the trick and the Dashboard component is forced to be rerendered even though the selected data are the same.


app/frontend/components/pages/txHistory/transactionHistory.tsx
```diff
@@ -346,5 +347,8 @@ const TransactionHistory = (): h.JSX.Element => {
     </div>
   )
 }
-
-export default TransactionHistory
+// hotfix using memo https://vacuum.atlassian.net/browse/ADLT-1039
+// Send form inputs are updating the global state causing this component
+// to re-render on every input change
+// with long transactionHistory the UX is bad, typing in form is laggy
+export default memo(TransactionHistory)
```
refi93, line 354: as discussed offline, let's research whether/how this "memo" wrapper is actually response to any changes in the TransactionHistory parameters so we clearly understand this wouldn't lead to any inconsistently shown data

app/frontend/components/pages/dashboard/dashboardPage.tsx
```diff
@@ -243,22 +244,25 @@ const DashboardPage = () => {
...
+      activeAccountIndex: state.activeAccountIndex,
+      shouldShowExportOption: shouldShowExportOptionSelector(state),
+      shouldShowWantedAddressesModal: state.shouldShowWantedAddressesModal,
+      shouldShowConfirmTransactionDialog: state.shouldShowConfirmTransactionDialog,
+      shouldShowSendTransactionModal: state.shouldShowSendTransactionModal,
+      shouldShowDelegationModal: state.shouldShowDelegationModal,
+      shouldShowVotingDialog: state.shouldShowVotingDialog,
+      walletOperationStatusType: state.walletOperationStatusType,
+    }),
+    isEqual
```
refi93, line 264: I'd put here a comment explaining why is this needed and if you found the solution somewhere, I'd reference the resource

app/frontend/helpers/connect.ts
```diff
@@ -48,7 +48,7 @@ const connect: Connect = _connect
 
 const useStore: any = _useStore
 
-type UseSelector = <T>(selector: (state: State) => T) => T
+type UseSelector = <T>(selector: (state: State) => T, fn?: (a: Object, b: Object) => boolean) => T
```
refi93, line 51:
```suggestion
type UseSelector = <T>(selector: (state: State) => T, compareFn?: (a: Object, b: Object) => boolean) => T
```
I'd name it "compareFn" to make the purpose of the function clear at first sight

refi93:
only minor remarks, otherwise LGTM

PR #1299, opened 2022-04-20
Improve initial wallet loading performance for medium sized multi-account wallets

DavidTranDucVL:
We currently use many async calls in a blocking way. This can be quite slow for multi-account medium/big-sized wallets. This PR changes account discovery and account loading to use async calls in a non-blocking way.

Tested with mnemonic and Ledger HW wallet.

### DIFF
--- a/app/frontend/wallet/account-manager.ts
+++ b/app/frontend/wallet/account-manager.ts
@@ -2,7 +2,6 @@ import {Account} from './account'
 import {CryptoProvider, CryptoProviderFeature} from '../types'
 import blockchainExplorer from './blockchain-explorer'
 import {UnexpectedError, UnexpectedErrorReason} from '../errors'
-import assertUnreachable from '../helpers/assertUnreachable'
 import {makeBulkAccountIndexIterator} from './helpers/accountDiscovery'
 import * as _ from 'lodash'
 
@@ -19,7 +18,7 @@ const AccountManager = ({
   blockchainExplorer,
   maxAccountIndex,
 }: AccountManagerParams) => {
-  const accounts: Array<ReturnType<typeof Account>> = []
+  let accounts: Array<ReturnType<typeof Account>> = []
 
   function getAccount(accountIndex: number) {
     return accounts[accountIndex]
@@ -50,9 +49,8 @@ const AccountManager = ({
 
   async function discoverAccounts() {
     if (
-      maxAccountIndex === accounts.length - 1 || accounts.length > 0
-        ? !(await accounts[accounts.length - 1].isAccountUsed())
-        : false
+      maxAccountIndex === accounts.length - 1 ||
+      (accounts.length > 0 && !(await accounts[accounts.length - 1].isAccountUsed()))
     ) {
       return accounts
     }
@@ -60,10 +58,14 @@ const AccountManager = ({
     const isBulkExportSupported = cryptoProvider.isFeatureSupported(
       CryptoProviderFeature.BULK_EXPORT
     )
-    const shouldExplore =
-      config.shouldExportPubKeyBulk && config.isShelleyCompatible && isBulkExportSupported
 
-    if (!shouldExplore) {
+    const shouldExploreOnlyOne = !(
+      config.shouldExportPubKeyBulk &&
+      config.isShelleyCompatible &&
+      isBulkExportSupported
+    )
+
+    if (shouldExploreOnlyOne) {
       await exploreNextAccount()
       return accounts
     }
@@ -86,25 +88,24 @@ const AccountManager = ({
         await account.ensureXpubIsExported()
       }
 
-      const areAccountsUsed = await Promise.all(
-        newAccountBatch.map((newAccount) => newAccount.isAccountUsed())
-      )
+      const firstUnusedNewAccountIndex = (
+        await Promise.all(newAccountBatch.map((newAccount) => newAccount.isAccountUsed()))
+      ).findIndex((isUsed) => isUsed === false)
 
-      const firstUnusedIndex = areAccountsUsed.findIndex((isUsed) => isUsed === false)
-      const shouldContinueExploring = firstUnusedIndex === -1
+      const foundUnusedAccount = firstUnusedNewAccountIndex !== -1
 
       const accountsToAdd = newAccountBatch.slice(
         0,
-        shouldContinueExploring ? areAccountsUsed.length : firstUnusedIndex + 1
+        !foundUnusedAccount ? newAccountBatch.length : firstUnusedNewAccountIndex + 1
       )
-      accountsToAdd.forEach((account) => accounts.push(account))
 
-      if (!(shouldExplore && shouldContinueExploring && _accountIndexEnd < maxAccountIndex)) {
-        return accounts
+      accounts = [...accounts, ...accountsToAdd]
+      if (foundUnusedAccount || _accountIndexEnd >= maxAccountIndex) {
+        break
       }
     }
 
-    return assertUnreachable(null as never)
+    return accounts
   }
 
   async function exploreNextAccount() {
--- a/app/frontend/wallet/account-manager.ts
+++ b/app/frontend/wallet/account-manager.ts
@@ -3,6 +3,7 @@ import {CryptoProvider, CryptoProviderFeature} from '../types'
 import blockchainExplorer from './blockchain-explorer'
 import {UnexpectedError, UnexpectedErrorReason} from '../errors'
 import {makeBulkAccountIndexIterator} from './helpers/accountDiscovery'
+import {isHwWallet} from '../wallet/helpers/cryptoProviderUtils'
 import * as _ from 'lodash'
 
 type AccountManagerParams = {
@@ -18,7 +19,8 @@ const AccountManager = ({
   blockchainExplorer,
   maxAccountIndex,
 }: AccountManagerParams) => {
-  let accounts: Array<ReturnType<typeof Account>> = []
+  type Account = ReturnType<typeof Account>
+  let accounts: Array<Account> = []
 
   function getAccount(accountIndex: number) {
     return accounts[accountIndex]
@@ -33,7 +35,7 @@ const AccountManager = ({
     })
   }
 
-  async function addNextAccount(account) {
+  async function addNextAccount(account: Account) {
     await account.ensureXpubIsExported() // To ensure user exported pubkey
     const isLastAccountUsed =
       accounts.length > 0 ? await accounts[accounts.length - 1].isAccountUsed() : true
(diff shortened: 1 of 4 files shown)


app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
 
   async function discoverAccounts() {
+    if (
+      maxAccountIndex === accounts.length - 1 || accounts.length > 0
+        ? !(await accounts[accounts.length - 1].isAccountUsed())
+        : false
+    ) {
+      return accounts
+    }
+
```
refi93, line 59: I think it can be rewritten like this, which would look more natural to me:
```suggestion
    if (
      (maxAccountIndex === accounts.length - 1)
      || (accounts.length > 0 && !(await accounts[accounts.length - 1].isAccountUsed()))
    ) {
      return accounts
    }
```

app/frontend/wallet/account.ts
```diff
@@ -330,20 +330,46 @@ const Account = ({config, cryptoProvider, blockchainExplorer, accountIndex}: Acc
...
-    const shelleyAccountInfo = await getStakingInfo(validStakepoolDataProvider)
-    const stakingHistory = await getStakingHistory(validStakepoolDataProvider)
-    const visibleAddresses = await getVisibleAddresses()
-    const transactionHistory = await getTxHistory()
-    const poolRecommendation = await getPoolRecommendation(
-      shelleyAccountInfo.delegation,
-      baseAddressBalance
-    )
-    const isUsed = await isAccountUsed()
+    const promiseBatch1 = (async () => {
```
refi93, line 333: can you clarify what is the idea behind these separate batches? Why can't all the data be fetched in the single batch, provided `Promise.all` is called on the two batches anyway?

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
-      if (accountIndex === accounts.length) await addNextAccount(newAccount)
-
-      return (
-        shouldExplore &&
-        isAccountUsed &&
-        accountIndex < maxAccountIndex &&
-        (await _discoverNextAccount(accountIndex + 1))
+
+    if (!shouldExplore) {
+      await exploreNextAccount()
```
refi93, line 67: looking at the code, won't this result in discovery of additional Byron accounts when the `discoverAccounts` function is called repeatedly? If so, this doesn't seem correct to me.

Also, I find it a bit contradictory that we check for `!shouldExplorer` only to "explore" one account - looking at the code, a more appropriate name would probably be something like `shouldExploreMoreThanOne`, or define it as the opposite `shouldExploreOnlyOne` - the latter sounds a bit more natural to me. I'd rather use such verbose names than have seemingly contradictory statements like this one, leading to confusion

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+      const firstUnusedIndex = areAccountsUsed.findIndex((isUsed) => isUsed === false)
+      const shouldContinueExploring = firstUnusedIndex === -1
+
+      const accountsToAdd = newAccountBatch.slice(
+        0,
+        shouldContinueExploring ? areAccountsUsed.length : firstUnusedIndex + 1
+      )
+      accountsToAdd.forEach((account) => accounts.push(account))
+
+      if (!(shouldExplore && shouldContinueExploring && _accountIndexEnd < maxAccountIndex)) {
```
refi93, line 102: the check for `shouldExplore` looks redundant because if it was false, we would return before entering this loop, so I'd consider removing it

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+        return accounts
```
refi93, line 103: I'd consider breaking here instead and returning the accounts at the very end of the function

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+    if (!shouldExplore) {
+      await exploreNextAccount()
+      return accounts
+    }
+
+    for (const [accountIndexStart, accountIndexEnd] of makeBulkAccountIndexIterator()) {
+      const _accountIndexStart = Math.max(accountIndexStart, accounts.length)
+      const _accountIndexEnd = Math.min(accountIndexEnd, maxAccountIndex)
+
+      if (_accountIndexStart > _accountIndexEnd) {
```
refi93, line 75: how can this condition ever evaluate to true? My first impression would be that this could happen only if the data were somehow malformed - what am I missing here?

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+      const areAccountsUsed = await Promise.all(
+        newAccountBatch.map((newAccount) => newAccount.isAccountUsed())
+      )
+
+      const firstUnusedIndex = areAccountsUsed.findIndex((isUsed) => isUsed === false)
+      const shouldContinueExploring = firstUnusedIndex === -1
+
+      const accountsToAdd = newAccountBatch.slice(
+        0,
+        shouldContinueExploring ? areAccountsUsed.length : firstUnusedIndex + 1
```
refi93, line 98: I guess this is equivalent and would look more natural:
```suggestion
        shouldContinueExploring ? newAccountsBatch.length : firstUnusedIndex + 1
```

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+      )
+      accountsToAdd.forEach((account) => accounts.push(account))
```
refi93, line 100:
```suggestion
      accounts = [...accounts, ...accountsToAdd]
```

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +49,66 @@ const AccountManager = ({
...
+      for (const account of newAccountBatch) {
+        await account.ensureXpubIsExported()
+      }
+
+      const areAccountsUsed = await Promise.all(
+        newAccountBatch.map((newAccount) => newAccount.isAccountUsed())
+      )
+
+      const firstUnusedIndex = areAccountsUsed.findIndex((isUsed) => isUsed === false)
+      const shouldContinueExploring = firstUnusedIndex === -1
```
refi93, line 94: The `areAccountsUsed` intermediate structure seems a bit confusing to me, I'd find it easier to reason about a `firstUnusedNewAccountIndex` ("new account" to not confuse it with an account index in the derivation path)

```suggestion
      const firstUnusedNewAccountIndex = (await Promise.all(
        newAccountBatch.map((newAccount) => newAccount.isAccountUsed())
      )).findIndex((isUsed) => isUsed === false)

      const shouldContinueExploring = firstUnusedIndex === -1
```

refi93:
I left a few suggestions to improve overall readability

app/frontend/wallet/account-manager.ts
DavidTranDucVL: I find that more confusing, as the value holds an array of booleans and not accounts.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +48,68 @@ const AccountManager = ({
...
+
+      const firstUnusedNewAccountIndex = (
+        await Promise.all(newAccountBatch.map((newAccount) => newAccount.isAccountUsed()))
+      ).findIndex((isUsed) => isUsed === false)
+
+      const foundUnusedAccount = firstUnusedNewAccountIndex !== -1
+
+      const accountsToAdd = newAccountBatch.slice(
+        0,
+        !foundUnusedAccount ? newAccountBatch.length : firstUnusedNewAccountIndex + 1
```
refi93, line 99: nitpick, I'd find it more natural to reverse the condition
```suggestion
        foundUnusedAccount ? firstUnusedNewAccountIndex + 1 : newAccountBatch.length
```

app/frontend/wallet/account-manager.ts
```diff
@@ -46,29 +48,68 @@ const AccountManager = ({
   }
 
   async function discoverAccounts() {
+    if (
+      maxAccountIndex === accounts.length - 1 ||
```
refi93, line 52: nitpick, just in case, I'd rather check for `<=` rather than strict equality
```suggestion
       accounts.length - 1 >= maxAccountIndex ||
```

app/frontend/wallet/account-manager.ts
refi93: I see, when trimming the start and end by the batches this can indeed happen

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account.ts
DavidTranDucVL: Refactored.

app/frontend/wallet/account-manager.ts
DavidTranDucVL: Fixed.

app/frontend/wallet/account-manager.ts
```diff
@@ -42,33 +46,80 @@ const AccountManager = ({
     ) {
       throw new UnexpectedError(UnexpectedErrorReason.AccountExplorationError)
     }
-    accounts.push(account)
+
+    // HW wallets can't handle paralell requests.
```
refi93, line 50:
```suggestion
    // HW wallets can't handle parallel requests.
```

refi93:
As discussed offline, if there are no observable gains in speed with mnemonics we have, even if they have 30 accounts, I'd rather not merge these changes which make the discovery of accounts more complex. If there were actual issues with speed for such accounts, I'd advise such users to migrate to NuFi which is designed better to handle it and I'd rather consider reworking the discovery to decouple it from account loading in which case the discovery should be fast (just check for used addresses) and the accounts would be then loaded with the history in parallel

DavidTranDucVL:
Closing this PR for reasons mentioned in https://github.com/vacuumlabs/adalite/pull/1299#issuecomment-1106112086. The related task for decoupling the discovery and loading: https://vacuum.atlassian.net/browse/ADLT-1168

PR #1302, opened 2022-04-26
[ADLT-1040] Computation of fees results in address-rediscovery

DavidTranDucVL:
https://vacuum.atlassian.net/browse/ADLT-1040

How to test: write a value into the send element and observe F12/network.

### DIFF
--- a/app/frontend/actions/accounts.ts
+++ b/app/frontend/actions/accounts.ts
@@ -7,6 +7,7 @@ import loadingActions from './loading'
 import commonActions from './common'
 import {AccountInfo, AssetFamily, Lovelace} from '../types'
 import * as assert from 'assert'
+import {getChangeAddress} from '../wallet/account'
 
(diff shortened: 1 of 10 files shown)


refi93 (approved):

app/frontend/state.ts
```diff
@@ -285,4 +286,6 @@ export const getActiveAccountInfo = (state: State) => {
   assert(state.accountsInfo[state.sourceAccountIndex] != null)
   return state.accountsInfo[state.activeAccountIndex]
 }
+export const getChangeAddress = (accountInfo: AccountInfo): Address =>
```
refi93, line 289: I'd put this helper rather in `app/frontend/wallet/account.ts` file, here I find it a bit misplaced as I see the "state" as something higher-level

refi93:
one remark regarding the placement of the utility function, otherwise LGTM

PR #1306, opened 2022-04-30
Improve how we display token information

DavidTranDucVL:
https://vacuum.atlassian.net/browse/ADLT-1180

# Changes

See the following screenshots. The new version is on the left side, the old version is on the right side.

## Digital assets

digitalAssets

Previously we have shown:
- `Off-chain asset name + Ticker` or `On-chain name` in the header
- `Policy ID`,  `Fingerprint`, and `Details` in the expanded content.

This had some problems. `On-chain name` can be hard to read and isn't very helpful for the user. Changes:
- Define a new label for tokens, which is always readable. The priority is the following:
  - `Icon + off-chain name + ticker` > `Onchain name` (must be human-readable) > `Fingerprint`
- Add all information to the description. This is the same as we do it on NuFi:
  - `Name`, `Ticker`, `Policy ID`, `Asset name`, `Fingerprint`

There is one more thing that I find a little bit problematic, and that is the usage of the terminology: `Name` (off-chain) and `Asset name` (on-chain). For now, I chose to display it the same way as we do on NuFi. However, I think this is something that should be improved.

## Digital assets ellipsizing

digitalAssetsEllipsis

Fixed the issue with ellipsizing.

## Select dropdown

selectDropdown

Fallback to fingerprint in the case of non-human-readable `off-chain name`.

## Select dropdown ellipsizing

selectDropdownEllipsis

Fixed the issue with ellipsizing.

## Send asset with off-chain metadata

sendRegisteredAsset

Previously we have shown `Policy ID` as grey text in brackets after the `icon + off-chain name + ticker`. This was from times when asset `fingerprint` wasn't a thing yet. It's a better universal identifier and should be used instead.

Previously we have used the `icon + off-chain name + ticker` in the total row. In most cases, this combination is too long and went off-screen. `ticker` is good enough to convey the information. Especially because the user has seen what he selected in the dropdown and we will show him a confirmation screen with all information anyway.

Previously, the icon was not visually vertically aligned with the text.

## Send asset without off-chain metadata with non-human-readable name

sendTokenEllipsis

Fixed the issue with ellipsizing in the dropdown selection element and total row.

No point in showing anything after the `fingerprint` in gray in brackets, as opposed to the assets with `off-chain` metadata.

## Transaction review

transactionReview

Make the information consistent with the information shown on the `Digital assets` screen. That being, show `token name + ticker`, `token policy ID`, `token asset name`, `fingerprint` in the same order.

Previously we have shown an icon next to the `token name`, which wasn't always an `off-chain name`. Which can cause confusion as on-chain data never contain icons. Now the icon (which is always off-chain) is next to the `off-chain name`, which is nullable (`N / A`).

## Transaction history

txHistory

and

txHistoryAssetsWithOffchainData

Previously we have shown `non-human-readable on-chain names` or more readable `icon + off-chain name + ticker`. Always followed by `policy ID` and `fingerprint`. In the case of a non-human-readable name, there is no reason to show it. `fingerprint` already acts as a universal identifier, so no reason for `Policy ID` either. Now we show only `icon + off-chain name + ticker` (if exists) and always `Fingerprint`.

## Transaction history ellipsizing

txHistoryEllipsis

Fixed issue with ellipsizing.

### DIFF
--- a/app/frontend/components/common/asset.tsx
+++ b/app/frontend/components/common/asset.tsx
@@ -10,6 +10,10 @@ import {assetNameHex2Readable} from '../../wallet/shelley/helpers/addresses'
 import printTokenAmount from '../../helpers/printTokenAmount'
 import {createTokenRegistrySubject} from '../../../frontend/tokenRegistry/tokenRegistry'
 
+const isHumanReadable = (value: string): boolean =>
+  // eslint-disable-next-line no-useless-escape
+  !!value && /^[a-zA-Z0-9!"#$%&'()*+,.\/:;<=>?@\[\] ^_`{|}~-]*$/.test(value)
+
 type LinkToAssetProps = {
   policyIdHex: string
   assetNameHex: string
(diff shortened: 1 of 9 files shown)


app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -255,17 +255,19 @@ const SendAdaPage = ({
...
   const displayDropdownSelectedItem = (dropdownAssetItem: DropdownAssetItem) => {
-    const {policyId, type} = dropdownAssetItem
+    const {fingerprint, type} = dropdownAssetItem
     return (
-      <div className="wrapper">
+      <div className="wrapper flex-nowrap shrinkable">
         <FormattedAssetItem {...dropdownAssetItem}>
-          {({formattedAssetIconName}) => {
+          {({formattedHumanReadableLabelType, formattedHumanReadableLabelWithIcon}) => {
+            const isFingerprint =
```
refi93, line 263: this seems a bit confusing - so the `fingerprint` passed may not always be a fingerprint? :thinking: I'd suggest choosing some more accurate name for that property if that's the case

app/frontend/components/common/asset.tsx
```diff
@@ -35,8 +46,12 @@ export const FormattedAssetItem = ({
 }: FormattedAssetItemProps & {
   children: (props: {
     icon: h.JSX.Element
-    formattedAssetName: h.JSX.Element | string
-    formattedAssetIconName: h.JSX.Element
+    formattedHumanReadableLabelType: FormattedHumanReadableLabelType
```
refi93, line 49: I'd consider grouping these "formattedHumanReadableLabel" under a single key which could be a union type with clearly defined fields depending on the type, as it's otherwise pretty hard to follow which of the properties apply when

app/frontend/components/common/asset.tsx
```diff
@@ -46,11 +61,13 @@ export const FormattedAssetItem = ({
...
   }) => h.JSX.Element
 }) => {
+  const missingValue = 'N / A'
+
   const metadata = useSelector((state) =>
     state.tokensMetadata.get(createTokenRegistrySubject(policyId, assetName))
   )
 
-  const Icon = () => {
+  const icon = (() => {
```
refi93, line 70: this is still a JSX element soI don't see a reason to rename it and not use it as `<Icon/>` in the code, or am I missing something? I'd also expect this element to receive the metadata as a parameter rather than using the `metadata` variable declared in the upper level scope, which seems pretty dirty even in the original code

app/frontend/components/common/asset.tsx
```diff
@@ -68,57 +85,122 @@ export const FormattedAssetItem = ({
...
+      labelWithIcon: (
+        <LabelWrapper>
+          <div className={`${styles.icon} ${styles.iconOffset}`}>{icon}</div>
+          {label}
+        </LabelWrapper>
+      ),
+      shortLabelWithIcon: (
+        <LabelWrapper>
+          <div className={`${styles.icon} ${styles.iconOffset}`}>{icon}</div>
+          {labelShort ?? label}
```
refi93, line 165: looking at the code above, isn't `labelShort` always defined, making the null-coalescing here redundant?

```suggestion
          {labelShort}
```

app/frontend/components/common/asset.tsx
```diff
@@ -68,57 +85,122 @@ export const FormattedAssetItem = ({
...
-    }
-    return '<no name>'
-  }
+  const {
+    labelType: formattedHumanReadableLabelType,
+    label: formattedHumanReadableLabel,
+    labelWithIcon: formattedHumanReadableLabelWithIcon,
+    shortLabelWithIcon: formattedShortHumanReadableLabelWithIcon,
+  } = (() => {
+    const {labelType, label, labelShort} = (() => {
```
refi93, line 96: this nested function makes it quite a pain to follow what's going on. I'd consider extracting it into a `getLabelVariants` functions and the parent function, extracted (to make the `FormattedAssetItem` component easier to follow as it's getting really chunky) and renamed to `getFormattedLabelVariants` could be using it internally

app/frontend/components/common/asset.tsx
refi93: tbh, this passing of formatted components to the children instead of data to to handle the formatting internally seems like an antipattern to me, making the reasoning about what exactly gets rendered a pain, on the other hand, I understand refactoring it would require even more effort without that much added value as I don't expect much more new data/rendering logic to be added in the foreseeable future - maybe data about NFTs but that's what I'd consider the tipping point to refactor/redesign this together with the overall UI more radically

app/frontend/components/pages/sendAda/confirmTransactionDialog.tsx
```diff
@@ -88,34 +88,43 @@ const SendAdaReview = ({
...
+            }) => {
               return (
                 <Fragment>
-                  <div className="review-label">Token fingerprint</div>
-                  <div className="review-amount">{formattedFingerprint}</div>
-                  <div className="review-label">Token policy Id</div>
-                  <div className="review-amount">{token.policyId}</div>
                   <div className="review-label">Token name</div>
-                  <div className="review-amount">{formattedAssetIconName}</div>
+                  <div className="review-value flex-nowrap">
```
refi93, line 104: Looking at this screenshot

It seems as we were falling back to the onchain name even when showing the `formattedOffChainName` which seems misleading to me. Shouldn't we show just `N/A` here in case there is no offChain name?

refi93: ok nevermind, I was looking at the wrong side of the picture :)

refi93:
as discussed offline, let's replicate NuFi UI when it comes to showing tokens so we don't have to reinvent/review the whole UI over again which would be counterproductive

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: The passed value is indeed a fingerprint. But normally we'd show the fingerprint in the grey text in brackets after the label. But if the label is already a fingerprint, there is no point in showing it again after it in the grey text in brackets. I'll change it to `isLabelFingerprint`, hopefully, that's less confusing.

app/frontend/components/common/asset.tsx
DavidTranDucVL: Refactored

app/frontend/components/common/asset.tsx
DavidTranDucVL: Refactored

app/frontend/components/common/asset.tsx
DavidTranDucVL: Refactored

app/frontend/components/common/asset.tsx
DavidTranDucVL: Refactored according to the first comment.

app/frontend/components/pages/sendAda/sendAdaPage.tsx
```diff
@@ -259,13 +259,13 @@ const SendAdaPage = ({
...
-            const isFingerprint =
-              formattedHumanReadableLabelType === FormattedHumanReadableLabelType.FINGERPRINT
+          {({formattedHumanReadableLabelVariants}) => {
+            const {type: labelType, label} = formattedHumanReadableLabelVariants
+            const isLabelFingerprint = labelType === FormattedHumanReadableLabelType.FINGERPRINT
             return (
               <Fragment>
-                {formattedHumanReadableLabelWithIcon}
-                {type === AssetFamily.TOKEN && !isFingerprint && (
+                {label}
```
refi93, line 267: given there was originally `formattedHumanReadableLabelWithIcon`, shouldn't this be the `withIcon` variant? Btw, I'd consider naming the `withIcon` property `labelWithIcon` for the sake of readability when used alone, same for `shortWithIcon`

refi93 (approved):

app/frontend/components/pages/sendAda/multiAssetsPage.tsx
```diff
@@ -71,36 +111,19 @@ const MultiAssetsPage = () => {
...
+                  <Item
+                    title="Policy ID"
+                    displayElement={formattedPolicy!}
+                    copyValue={asset.policyId}
+                  />
+                  <Item title="Asset name" displayElement={formattedOnChainName!} />
+                  <Item
+                    title="Fingerprint"
+                    displayElement={formattedFingerprint!}
+                    copyValue={asset.fingerprint!}
```
refi93, line 125: I'd find it convenient to add a link here to the block explorer so the user can check there details about the token and gain confidence they are indeed sending the right one

refi93: on the other hand, the user can do so from the list of assets they hold, and it's more natural to do it from there, so after a second thought, there isn't that much added value to it indeed

app/frontend/components/pages/sendAda/sendAdaPage.tsx
DavidTranDucVL: Yes, it should be with an icon, fixed. Also refactored labels.

PR #1312, opened 2022-05-20
Export: add from address & txsummary with from/to

szist:
Adds received from to exports. For now choosing the first non-wallet address.
E.g. motivation: ability to gather all the addresses that would sent donations to a given account.

Extends tx summary with the from/to addresses.

### DIFF
--- a/app/frontend/components/pages/txHistory/transactionHistory.tsx
+++ b/app/frontend/components/pages/txHistory/transactionHistory.tsx
@@ -13,6 +13,7 @@ import {
   TxSummaryEntry,
   Token,
   AssetFamily,
+  Address,
 } from '../../../types'
 import {AdaIcon} from '../../common/svg'
 import {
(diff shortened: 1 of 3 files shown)


DavidTranDucVL (approved):
LGTM.

Regarding UX, wouldn't it make sense to include the `Received from` in each row of the TX? We do this for `Transaction ID`. Currently, token rows, which are always multirow entries related to one TX, don't contain this `Received from` address.

Also, I was under the impression, that one TX could contain inputs from "different wallets" as long as it includes the signature of related actors. So that might cause some weird edge cases with the current approach.

refi93 (approved):

refi93:
> Also, I was under the impression, that one TX could contain inputs from "different wallets" as long as it includes the signature of related actors. So that might cause some undesirable edge cases with the current approach.

Fair point, I added a disclaimer to the column titile pointing out this edge-case

PR #1327, opened 2022-07-25
Migrate amounts to bignumber [ADLT-1498]

refi93:
## Motivation:

Wallet was failing to send bigger amounts of tokens than `MAX_SAFE_INTEGER` (2^53 - 1). It indeed should fail with an error and such tx shouldn't be created/submitted in the first place (one of the commits is fixing that check), but it's time to finally migrate anyway as any such wallet with too big amount of tokens may become unusable as such utxos couldn't be serialized as change outputs in the tx.

Therefore, apart from fixing the original bug (incorrect validation of max amounts in tx plan), this PR is migrating to amounts represented as big numbers. To have an easier time updating the tests, I also refactored the tests to TS in a separate commit.

Then, while testing the app I noticed several bugs which are being fixed in standalone commits:
* missing stake pool info in delegation tx summary screen
* sign button in pool registration modal wasn't disabled while signing on hw wallet
* other minor issues, check commits

## How to review:
The second commit contains most of the changes and it's pretty mechanical so it's IMHO not worth reviewing it line by line, rather just glancing over it to get what went on. After that I updated the tests, still with old tx fee computation that didn't allocate bytes for MAX_UINT64, just MAX_SAFE_INTEGER - these should run without issues, and only then I made a commit to update fee computation and update tests accordingly. Then I left aside commits which fixed bugs that I noticed after the migration plus some preexisting bugs - these are very short and should be pretty straightforward to review

## Testing:

try signing/delegation/voting with mnemonic/ledger/trezor/bitbox - I did so myself already and should work fine by now


server/helpers/parseTxBody.js
```diff
@@ -1,6 +1,6 @@
-const cbor = require('borc') // TODO parse bignumbers properly
+const cbor = require('cbor') // TODO parse bignumbers properly
 
-const coinsFromOutputAmount = (outputAmount) =>
+const coinsFromOutputValue = (outputAmount) =>
```
xdzurman, line 3: Since we're changing the function name here I'd also rename the function parameter to outputValue

server/helpers/parseTxBody.js
```diff
@@ -12,8 +12,14 @@ const parseTxBodyOutAmount = (txBody) => {
    * amount = coins | [coins, multiasset]
    */
   const firstTxOutput = cbor.decode(txBody)[0].get(1)[0]
-  const outputAmount = firstTxOutput ? firstTxOutput[1] : 0
-  return coinsFromOutputAmount(outputAmount)
+  // see https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl#L75
+  const parseValueFromOutput = (out) =>
```
xdzurman, line 16: The function parameter here does not seem to be used. 
Wouldn't it be sufficient to write this as:
```ts
  const parsedValueFromOutput =
    firstTxOutput instanceof Map
      ? firstTxOutput.get(1) // babbage-era
      : firstTxOutput[1] // pre-babbage

  const outputValue = firstTxOutput ? parsedValueFromOutput : 0
  return coinsFromOutputValue(outputValue)
```

xdzurman:
image
image
Not sure if introduced in this PR, most likely not, just throwing this in here

server/helpers/parseTxBody.js
refi93: fixed :+1:

server/helpers/parseTxBody.js
refi93: fixed :+1: It also puzzled me why didn't eslint warn me about such basic mistake and turns out the `no-unused-vars` eslint rule wasn't configured to check for unused function parameters, so I enabled it

refi93:
> Not sure if introduced in this PR, most likely not, just throwing this in here (seen when sending a `stake delegation` tx)

Turned out to be caused by both the button content and the tooltip being both managed in CSS - this seems to be only instance of it in the code and I solved it by putting the tooltip in a wrapper span instead of the button itself

PeterBenc (approved):
I tested a couple of transactions with a mnemonic/ledger. Looks good except for the UI bugs Kamil already posted.

refi93:
> Text is overflowing in case of big amounts

@xdzurman solved in https://github.com/vacuumlabs/adalite/pull/1327/commits/69bf65bb8d9f4c147131c14e615afb65649a1ec4

server/helpers/parseTxBody.js
refi93: addressed in https://github.com/vacuumlabs/adalite/pull/1327/commits/bfb1a66bcc25e55ad1238650d1df6185019dae9f

xdzurman:
image
image
Not related to other comments but - these tooltips don't work, did they work before the span wrapper? (Checked - they don't work on https://testnet.adalite.io either so don't consider this as a requirement for this PR to be merged)

xdzurman (approved):

refi93:
> Not related to other comments but - these tooltips don't work, did they work before the span wrapper? (Checked - they don't work on https://testnet.adalite.io/ either so don't consider this as a requirement for this PR to be merged)

Yeah this "tooltip" is actually just the default icon for tokens that don't have one and I got confused as well. To remove that confusion, I'm replacing it with an empty grey circle

PR #1335, opened 2022-08-03
Add top-level error boundary [ADLT-1518]

refi93:
## Problem

Errors thrown while rendering a component were blocking rendering until now, effectively freezing the app which is very bad UX, plus such errors wouldn't have a chance to reach us without somebody proactively reporting it which defeats the purpose of error tracking.

## Solution

This PR adds a top-level "ErrorBoundary" component which intercepts any errors during rendering and falls back to the unexpected error modal, and if the error is caught at that level, we tell the error modal to refresh the page after being closed

## Testing

Try adding some test errors anywhere in the components/actions, like `throw new Error('xxx')` and check that a proper error modal is shown. e.g. an error thrown in router.tsx would previously result in a blank page and now the error modal should be shown.

Note: to be able to see the error modal, don't forget to set `ADALITE_SENTRY_DSN_WEB` env variable

### DIFF
--- a/app/frontend/components/common/unexpectedErrorModal.tsx
+++ b/app/frontend/components/common/unexpectedErrorModal.tsx
@@ -5,13 +5,19 @@ import actions from '../../actions'
 import Modal from './modal'
 import Alert from './alert'
 import submitFeedbackToSentry from '../../helpers/submitFeedbackToSentry'
+import * as Sentry from '@sentry/browser'
 
 interface Props {
   sendSentry: any
   closeUnexpectedErrorModal: () => void
+  reloadPageOnClose: boolean
 }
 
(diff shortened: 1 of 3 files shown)


app/frontend/components/common/unexpectedErrorModal.tsx
```diff
@@ -31,8 +37,14 @@ const UnexpectedErrorModal = ({sendSentry, closeUnexpectedErrorModal}: Props) =>
         sendSentry.resolve(false)
       }
       closeUnexpectedErrorModal()
+
+      if (reloadPageOnClose) {
+        Sentry.close(5000).then(() => {
```
refi93, line 42: this ensures the exception gets actually submitted before reloading the page

PeterBenc (approved):
Tested, works, looks good, thanks :+1:

AndrejZelonka (approved):

PR #1342, opened 2022-08-26
deps: @trezor/connect v9

mroz22:
Hi there!

We have released the first non-beta version of @trezor/connect v9. Could you please have a look if it works? I was able to run application on localhost with it and do some cardano operations against trezor device

Connect v9 was mostly introduced due to refactoring from flow to typescript. There shouldn't be many breaking changes https://github.com/trezor/trezor-suite/blob/develop/packages/connect/CHANGELOG.md


Hannsek:
Hello, 
Just letting you know that we have released Connect v9. It is now GA. https://www.npmjs.com/package/@trezor/connect/v/9.0.0

refi93 (approved):

refi93:
thank you very much @mroz22 @Hannsek , merged to develop

PR #1345, opened 2022-09-09
Fix coin selection for Byron addresses with tokens

refi93:
Motivation: there was a Byron address with many token utxos where the user struggled to send an ADA tx because the resulting tx was too big. Turns out our coin selection algorithm was prioritizing "non-staking" utxos regardless of whether they had tokens or not, causing the tx plan to build unnecessarily big txs when trying to send ADA out of such address.

Changes: This PR combines staking and non-staking utxos into a single list before splitting it into the token and ada utxos, adding proper priority even to "non-staking" utxos when sending ADA out of the wallet

How to test: Try sending ADA out of the review app to check tx building still works


DavidTranDucVL (approved):
LGTM

PR #1356, opened 2022-10-27
Update min ADA calculation formula

DavidTranDucVL:
Jira issue: [ADLT-1697](https://vacuum.atlassian.net/browse/ADLT-1697)

**Motivation**
Update min ADA calculation formula to match latest changes introduced in Babbage era.

**Previous behavior**
Based on:
- https://github.com/input-output-hk/cardano-ledger/blob/8817304c300dd60099054dd66d9a1ecbd33bfb3a/doc/explanations/min-utxo-mary.rst

**New behavior**
Based on:
 - https://cips.cardano.org/cips/cip55/#thenewminimumlovelacecalculation
 - https://hydra.iohk.io/build/15339994/download/1/babbage-changes.pdf page 9

**Changes**
Update function that implements mentioned formula.

**How to test**
Try making a transaction with tokens and without tokens, try using max button, try making transaction with 1 ADA.

**Checklist**

- [x] Branch name format is similar to `adlt-594-add-pr-template`
- [x] PR name format is similar to `Add PR template [ADLT-594]`
- [x] Commit messages are capitalized, use the imperative mood and their subject line is limited to 50 characters
- [x] You are set as the assignee
- [x] Reviewers are set
- [x] Appropriate labels are added
- [x] I have pointed out weird/non-standard changes (if any) by commenting on them
- [ ] QA signoff requested (if applicable)

[ADLT-594]: https://vacuum.atlassian.net/browse/ADLT-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/frontend/wallet/shelley/transaction/utils.ts
+++ b/app/frontend/wallet/shelley/transaction/utils.ts
@@ -1,53 +1,35 @@
 import BigNumber from 'bignumber.js'
-import distinct from '../../../helpers/distinct'
 import {Address, CertificateType, Lovelace, TokenBundle} from '../../../types'
-import {aggregateTokenBundles} from '../../helpers/tokenFormater'
+import {encodeCbor} from '../../helpers/cbor'
 import {TxCertificate, TxPlanAuxiliaryData, TxInput, TxOutput, TxWithdrawal} from '../../types'
-import {MIN_UTXO_VALUE} from './constants'
+import {cborizeSingleTxOutput} from '../shelley-transaction'
 import {estimateTxSize} from './estimateTxSize'
 
-export const computeMinUTxOLovelaceAmount = (tokenBundle: TokenBundle): Lovelace => {
-  // based on https://github.com/input-output-hk/cardano-ledger-specs/blob/master/doc/explanations/min-utxo.rst
-  const quot = (x: number, y: number) => Math.floor(x / y)
-  const roundupBytesToWords = (x: number) => quot(x + 7, 8)
-  // TODO: this to network config or constants
-  const minUTxOValue = new BigNumber(MIN_UTXO_VALUE)
-  // NOTE: should be 2, but a bug in Haskell set this to 0
-  const coinSize = 0
-  const txOutLenNoVal = 14
-  const txInLen = 7
-  const utxoEntrySizeWithoutVal = 6 + txOutLenNoVal + txInLen // 27
+export const computeMinUtxoLovelaceAmount = (
+  address: Address,
+  tokenBundle: TokenBundle
+): Lovelace => {
+  // Reference:
+  // - https://cips.cardano.org/cips/cip55/#thenewminimumlovelacecalculation
+  // - https://hydra.iohk.io/build/15339994/download/1/babbage-changes.pdf page 9
+  const OVERHEAD = 160
 
-  // NOTE: should be 29 but a bug in Haskell set this to 27
-  const adaOnlyUtxoSize = utxoEntrySizeWithoutVal + coinSize
+  // taken from https://cexplorer.io/params
+  const COINS_PER_UTXO_WORD = 34482
+  const COINS_PER_UTXO_BYTE = Math.floor(COINS_PER_UTXO_WORD / 8)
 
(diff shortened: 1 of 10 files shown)


app/frontend/wallet/shelley/transaction/utils.ts
```diff
@@ -1,53 +1,35 @@
...
 
-  // NOTE: should be 29 but a bug in Haskell set this to 27
-  const adaOnlyUtxoSize = utxoEntrySizeWithoutVal + coinSize
+  // taken from https://cexplorer.io/params
+  const coinsPerUTxOByte = 4310
 
-  // this ensures there are only distinct (policyId, assetName) pairs
-  const aggregatedTokenBundle = aggregateTokenBundles([tokenBundle])
+  // 80kb is current max block size
+  const coinsUpperBound = new BigNumber(80 * 1024 * coinsPerUTxOByte) as Lovelace
```
gabrielKerekes, line 21: It's not clear to me why `coinsUpperBound` depends on max block size. Could you please explain? I didn't find any explanation in the referenced docs.

app/frontend/wallet/shelley/transaction/utils.ts
```diff
@@ -1,53 +1,35 @@
...
+  const txOutput: TxOutput = {
+    isChange: false,
+    address,
+    coins: coinsUpperBound,
+    tokenBundle,
+  }
 
-  const size =
-    6 + roundupBytesToWords(numAssets * 12 + sumAssetNameLengths + numPIDs * policyIdSize)
+  const serSizeTxOut: number = encodeCbor(cborizeSingleTxOutput(txOutput)).length + 1
```
gabrielKerekes, line 30:
```suggestion
  const serializedTxOutputSize: number = encodeCbor(cborizeSingleTxOutput(txOutput)).length + 1
```

(nit: is the explicit type needed?)

app/frontend/wallet/shelley/transaction/utils.ts
```diff
@@ -1,53 +1,35 @@
...
+): Lovelace => {
+  // Reference:
+  // - https://cips.cardano.org/cips/cip55/#thenewminimumlovelacecalculation
+  // - https://hydra.iohk.io/build/15339994/download/1/babbage-changes.pdf page 9
+  const overhead = 160
 
-  // NOTE: should be 29 but a bug in Haskell set this to 27
-  const adaOnlyUtxoSize = utxoEntrySizeWithoutVal + coinSize
+  // taken from https://cexplorer.io/params
+  const coinsPerUTxOByte = 4310
```
gabrielKerekes, line 18: I'd perhaps rather declare it as:
```
const coinsPerUtxoWord = 34482
const coinsPerUtxoByte = coinstPerUtxoWord / 8
```
Since the param currently doesn't really exist (the explorer is showing it as `coins_per_utxo_size` and perhaps I'd rather reference [this](https://book.world.dev.cardano.org/environments/mainnet/alonzo-genesis.json) as a source, what do you think?

gabrielKerekes (approved):
Minor suggestions and one question otherwise LGTM - but as mentioned offline, I don't have much AdaLite related context.

app/frontend/wallet/shelley/transaction/utils.ts
refi93: my understanding is that we want to use the max possible amount of coins as a placeholder and I don't see a relation with the block size either. In the CDDL the type of "coin" is "uint" which I didn't find a direct definition for, but I'm pretty sure it's roughly 2^64-1 which is also the limit HW wallets work with

app/frontend/wallet/shelley/transaction/utils.ts
refi93: encodeCbor is actually not properly typed, returning `any` - a proper fix would be to add an explicit type to that function and then this casting would no longer be needed it, I'll do it

refi93 (approved):

PR #1361, opened 2022-12-16
Switch catalyst registration reward destination from stake to base address

refi93:
## Motivation

The initial motivation was to change the address to get rewards from user's staking address to their first base address, because that's how the registrations will change in the next catalyst round. While I was at that, I also made a bunch of other updates/fixes described below

## Changes
* change catalyst registration reward address from stake address to base address
* update ledger/trezor accordingly and update ledger to prerelease lib to be able to handle it once it's launched
* update from Node 14 to Node 18 (note that so far the unit tests were failing, turns out the solution was to run chrome headless with the `--no-sandbox` flag)
* update cypress to latest version, i.e. 12, because v9 was failing with some weird ssl error on Node 18

## How to test
* try creating a catalyst registration with mnemonic/trezor/ledger


app/cypress.config.ts
```diff
@@ -0,0 +1,20 @@
+import {defineConfig} from 'cypress'
+
```
refi93, line 2: cypress 12 detected the old version files and offered me to migrate them automatically which involved renaming the test files and config for cypress - this is the result and it seems to work without issues

PeterBenc (approved):
LGTM, thanks :+1: tested with mnemonic and ledger, seems to be working

PR #1371, opened 2023-04-06
Support Redis 7 [ADLT-2074] (WIP)

refi93:


refi93:
not needed, redis 6.2 will still still be supported

PR #1384, opened 2023-08-10
Promote catalyst from login page

AndrejZelonka:
Instructions:
image

PR #1388, opened 2023-10-12
Revert catalyst promotio banner

AndrejZelonka:
reverting https://github.com/vacuumlabs/adalite/pull/1384/commits/1211bc2fc8a211c3bcc7174bd35d22e4ab08b8f3


PeterBenc (approved):
LGMT, btw does this commit only revert that commit as a whole or only partially? Since you haven't used the git revert functionality I had to go through the changes not being sure if its full revert or only partial :thinking: 

In case its only partial revert, lets say you wanted to keep some nice refactor from that commit, I would first do the basic revert, and then add that refactor again in the next commit so its obvious what got added 

At least I would do it like that

AndrejZelonka:
@PeterBenc yes, since you are the co-author of the commit I wasn't sure if there was added something from your side, and I went line-by-line to achieve desired state. At the end, it was basically reverted as a whole. 

I've pushed the proper revert commit.

PR #1389, opened 2023-10-13
Fix audit

PeterBenc:


refi93 (changes requested):
resolutions aren't needed, running `npx yarn-audit-fix` which would touch just `yarn.lock` (respecting `package.json`) is enough - tried locally

refi93:
fix already pushed directly to develop, closing

PR #1395, opened 2024-01-10
Catalyst news 10.1.2024

PeterBenc:


app/frontend/localStorage.ts
```diff
@@ -1,6 +1,6 @@
 const localStorageVars = {
   WELCOME: 'dontShowDisclaimer',
-  INFO_MODAL: 'dontShowInfoModal_10.08.2023',
+  INFO_MODAL: 'dontShowInfoModal_01.10.2024',
```
refi93, line 3: nitpick, we usually use here the european format
```suggestion
  INFO_MODAL: 'dontShowInfoModal_10.01.2024',
```

package.json
```diff
@@ -111,6 +111,7 @@
     "eslint/**/glob-parent": "^5.1.2",
     "eslint/**/ansi-regex": "^5",
     "prettier-eslint-cli/**/ansi-regex": "^5",
-    "isomorphic-fetch/**/node-fetch": "^2.6.7"
+    "isomorphic-fetch/**/node-fetch": "^2.6.7",
+    "browserify-sign": "4.2.2"
```
refi93, line 115: I checked and `yarn-audit-fix` is able to resolve this without resorting to resolutions which is unnecessarily fragile

refi93 (approved):

PR #1403, opened 2024-03-18
limit showing long token lists/tx history by default [ADLT-2545]

refi93:
## Motivation

Wallets with many assets (thousands) and very long tx history may currently slow down the wallet UI to the point of being unusable/crashing the website. This PR fixes the issue by limiting the amount of assets/tx history items shown by default to 50 and showing more only if user explicitly clicks on "show more" at the bottom of the trimmed list

Moreover, some backend requests on such big wallets may take several tens of seconds to load which causes unnecessary cache misses in our simplistic time-based cache, because functions depending on such slow queries may cause their fetch multiple times if some of them awaited the result for more than 15s (current cache time). Given we invalidate the cache after transactions and allow that also when user clicks refresh-button, it seems fair enough to increase the query cache time to 60s which should solve most problems in practice. Of course, we should speed up our backend queries as well, sooner or later, this is just a hotfix

## How to test

* Try with our test wallet which already has lots of tokens/txs, check that wallet behaves as expected
* Check that tx history/balance is properly updated after performing a transaction/refreshing the wallet

### DIFF
--- a/app/frontend/components/pages/delegations/stakingHistoryPage.tsx
+++ b/app/frontend/components/pages/delegations/stakingHistoryPage.tsx
@@ -15,8 +15,8 @@ import {
   RewardWithdrawal,
   StakingKeyRegistration,
 } from '../../../types'
-import Alert from '../../common/alert'
 import {useActiveAccount} from '../../../selectors'
+import {useState} from 'preact/hooks'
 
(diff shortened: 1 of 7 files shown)


xdzurman (approved):
nice, see UX of the button please

PR #1416, opened 2024-07-03
Update Trezor lib & devices

martykan:
This PR includes 2 changes: 
1. Updating `@trezor/connect-web` to the latest release 9.3.0. This version should resolve issues with the connection process that we were seeing with Adalite in particular.  
2. Add mentions of our new device - Trezor Safe 5 and update the device image


refi93:
Great thanks @martykan!

> This version should resolve issues with the connection process that we were seeing with Adalite in particular.

could you please elaborate what kind of issues were you facing so that we can try reproduce and understand impact on users?

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
```diff
@@ -63,9 +63,12 @@ const ShelleyTrezorCryptoProvider = async ({
 }: CryptoProviderParams): Promise<CryptoProvider> => {
   const derivationScheme = derivationSchemes.v2
 
-  TrezorConnect.manifest({
-    email: ADALITE_SUPPORT_EMAIL,
-    appUrl: config.ADALITE_SERVER_URL,
+  await TrezorConnect.init({
```
refi93, line 66: If on the first try user doesn't connect this may fail with a `TrezorConnect has been already initialized`.

Is there perhaps a way to check whether Trezor connect has been initialized, avoiding the duplicate initialization/error?

Alternatively, I guess it could be good enough to just catch this error if it happened, because looking at the function, the double initialization shouldn't have any side-effects beyond throwing the error: https://github.com/trezor/trezor-suite/blob/231bbff772395f5b9522222f6d451de55a47ddda/packages/connect/src/index.ts#L111

what do you think @martykan ?

refi93 (changes requested):
see comment on "already initialized" error, otherwise LGTM 👍

app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts
martykan: You're right, good catch. I added a call to the dispose method first

refi93 (approved):

PR #1452, opened 2025-03-16
fix: replace express mung

marklai1998:
[express-mung](https://github.com/richardschneider/express-mung) is stale and hasn't been updated for years.

I've forked the project and updated it to modern stack

### DIFF
--- a/server/middlewares/statsRedis.js
+++ b/server/middlewares/statsRedis.js
@@ -5,7 +5,7 @@ const client = redis.createClient(process.env.REDIS_URL, {
     rejectUnauthorized: false,
   },
 })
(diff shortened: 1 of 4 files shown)


server/middlewares/statsRedis.js
```diff
@@ -38,7 +38,8 @@ const trackVisits = (req, res, next) => {
   next()
 }
 
-const trackTxSubmissions = mung.json((body, req) => {
+const trackTxSubmissions = jsonMiddleware((body, req, res) => {
+  if (res.statusCode >= 400) return body
```
marklai1998, line 42: original express-mung behavior

PR #1454, opened 2025-06-09
Update trezor connect

refi93:
Update trezor connect to v9.6.0 to make AdaLite ready for new trezor connect solution

Test that connecting with Trezor still works and tx sending works as well


PeterBenc (approved):
Tested, works, LGTM

PR #1472, opened 2026-03-26
feat: Add Exodus derivation support

roccomuso:
## Summary

Users who created a Cardano wallet in Exodus with a 12-word recovery phrase could not match AdaLite’s default derivation (Daedalus-style v1 for 12 words). This PR adds an optional Exodus derivation path so those mnemonics resolve to the same addresses as Exodus.

Exodus path is only offered when the mnemonic has 12 words; longer phrases clear the toggle automatically.

### What’s included

- New derivation scheme exodus with BIP39 seed + BIP32 paths under m/44'/1815'/… (aligned with Exodus-style receive addresses).
- Mnemonic login: for 12-word phrases only, a checkbox “Use Exodus wallet derivation path” (with copy and sidebar link to Exodus docs/support).
- Wallet behavior: isExodusWallet drives address discovery limits and Byron feature handling where Exodus differs from legacy v1/v2 flows; optional exodusBip39Seed is kept for derivation and wallet JSON export when relevant.
- Dependencies: bip39, @scure/bip32, and Yarn nodeLinker: node-modules (plus lockfile updates) to support the derivation stack.

### Test

- [x] Unit tests passing
- [x] Got expected address generated
- [x] send/receive ADA

### Screenshot

image

(other examples, not related to the testing seed in the screenshot above)

image

image

PR #1484, opened 2026-08-13
Add exodus and icarus derivation support

refi93:
follow up of https://github.com/vacuumlabs/adalite/pull/1472 with polishing and extra features (Icarus support for 12-word mnemonics) on top


roccomuso (approved):

app/frontend/wallet/helpers/mnemonicToWalletSecretDef.ts
```diff
@@ -9,14 +15,37 @@ const guessDerivationSchemeFromMnemonic = (mnemonic) => {
...
   }
 
+  if (useExodusDerivationPath) {
+    const seed = bip39.mnemonicToSeedSync(mnemonic)
+    return {
+      rootSecret: getExodusRootSecretFromSeed(seed),
+      derivationScheme: derivationSchemes.exodus,
+      exodusBip39Seed: seed,
+    }
+  }
```
PeterBenc: NP: I assume that useExodusDerivationPath is what is set by the select in the UI by user when he enters 12 word mnemonic, in that case id say that we could be more descriptive here and split the two options (deadealos v1 vs exodus) with some union instead of flag, which could make the logic in this function more straight-forward.

app/frontend/wallet/shelley/shelley-address-provider.ts
```diff
@@ -12,6 +12,28 @@ const shelleyPath = (account: number, isChange: boolean, addrIdx: number): BIP32
   ]
 }
 
+/** BIP44 m/44'/1815'/{account}'/{change}/0; Exodus increments account per new receive address */
+const exodusPath = (accountSlot: number, isChange: boolean): BIP32Path => {
+  return [
+    HARDENED_THRESHOLD + 44,
```
PeterBenc: NP: are there `harden` `unharden` utils somewhere?

app/frontend/wallet/shelley/helpers/addresses.ts
```diff
@@ -60,8 +60,12 @@ export const xpub2ChainCode = (xpub: Buffer) => xpub.slice(32, 64)
 // takes xpubkey, converts it to pubkey and then to 28 byte blake2b encoded hash
 const xpub2blake2b224Hash = (xpub: Buffer) => getPubKeyBlake2b224Hash(xpub2pub(xpub))
 
-// TODO: do this more precisely
-export const isShelleyPath = (path: BIP32Path) => path[0] - HARDENED_THRESHOLD === 1852
+/** CIP-1852 or Exodus BIP44 (44'/1815'/...); both use Shelley witnesses, not Byron */
+export const isShelleyPath = (path: BIP32Path): boolean => {
+  const purpose = path[0] - HARDENED_THRESHOLD
+  if (purpose === 1852) return true
+  return path.length >= 5 && purpose === 44 && path[1] - HARDENED_THRESHOLD === 1815
```
PeterBenc: Id suggest: 
```
const purpose = path[0] - HARDENED_THRESHOLD
if (purpose === 1852) return true
const coin = path[1] - HARDENED_THRESHOLD
return path.length === 5 && purpose === 44 && coin === 1815 // note that i chagned the path.length >= 5 to path.length === 5
```

btw return path.length === 5 && purpose === 44 && coin === 1815 is basically "isExodusPath" right?


Btw does this function rely on the path coming from a specific place right? It has to be "account defining" path, so it length 3 in normal cases and length 5 in case of exodus right? Because otherwise a normal byron path of lenght 5 could be interpreted as Shelley right?

app/frontend/wallet/shelley/shelley-js-crypto-provider.ts
```diff
@@ -207,7 +231,14 @@ CryptoProviderParams): Promise<CryptoProvider> => {
   }
 
   function isFeatureSupported(feature: CryptoProviderFeature) {
-    return feature !== CryptoProviderFeature.POOL_OWNER
+    if (feature === CryptoProviderFeature.POOL_OWNER) {
+      return false
+    }
+    // Exodus imports resolve to a single Shelley path; there is no v1/v2 Byron account tree.
+    if (feature === CryptoProviderFeature.BYRON && isExodusWallet) {
```
PeterBenc: Are there any other features that exodus does not suport, considering there is only one path for everything? like voting?

app/frontend/wallet/shelley/shelley-address-provider.ts
```diff
@@ -38,7 +68,9 @@ export const getAccountXpub = async (
   cryptoProvider: CryptoProvider,
   accountIndex: number
 ): Promise<_XPubKey> => {
-  const path = shelleyStakeAccountPath(accountIndex).slice(0, 3)
+  const path = isExodusScheme(cryptoProvider)
+    ? exodusAccountPathPrefix(accountIndex)
```
PeterBenc: This one I'd like resolved before merge.

For Exodus, every level goes independently through `HDKey.derive(...)` → `seedToKeypairV1(...)`, so the xpub at `m/44'/1815'/0'` is **not** the BIP32-Ed25519 parent of the address key at `m/44'/1815'/0'/0/0`. Verified against the actual deps:

```
account-derived child xpub   b7d41b1ee8b9095b309ff1cf1bb8d2da…
actual address xpub          83209e75359f33ce2969d90c53cc7441…
account xpub is parent?      false
```

That value is surfaced verbatim on **Advanced → Keys** as "Shelley extended public key (44'/1815'/0')". Anyone who copies it into a watch-only tool or explorer gets a wallet that isn't theirs — silently, with no error.

`byronAccountXpub` already returns `null` for Exodus via the `BYRON` feature gate and `keys.tsx` guards on it, so the plumbing for hiding a key is already there. Could we either hide the Shelley xpub for Exodus the same way, or render the actual address-level path instead of the account prefix?

Same applies one line down to the "Staking key CBOR hex" row — for Exodus that path is the payment key too, so the label is at best confusing.

app/frontend/wallet/address-manager.ts
```diff
@@ -37,6 +44,10 @@ const AddressManager = ({addressProvider, gapLimit, blockchainExplorer}: Address
   }
 
   async function discoverAddresses(): Promise<Address[]> {
+    if (fixedDiscoveryCount != null) {
```
PeterBenc: NP, and it's the same shape as the `mnemonicToWalletSecretDef` cleanup you already did one layer up.

`AddressManager` now takes both `gapLimit` and `fixedDiscoveryCount`, but when `fixedDiscoveryCount` is set `gapLimit` is dead — except it's still validated in the constructor, so a caller has to pass a meaningful-looking value that is then ignored. `account.ts` works around this by passing `EXODUS_ADDRESS_SCAN_LIMIT` as *both*, which reads like a coincidence rather than an intent.

Two mutually exclusive modes want to be a union:

```ts
type Discovery =
  | {mode: 'gap'; gapLimit: number}
  | {mode: 'fixed'; count: number}
```

Relatedly, `account.ts` derives three separate params (`gapLimit`, `fixedDiscoveryCount`, `skipInternalAddresses`) from one `isExodus` boolean at the call site, so the "what Exodus discovery looks like" rule is spread across the caller rather than living in one place. Not blocking, but if you're touching it anyway it'd collapse nicely.

app/frontend/wallet/keypass-json.ts
```diff
@@ -81,13 +81,20 @@ function parseWalletExportObj(walletExportObj) {
...
+  }
+
   const {passwordHash: b64PasswordHash, walletSecretKey: b64WalletSecret} = walletExportObj.wallet
   const passwordHash = Buffer.from(b64PasswordHash, 'base64')
   const derivationScheme = Object.values(derivationSchemes).find(
     (x) => x.keyfileVersion === walletExportObj.fileVersion
   )
 
-  if (derivationScheme === undefined) {
+  if (derivationScheme === undefined || derivationScheme.type === 'exodus') {
```
PeterBenc: NP: this is guarded twice. The explicit `fileVersion === derivationSchemes.exodus.keyfileVersion` check above already throws (with the better message), so `derivationScheme.type === 'exodus'` here can never be reached.

Not harmful — belt and braces on a security boundary is defensible — but if you'd rather have one, I'd keep this one and drop the earlier block, since it can't drift out of sync with `derivationSchemes`.

PeterBenc (approved):
LGTM, but check my comments pls, at least one is worth considering. I initially notices some issues with isShelleyPath which you fixed in later commit, thanks..

app/frontend/wallet/shelley/shelley-address-provider.ts
refi93: Fixed in e302b913. `getAccountXpub` now returns `null` for Exodus (same pattern as Byron), and Advanced → Keys hides both the Shelley xpub and the staking-key CBOR / `stake.vkey` row. Reward address and staking key hash stay visible — those still describe the stake credential on the receive address.

app/frontend/wallet/keypass-json.ts
refi93: Collapsed to a single guard keyed off `derivationScheme.type === 'exodus'` so it cannot drift from `derivationSchemes`, and kept the more specific error message.

app/frontend/wallet/address-manager.ts
refi93: Addressed in the same commit. `AddressManager` now takes a `discovery` union (`{mode: 'gap', gapLimit}` | `{mode: 'fixed', count}`), so fixed-mode callers no longer pass a dummy gap limit. Exodus still sets `skipInternalAddresses` next to that at the `MyAddresses` call site.

